PR #1, opened 2015-10-01
Sydb xenodata

sydb:
As there have been no complaints, I think the <xenoData> branch is now ready to be merged into master.


hcayless:
That looks a bit terrifying. Are you sure you’re working off the right copy of the repo?

sydb:
> That looks a bit terrifying. Are you sure you’re working off the
> right copy of the repo?

No, I'm not at all sure; did I do it wrong? I'm was trying to request
that someone merge syd-xenodata into master. But the commit messages
below make it look like it's the right branch. They list everything
that happened before it was branched off, too, but it does list up to
the latest change in sydb-xenodata, which was to predeclare
macro.anyXML. 

From where was this mail sent, and to where is it going? I'm CCing
the council list in case it winds up in a bit bucket.

>

PR #1354, opened 2015-10-26
added some parameters for Jenkins port etc.

peterstadler:
some modifications to ease installation. Tested on Debian 8


jamescummings:
Looks good to me.  I'll have to check if if oxford jenkins can send as tei-council...

PR #1443, opened 2016-04-04
fix ISBN to correspond with those in idno spec

stuartyeates:


peterstadler:
Looks like a corrigible error to me

peterstadler:
Thanks for spotting!

PR #1551, opened 2016-12-03
Sydb model attr checks

sydb:
James, Magdelena, or someone else who actually has significant processing model experience should probably test these before pulling.


hcayless:
Testing fails with the following in modelSequence in testall.xsd:

````xml
<xs:element maxOccurs="unbounded" ref="tei:model"/>
<xs:element maxOccurs="unbounded" ref="tei:model"/>
````
which is nondeterministic, and presumably should be
````xml
<xs:element ref="tei:model"/>
<xs:element maxOccurs="unbounded" ref="tei:model"/>
````

sydb:
Yes and no. The minor issue is that your suggested solution is still no-deterministic. But more importantly, I don't think this error has anything to do with the stuff in the pull request. We are getting this error in dev already, aren't we?

hcayless:
Um. No it isn't. Oxygen likes it anyway. And no, the tests are passing in dev.

sydb:
OK. Will look into it this afternoon.

hcayless:
Never mind. I had my Stylesheets copy misconfigured. Trying it again with the dependency on your other pull request properly satisfied has a happier outcome :-)

sydb:
BTW, dear reader, Hugh was correct. Obviously `( M, M* )` is deterministic. Don’t know what I was thinking.

PR #1612, opened 2017-03-13
added a check for rnv with a URL hint

bansp:
`rnv` is invoked in the Makefile under Test/ but is not checked for, which may lead to some bitterness. Furthermore, neither of the *ubuntu systems that I run recognizes `rnv` among the software packages. So I added a little URL hint on where to find it.

### DIFF
--- a/P5/Makefile
+++ b/P5/Makefile
@@ -47,7 +47,7 @@ check.stamp:
 	@command -v  java || exit 1
 	@echo -n xmllint:
 	@command -v  xmllint || exit 1
-	@echo -n rnv (https://sourceforge.net/projects/rnv/):
+	@echo -n rnv (https://github.com/hartwork/rnv):
 	@command -v  rnv || exit 1
 	touch check.stamp
 
--- a/P5/Makefile
+++ b/P5/Makefile
@@ -47,6 +47,8 @@ check.stamp:
 	@command -v  java || exit 1
 	@echo -n xmllint:
 	@command -v  xmllint || exit 1
+	@echo -n rnv (https://sourceforge.net/projects/rnv/):
+	@command -v  rnv || exit 1
 	touch check.stamp
 
 p5.xml: ${DRIVER} Source/Specs/*.xml Source/Guidelines/en/*.xml p5odds.odd check.stamp
--- a/P5/Makefile
+++ b/P5/Makefile
@@ -47,7 +47,7 @@ check.stamp:
 	@command -v  java || exit 1
 	@echo -n xmllint:
 	@command -v  xmllint || exit 1
-	@echo -n rnv (https://github.com/hartwork/rnv):
+	@echo -n rnv (https://github.com/dtolpin/RNV):
 	@command -v  rnv || exit 1
 	touch check.stamp
 


martindholmes:
We have to build it ourselves in the Jenkins server build:

https://raw.githubusercontent.com/TEIC/Jenkins/dev/docker/Dockerfile

~~~~~
# We need to build rnv locally since it's no longer packaged for
# Debian. Doing this before installing other stuff because it's
# quicker for testing purposes. We need make for that. Do apt-utils
# first so that it's available when other stuff is being installed.
# libexpat-dev is required to build rnv.
RUN apt-get update && apt-get --yes --force-yes --no-install-recommends install apt-utils
RUN apt-get --yes --force-yes --no-install-recommends install make build-essential libexpat-dev

RUN git clone https://github.com/dtolpin/RNV.git rnv && \ 
    cd rnv && \ 
    make -f Makefile.gnu rnv && \
    cp rnv /usr/bin/ && \
    cd ../
~~~~~

We did have some interest from a previous rnv maintainer to bring it back into the Debian folder, but nothing came of it.

bansp:
So if I interpret your reply correctly, the Docker package is taken care of. But if someone decides to use the cloned source, they need to do their own investigation, and the proposal here was supposed to make it easier on such people (because `make check` is recommended in at least one online help document).

martindholmes:
Sorry Piotr, I wasn't clear: my comment was intended to show where to get the latest version from; it's github, not sourceforge. The project has bounced between two maintainers and it's quite hard to track down who has it. I would point at the dtolpin git repo rather than sourceforge.

bansp:
Ahhh! Thanks! I'll correct the request (and will also reinstall rnv on my system... :-))

bansp:
Oh, actually, @martindholmes, you may want to change the address from dtolpin's repository to https://github.com/hartwork/rnv , where the current code is kept, and it's in sync with SF, which is apparently used as the distribution platform. I modified the URL in the hint to the above as well.
(And I hope that it's automatically reflected in the pull request.)

martindholmes:
@bansp I think I've been round the cycle of hartwork and dtolpin a few times over the last few years, and I think at the time I was working on it I just picked the one that was most up to date. As far as I can see, the sourceforge version is long out of date (last commit 2012) but no-one has released an update recently. I'd be glad to hear that Sebastian (hartwork) was working on it again; at one point, he was interested in trying to get it back into Debian. But it looks like nothing has happened since the 2015 activity. My guess would be that the codebases are basically identical in the two GitHub repos, but if you find a difference let me know and I'll change the docker script.

bansp:
I think you're right that they are identical, and untouched for 2 years. I trusted the remark at dtolpin's ("current version is at https://github.com/hartwork/rnv") without verification.

peterstadler:
@bansp If you adjust the Url in your PR I'm happy to merge it :)

bansp:
Peter, I'm actually lost: should I update to a repository at the top of which the user is going to read "go elsewhere"? Bah, yeah, let them. I'll update in a moment.

bansp:
Done.

martindholmes:
I'll contact the person Sebastian Pipping at some point and see if there's any realistic chance that rnv could end up back in the Debian fold. That would be ideal.

peterstadler:
Ok, I should have looked more closely …
Now reverted the change because a) Mr. Jenkins complained, and b) the `rnv` usage seems optional, not mandatory for the build (see line 184 of the Makefile, where a check is in place for `rnv` already). Nevertheless, I added the URL there thanks to  @bansp's PR.

bansp:
Thanks, @peterstadler. The reason I thought rnv would be good to explicitly check for is that if you don't have it on the system, the make process fails somehow -- I forget how exactly, now; maybe it fails partially, the point being that if you don't watch the terminal, you have a chance to miss the information, and then you get puzzled when you expect your system to be like someone else's but for some cryptic reason it isn't. I don't know what Mr. Jenkins complained of, maybe rnv is not in his pocket? :-)

One way or another, if the build process assumes some tools, whether they are optional or very specialized (like the xetex you mentioned), _that information should be shoved under the user's eyes, any which way_. It should never be hidden or else it's an innocent onset of a path towards magical obfuscation and, in the end, user-infuriation. 

`make check` seemed like a sensible compromise. It's called by the user, it doesn't interfere, it can provide information on both optional and obligatory dependencies. I'd say that if Mr. Jenkins complains, he may need a kick or two, but the information should be there, for the user. Mr. Jenkins should serve, rather than rule...

martindholmes:
One of the reasons we committed to the whole Jenkins approach was that it was becoming increasingly onerous for anyone to have a system set up and maintained that included all the tools needed to run the whole build; and it was also increasingly unlikely that a single person working on a specific issue would actually need to run the entire build. I think the combination of Jenkins and Hugh's docker instance for people who need to build is a good one.

peterstadler:
@bansp I agree with you that the build process could (should) give better output to the user. Your  PR was an attempt in that direction but it turned out the change wasn't as simple as I hoped.

* many targets depend on the check target, so it's not only for the user's convenience
* we'd need to be more explicit for which targets (= which parts of the build) the tools are necessary to give proper feedback (which would need more research)
* I reverted the change to get back a working build. That doesn't mean we can't fix it :)

bansp:
Gentlemen, at the risk of sounding uncool and non-gittish, let me admit that I failed to install from dtolpin, then failed to install from hartwork (configure complained), and succeeded only with the package from SourceForge (configure first gently admonished me about the lack of expat, and afterwords worked smoothly). This is on my laptop. Earlier, I also installed from SF on my desktop. Conclusion: for the undergeeks, this is the address to use: https://sourceforge.net/projects/rnv/

bansp:
Hi again. I compiled rnv but fortunately forgot to install it :-), so now I can show you exactly what happens:
I ran `make` in TEI/P5, and this is what I see, after 10 minutes or so:

```
...
BUILD SUCCESSFUL
Total time: 17 seconds
java -jar ../Utilities/lib/trang.jar testplus.rng testplus.rnc
xmllint --noent --dropdtd  testplus.xml | rnv testplus.rnc
/bin/sh: 1: rnv: not found
Makefile:52: recipe for target 'testplus.special' failed
make[1]: *** [testplus.special] Error 127
make[1]: Leaving directory '/home/piotr/projects/LingSIG/TEI/P5/Test'
Makefile:196: recipe for target 'test' failed
make: *** [test] Error 2
```

My point being: I don't know if it's optional, and the make process definitely doesn't tell me that. It just aborts with an error, after _quite_ a while of milling stuff around on my disk. Frusstratiiion. We don't want our users frustrated! :-)

martindholmes:
@bansp Are regular users really running the test suite? I would think the regular build process would be the most likely thing for them to run. And if anyone is really serious about building and testing, because they're actually developing on TEI, would they not be better off using Hugh's docker instance?

lb42:
I understand that the docker thingie is for Windows? In what sense would using Windows make people better off? (Don't answer that)

bansp:
@martindholmes At which point have I said "test suite"? :-) I ran `make` under TEI/P5...

bansp:
Aha, and I sometimes think that I am serious about building and testing, but until now I thought I wasn't risking any kind of stigma to run straight off the source... :-)

martindholmes:
@bansp Whatever targets you're running from the main makefile, you're ending up running the test suite; would most people need to do that?
@lb42 Docker runs on Linux, Win and Mac. Hugh was recently asking for someone to test his instance on Windows since he couldn't do it himself; Elisa was offering to try because she has a Windows instance. Perhaps that's why you're thinking it's Windows-only?

lb42:
No, actually I was thinking of an email Hugh sent to the council list 21/12/2016 when I asked whether I needed to pay attention to this Docker thingummy. I quote:
 
Probably superfluous on Linux, yeah.

bansp:
@martindholmes Ah, now I think I understand your first question: you were implying that the test suite part should not be part of the regular build process. I'm wholeheartedly with you on that. Should I prepare a separate ticket?

martindholmes:
@bansp What targets were you running? Jenkins runs `clean deb dist`; presumably you don't need deb, so `make clean dist` would I think give you what you need for a build, and I don't think that would run the tests, would it?

bansp:
Gee, I'll try that, thanks. I followed the  instructions from the readme for the Github repo, and made sure to take the path of a "typical user", so without juggling with targets. In other words, I ran... `make`.
Here is the doc: http://www.tei-c.org/Guidelines/P5/get.xml , should it perhaps be modified with a hint on how to skip the tests and get `p5subset.xml` and the potentially modified chapters?

martindholmes:
@bansp I agree that we should update the list of targets on that page and provide some sample sets of targets for specific tasks. Part of the problem is that the checks are done in the P5/Makefile (check.stamp) but the main dependency is in the P5/Test/Makefile. As far as I can see, all the targets that make use of rnv in the main Makefile test for it before invoking it, and fail silently:

` -command -v  rnv && rnv -v p5odds.rnc p5.xml`

whereas in the P5/Test/Makefile, it's simply assumed to be there:

` rnv frag.rnc testfrag.xml`

I don't want to impose a requirement for rnv on the main Makefile when it's not actually needed; that will just get in people's way. But if you end up running the tests, we should check for it and fail gracefully with an informative message.

bansp:
A remark: an update of the target list will always be appreciated, but I think that there's something wrong from the perspective of user-oriented design (or simply from the perspective of "keeping things simple") when `make` alone is more complex than a complex command. That is, I would expect to run `make` to just _get the basic things done_, and e.g. to need to `make test` in order to get the tests processed.

martindholmes:
Anyone who's at the level where they're running make on their machine is not a naive user, surely? This is not a design issue; we're not expecting normal users to build the TEI project. I'm all in favour of making things clearer for everybody, but the idea that we apply a complete set of design principles to the command-line build process seems like overkill to me.

bansp:
If you have a story for why not, I won't be protesting. I don't know about levels of naivety :-) It seemed to me that the get.xml article was directed at a fairly naive user who would like to move a step further, in helping fix stuff and so on (so OK, maybe "half-naive"). And I'm saying this to explain myself, rather than to argue. I did read the doc before attempting anything, and was sure that `make` was exactly what I should do in order to get the most current p5subset.xml, and that the time for all the tests to roll is just the extra cost of it all.

martindholmes:
Perhaps that document should now basically be saying "Install docker; get Hugh's docker container; map the folder with your TEI source into the docker instance; run make operations in docker".

But this is becoming ludicrous: there IS no dependency on rnv. It's a nice-to-have alternative validation method that we keep around because it's useful to have more than one way of validating against a RelaxNG schema, but we debated dropping it when it got difficult. If we just change these things:

`rnv frag.rnc testfrag.xml`

like this:

`rnv && rnv frag.rnc testfrag.xml`

in the Tests Makefile, won't that solve your problem completely?

bansp:
I don't have a problem any longer, because I have installed rnv on both my systems. I was rather concerned with other "half-naive" users hitting this issue and getting frustrated, or worse: discouraged.

As to your question, directly: will such a check mean that success may get reported even in the absence of a successfully completed bunch of tests?

martindholmes:
This is a typical fragment:
~~~~~
test-frag:
	@echo BUILD: Testing files which exercise inclusion of fragments
	${JING} -t -c frag.rnc testfrag.xml
	rnv frag.rnc testfrag.xml
~~~~~

The XML is validated first with Jing, then with rnv. The rnv validation is just nice-to-have; there's no expectation that rnv will find some invalidity that Jing doesn't notice. If the file is invalid, Jing will catch it and fail the build.

martindholmes:
Line 94 already has this:

`-rnv detest.rnc detest.xml > detest2.log 2>&1`

where the leading hyphen means "ignore the status resulting from this command", so the build won't fail if the command returns an error. On my machine, I see this:

~~~~~
rnv detest.rnc detest.xml > detest2.log 2>&1
Makefile:88: recipe for target 'detest' failed
make: [detest] Error 1 (ignored)
~~~~~

so the test is actually failing. Was this the one you were seeing? 

The fact that this test fails on my machine suggests that it's broken and not really useful anyway; the irony is that if you have rnv and the test runs and fails, the result is ignored, but if you don't have rnv you get a proper failure. I'm going to change that line to:

`-command -v rnv && rnv detest.rnc detest.xml > detest2.log 2>&1`

so that it's still there, but it only tries to run if rnv is available, and then it fails silently. Someone, though, should investigate what it's testing and why it currently fails.

martindholmes:
Made that change in commit 2e949c2.

lb42:
The detest tests are ones with deliberate errors arent they? So it s expected to fail.

martindholmes:
Doh! Of course they are. Is my change destructive, then? I can't really see an alternative approach.

lb42:
How about leaving it alone?

martindholmes:
Piotr is adamant that the current behaviour combined with the current documentation is unacceptable because it throws up errors for naive users just running "make" with no targets, and that's what he says most people will do.

lb42:
I haven't been looking at this closely, but I see three options. (1) test for the existence of rnv at the start of make check (2) remove rnv from make check (3) tell piotr we have no evidence of anyone except him finding this a problem

bansp:
I'd rather say "concerned" than "adamant", and I'm also far from proclaiming myself a spokesperson for "naive users", so I'm fine with all 3 solutions, (2) sounding the least attractive.

martindholmes:
If we test for rnv at the beginning of everything, then the vast majority of users who are not (presumably) running the tests because they've picked the make targets they want will see an error that they don't need to see. A better option would be to test for it at the beginning of the test Makefile, I think. That specific detest that uses it could be done with Jing instead, though.

martindholmes:
Committed a better fix in commit fd0e073. The Test Makefile is now checking for rnv and warning about its absence only for targets that need it. If all is well with Mr Jenkins, I believe we can delete the branch and close this ticket.

bansp:
Thank you, Martin :-)

martindholmes:
Thank me when it's actually working. :-( This is my check:

~~~~~
checkrnv:
    $(if $(shell command -v rnv 2> /dev/null),$(info Found rnv),$(warning WARNING: rnv is not installed. Some targets may fail.))
~~~~~

and on Jenkins, although rnv is installed and working, the warning is returned and the build shows as unstable. Any experts on make who could tell me what's wrong with my line?

martindholmes:
This is still strangely failing, not in TEIP5-Test-dev but in TEIP5-dev. Despite the fact that if I (as the jenkins user) in the Test folder run:

make check rnv

I get this back:

make checkrnv
Found rnv
make: Nothing to be done for `checkrnv'.

I still see the error message during the actual build run. I'm at a loss here.

lb42:
I can't reproduce this. I did a git pull and deleted my local copy of rnv.  In my local installation I did "make clean; check" which rolled along happily to conclusion. 

Then I did a "make test" and got the expected message:
~~~~
lou@foxglove:~/Public/TEI/P5$ make test
BUILD Run test cases for P5
(cd Test; make XSL=/usr/share/xml/tei/stylesheet VCS=git)
make[1]: Entering directory '/home/lou/Public/TEI/P5/Test'
Makefile:29: WARNING: rnv is not installed. Some targets may fail.
~~~~
Where's the beef?

martindholmes:
You got the correct warning because (presumably) rnv is not installed on your system; but rnv IS installed and working on Jenkins, but the warning still appears. But not in TEIP5-Test-dev; only in TEIP5-dev.

It's got me proper befuddled.

martindholmes:
I think this is now fixed in commit 7a1e14543c216e5019eab5d24d5bafa44c60c635.

martindholmes:
@bansp, can you test on your setup now? OK to delete this branch?

bansp:
Thanks, I'll let you know tomorrow at the latest.

bansp:
I started by running `make` with rnv still installed, and I can already see something useful for me that I don't see when I do `make schemas` and `make web-html`, which are now the two targets that I restrict myself to using. Namely, `make valid` fails with a long java stack trace, and it does so only on my branch, because on `dev` everything is fine. I get this:

```
BUILD: Check validity with special-purpose XSL code, looking for bad links etc
ANT_OPTS="-Xss2m -Xmx752m -Djava.awt.headless=true" ant -q -lib Utilities/lib/saxon9he.jar -f antbuilder.xml -DXSL=/usr/share/xml/tei/stylesheet validators
     [echo] Run Schematron script (normal part of Guidelines)
     [echo] Run Schematron script (Examples in Guidelines marked as valid)

BUILD FAILED
/home/piotr/projects/LingSIG/TEI/P5/antbuilder.xml:190: javax.xml.transform.TransformerConfigurationException: net.sf.saxon.s9api.SaxonApiException: Errors were reported during stylesheet compilation
(+ many, many lines of trace)
```

Let me stress that this isn't a problem with the `dev` branch and I'm just taking an opportunity to ask where I can look for the more detailed diagnostics. I have a whole new directory, `valid/`, full of many little specs, but I can't see how I could use them for investigating (looked at the freshest and at the last one in terms of the filename, but they give me no clue). Would you guys please drop me a hint while I'm uninstalling `rnv` and testing without it??

bansp:
Now, without rnv, it obviously fails but only half a screen before it aborts, I can see a warning message, so, well, it's something. This is the end of the output:
```
Validate XSD using Jing ...
java -jar ../Utilities/lib/jing.jar testall.xsd testall.xml
WARNING: rnv is not installed. Some targets may fail.
BUILD:  Test schema from (special) testplus.odd
xmllint --xinclude testplus.odd > xi_testplus.odd
ANT_OPTS="-Xss2m -Xmx752m -Djava.awt.headless=true" ant -lib ../Utilities/lib/saxon9he.jar:../Utilities/lib/jing.jar -Dtrang=../Utilities/lib/trang.jar  -DdefaultSource=`pwd`/../p5subset.xml -DXSL=/usr/share/xml/tei/stylesheet -f antruntest.xml -Dtestfile=testplus.xml -DoddFile=xi_testplus.odd -Doutputname=testplus validateodd compileodd rng validaterng cleanup
Buildfile: /home/piotr/projects/LingSIG/TEI/P5/Test/antruntest.xml

validateodd:
     [echo] Validate xi_testplus.odd as ODD
     [xslt] Processing /home/piotr/projects/LingSIG/TEI/P5/Test/xi_testplus.odd to /dev/null
     [xslt] Loading stylesheet /home/piotr/projects/LingSIG/TEI/P5/p5odds.message.isosch.xsl

compileodd:
     [echo] Do ODD processing using ../Utilities/lib/saxon9he.jar and source /home/piotr/projects/LingSIG/TEI/P5/Test/../p5subset.xml
     [xslt] Processing /home/piotr/projects/LingSIG/TEI/P5/Test/xi_testplus.odd to /home/piotr/projects/LingSIG/TEI/P5/Test/testplus.odd.processed
     [xslt] Loading stylesheet /usr/share/xml/tei/stylesheet/odds/odd2odd.xsl
Warning: XML resolver not found; external catalogs will be ignored

rng:
     [echo] XSLT generate RELAXNG testplus.rng using /home/piotr/projects/LingSIG/TEI/P5/Test/../p5subset.xml
     [xslt] Processing /home/piotr/projects/LingSIG/TEI/P5/Test/testplus.odd.processed to /home/piotr/projects/LingSIG/TEI/P5/Test/testplus.rng
     [xslt] Loading stylesheet /usr/share/xml/tei/stylesheet/profiles/default/relaxng/to.xsl
Warning: XML resolver not found; external catalogs will be ignored

validaterng:
     [echo] Validate testplus.xml against RELAXNG testplus.rng

cleanup:
   [delete] Deleting: /home/piotr/projects/LingSIG/TEI/P5/Test/testplus.odd.processed

BUILD SUCCESSFUL
Total time: 6 seconds
java -jar ../Utilities/lib/trang.jar testplus.rng testplus.rnc
xmllint --noent --dropdtd  testplus.xml | rnv testplus.rnc
/bin/sh: 1: rnv: not found
Makefile:55: recipe for target 'testplus.special' failed
make[1]: *** [testplus.special] Error 127
make[1]: Leaving directory '/home/piotr/projects/LingSIG/TEI/P5/Test'
Makefile:196: recipe for target 'test' failed
make: *** [test] Error 2
```

Thanks for addressing this, Martin!

martindholmes:
I can't think of a simple solution here. We can't have the build process ignore failures if rnv is installed, otherwise we're wasting our time using rnv to check anything; the whole point is for our Jenkins boxes (rather than ordinary users) to report problems as soon as they occur. I guess we need a slightly more complicated logic:

 - if rnv is not installed, figure out whether you're running on Jenkins or not; if you are, then fail; if you're not, then set a variable and warn the user.
 - if rnv is installed, then run rnv tasks as normal
 - if rnv is not installed, and you're not on Jenkins, don't run the rnv tasks

Another option is to remove the test target from the default set, so that ordinary users, who are not likely to care about the tests anyway, and who don't know how to figure out which make targets they need to run, just don't get the tests. Surely you'd only run the tests if you were a sophisticated user working on the source code, in which case you'd figure out or ask how to use the docker builder or install rnv for yourself?

bansp:
I see two not necessarily mutually exclusive solutions here: (i) Lou's solution (3) from above ("tell Piotr..."), and (ii) indeed removing the test target from the default -- after all, `make`, `make test`, and `make install` is not an unthinkable order or things.

martindholmes:
The default is actually a very odd collection of things:

default: validate exemplars test html-web

I don't see who that would be useful for at all. Perhaps it would be better if default just showed a detailed explanatory message providing a list of sample use-cases with the targets that would be suitable for them?

bansp:
That seems to be a user- and system-friendly move.

Since the initial posting in this thread, I have learned to use `make clean schemas` (most often) and also `make clean web-html` (when I hook my stuff up to the Guidelines and want to see how it renders). I _might_ start using `make validate`, depending on the outcome of https://github.com/TEIC/TEI/issues/1627 (because at present, it looks like it fails where it shouldn't).

Off-topic (since it's probably a matter for another issue), `make clean` doesn't clean everything (depending on the targets that have been used; minimally, it seems to lack `rm repodate.xml`).

Issue #1643, opened 2017-05-22
jTEI should include <licence> element


rvdb:
In order to adequately express the new author agreements in jTEI articles, the &lt;licence> element would be needed: 

    <availability>
      <licence target="https://creativecommons.org/licenses/by/4.0/">
        <p>For this publication, a Creative Commons Attribution 4.0
          International license has been granted by the author(s), who retain full
          copyright.</p>
      </licence>
    </availability> 

Since the jTEI ODD currently doesn't include the &lt;licence> element, this should be added.

PR #1644, opened 2017-05-22
added licence element to jTEI ODD

rvdb:
fixes https://github.com/TEIC/TEI/issues/1643


rvdb:
fixes https://github.com/TEIC/TEI/issues/1643

Issue #1670, opened 2017-07-22
att.linguistic for <w> and <pc>


bansp:
### Quick links:
1. [diff of the pull request](https://github.com/TEIC/TEI/pull/1671/files)  (will be kept synced against TEIC/TEI/dev)
1. [suggested text of the relevant chapter](http://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-wordAttributes/lastSuccessfulBuild/artifact/P5/release/doc/tei-p5-doc/en/html/AI.html#AILC) (minimal changes, pending acceptance)
1. [suggested documentation of att.linguistic](http://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-wordAttributes/lastSuccessfulBuild/artifact/P5/release/doc/tei-p5-doc/en/html/ref-att.linguistic.html)

### Introduction and summary
We propose an attribute class that gathers token-level attributes facilitating simple linguistic annotation. We propose that `<w>` and `<pc>` should be its members by default. At the same time, grouping these attributes in a separate class (rather than defining them inside an element) makes it possible for other token-level elements to be added to that class, in TEI customizations. Our proposal addresses both the future uses (extending the TEI offer towards corpus linguists) as well as the existing resources (enriching the existing, for example literary or historical corpora, with basic linguistic annotation).

The proposed class, att.linguistic, contains first of all two linguistic attributes originally defined inside `<w>`, namely `@lemma` and `@lemmaRef`. In addition to these, we suggest to add the following: 

* `@pos` (for part-of-speech)
* `@msd` (for morphosyntactic description)
* `@reg` (for the regularized form in e.g. historical corpora)
* `@join` to signal the absence of whitespace next to the element -- this is meant for all non-mixed-content contexts where the whitespace information is not conveyed by the actual presence of whitespace characters

We stress that the goal of this feature request is not to facilitate in-depth linguistic annotation, but rather to equip "off-the-shelf" TEI in the very basic tools that linguists can use, and that non-linguists can safely add to their existing resources in order to enhance them. This is why we address the TEI namespace directly. Below, we first present the motivation for this feature request, offer a detailed description of the suggested attributes, and then address some potential counterproposals and provide further examples.

[...]

PR #1671, opened 2017-07-22
att.linguistic for <w> and <pc>

bansp:
This pull request is described in issue #1670 .

It would probably be most productive to keep any discussion of this pull request in the comments to the issue, so as to minimize the potential confusion.


sydb:
In commit 33348e3 I have added an example to w.xml, as validation really wants one there. (And it’s not a bad idea.)

bansp:
Thank you, Syd. I hope it's all right if I add the `@join` attribute there?
We're in the process of preparing the final version of our article, so that's where the energy is currently being channelled, but I haven't forgotten about the promised pull request for documentation :-)

PR #1683, opened 2017-08-31
expanded front/div to all types supported by OpenEdition (see https:/…

rvdb:
…/github.com/OpenEdition/tei.openedition/wiki/Creating-a-TEI-document-in-Lodel-1.0#author-notes-editors-notes-errata-acknowledgements)

### DIFF
--- a/P5/Exemplars/tei_jtei.odd
+++ b/P5/Exemplars/tei_jtei.odd
@@ -2212,9 +2212,9 @@
           <elementSpec ident="div" mode="change">
             <constraintSpec ident="jtei.sch-divtypes-front" scheme="isoschematron">
               <constraint>
-                <sch:rule context="tei:div[@type = ('abstract', 'acknowledgements')]">
+                <sch:rule context="tei:div[@type = $div.types.front]">
                   <sch:assert test="parent::tei:front">
-                    Abstracts (<sch:name/>[@type="abstract"]) and acknowledgements (<sch:name/>[@type="acknowledgements"]) may only occur inside front.
+                    A text division of type <sch:value-of select="@type"/> may only occur inside front.
                   </sch:assert>
                 </sch:rule>
               </constraint>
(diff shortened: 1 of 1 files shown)


rvdb:
@martindholmes , I've asked you to review this, since you probably have sensible opinions on this. I'm proposing an expansion of the `front/div/@type` values to all types supported by the [OpenEdition](https://github.com/OpenEdition/tei.openedition/wiki/Creating-a-TEI-document-in-Lodel-1.0#author-notes-editors-notes-errata-acknowledgements) schema. The sole differences from OpenEdition are:

- instead of encoding author's and editor's notes in `<note>`, in the jTEI schema I'm proposing a uniform `<div>` encoding for all sections that an occur inside `<front>`
- I'm proposing a more uniform, human-readable list of `front/div/@type` values for the jTEI schema

While a fairly trivial change of the closed value list for `div/@type` in the ODD file, one thing to consider is that not all types are meant for authors (i.e. 'editorialIntroduction', 'editorNotes', 'corrections'). I've tried to make that explicit by pointing this out in the respective `//attDef/valList/valItem/desc` descriptions, and grouping the 'editorial' types at the end of the list.

I've also updated the Schematron checks so they allow for all of these types in the `<front>` section (only).

It would be helpful if you could check this makes sense, and that the documentation is clear enough in this respect.

martindholmes:
Hi @rvdb. I don't see any problem with it. I don't know what the use-case for the new types is either, though. We don't necessarily have to support everything supported by OpenEdition if we don't have a need for it, and I don't remember any cases of author notes, editor notes or corrections. Corrections would be useful for sure, if something in an article needs to be updated after the initial publication. Do you have new articles in process that can use these values?

rvdb:
Hi @martindholmes , we indeed have a concrete need for corrections. Of course I can imagine use cases for all other categories, but -except for acknowledgements- we haven't encountered the need in real life so far. Of course, there's no way to silently accommodate for all types: they'll be visible in the ODD documentation, auto-completion lists, and Schematron feedback. And once they're visible in the authoring process, there's no doubt the need will arise.

rvdb:
So, if commenting out unneeded values for the time being is a compromise, I'd be happy to do so (though I don't know what I'd have to do next in Git land...)

martindholmes:
Don't worry about commenting out; please just go ahead and merge. I was just being cautious. :-)

Issue #1700, opened 2017-10-27
grouping issue in P5 guidelines layoutDesc examples


duncdrum:
``<exemplum versionDate="2008-04-06" xml:lang="fr">`` contains english prose, no french.
``<exemplum xml:lang="zh-TW">`` and ``<exemplum xml:lang="en">`` both mix English and Chinese prose

```xml
<exemplum xml:lang="zh-TW">
    <egXML xmlns="http://www.tei-c.org/ns/Examples">
      <layoutDesc>
        <layout columns="2" ruledLines="42">
          <p><locus from="f12r" to="f15v"/> 兩欄共42行以墨水圈選、標誌，欄間有直線分隔。</p>
        </layout>
        <layout columns="3">
          <p><locus from="f16"/>Prickings for three columns are visible.</p>
        </layout>
      </layoutDesc>
    </egXML>
  </exemplum>
  <exemplum xml:lang="en">
    <egXML xmlns="http://www.tei-c.org/ns/Examples">
      <layoutDesc>
        <layout columns="2" ruledLines="42">
          <p><locus from="f12r" to="f15v"/>
   2 columns of 42 lines pricked and ruled in ink, with 
   central rule between the columns.</p>
        </layout>
        <layout columns="3">
          <p xml:lang="zh-TW"><locus from="f16"/>小孔的三欄可見.</p>
        </layout>
      </layoutDesc>
    </egXML>
  </exemplum>
```

this seems rather non-sensical or is missing an ``xml:lang="en"`` attribute. More likely, each exemplum was intended to contain only the language specified by its ``xml:lang`` attribute.

jamescummings:
The correct seems good, but what about the french ones?  It seems strange that we have examples marked as french which are just copies of the english one.

duncdrum:
agreed but my french isn't up to fixing this.

jamescummings:
I was more wondering if we should just remove them. Flagging as needs discussion in case there are any objections. Regardless, we should just accept your patch.

duncdrum:
@jamescummings I don't know how the website generates the contents for the language specific guidelines ala ``http://www.tei-c.org/release/doc/tei-p5-doc/de/html/ref-layout.html`` hence I didn't touch the french. If you want me to delete or comment-out that block I can add another commit to the PR. 
If any francophone gitHub collaborators want to submit a french translation to my PR, push away.

PR #1701, opened 2017-10-27
move <p> to proper language exemplum

duncdrum:
see TEIC/TEI#1700


duncdrum:
Please close if the examples are correct, or if I edited the wrong files.

PR #1733, opened 2018-01-18
updated references: revues.org --> journals.openedition.org

rvdb:


rvdb:
@martindholmes I'll quote from the [English announcement](https://journals.openedition.org/10580?lang=en): 

> An automatic redirection is set up for all journals, which also change address.

This seems to express the intention to keep the old URLs working. They still are, at least, see e.g.: http://jtei.revues.org/1433.

rvdb:
Ok, I've received confirmation from the OpenEdition Journals Team about the long-term URL redirection strategy:

> B​e assured that the redirection will be kept permanently. There is no need no update​ "historical" links in already published articles (unless you have time for that and prefer to do so)

martindholmes:
That's good to know.

PR #1764, opened 2018-04-08
Mdh issue 1453

martindholmes:
Build of the Guidelines is available here:

http://hcmc.uvic.ca/people/martin/tei/mdh_issue_1453/

with specific pages that have changes here:

http://hcmc.uvic.ca/people/martin/tei/mdh_issue_1453/PH.html#PHFAX
http://hcmc.uvic.ca/people/martin/tei/mdh_issue_1453/ref-path.html
http://hcmc.uvic.ca/people/martin/tei/mdh_issue_1453/ref-zone.html
http://hcmc.uvic.ca/people/martin/tei/mdh_issue_1453/ref-surface.html


martinascholger:
Thank you for this, @martindholmes! Here are my remarks and questions:
(1)  the definition of `path` says "defines any line passing through two or more points within a surface element." Actually, `path` is allowed within surface, zone and line. What about "on a surface" (without element) or "two-dimensional area"?  
(2) I try to think of an example where I have a `path` element within a `line`. A line of text contaning something like a scribble or sign? Do you have an example for this?
(3) in `<path>` specs: I wonder if you want to put the comment in Schematron into `<remarks>`?
(4) While looking at the definitions (general and specific) for `@points`, I noticed that `@points` (in `att.coordinated`) is defined as "specified by the other attributes". I'm confused about "the other" attributes. Shouldn't that be "parent element" too?  
(5) Should I provide a German translation for the `path` spec?

martindholmes:
Thanks for looking at this @martinascholger! And thanks for excellent questions:

1. I modelled `<path>` on `<zone>`, so it can appear anywhere that `<zone>` can appear; `<zone>` can nest, so `<path>` can appear within `<zone>`. The definition of `<zone>` also fails to mention that it can appear within `<zone>` or line, so I refrained from mentioning it here too. The normal case would I think be for both elements to appear within `<surface>`; I have use-cases for nested `<zone>`s, and for `<path>`s within `<zone>`s, which I hope to talk about in Tokyo, but I don't know whether we should make these slightly obscure cases front-and-centre. But if you think the definition for `<path>` should change, then that for `<zone>` should change too.
2. I have no idea why anyone would use `<zone>` or `<path>` within `<line>`, but I think that's a question for the genetic editing people. I assumed that if they have a use for the former, they may have a use for the latter. If in fact it's an error that `<path>` is allowed in `<line>`, then the same error pertains to `<zone>`.
3. Good idea, I'll do it now.
4. I think that is an error, but I see how it arose; IIRC the original scenario for `@points` was that it would be accompanied by `@ulx`, `@lry` and friends, which would provide a rectangular bounding box for processors that could not handle non-rectangular shapes. However, the bounding box is trivial to calculate, so that's redundant. We should remove that bit of text completely, because in actual fact the shape may cover areas outside the bounding box of the parent element; it just uses the coordinate matrix of the parent element, and need not be confined within it.
5. Yes please!

martinascholger:
1) Ah, I see. I just didn't check the definition for `<zone>`. I think we can leave the description as it is since we usually don't mention every element that is a child of another element. 
2) I was just curious. Maybe this would be a case? 
`<line ulx="" uly="" lrx="" lry=""><del>Bauen</del><path points="" />`: https://gams.uni-graz.at/iiif/o:sker.01/IMAGE.014/full/full/0/default.jpg 
3) --
4) Then the definition for `@points` in `att.coordinated` (and accordingly `@points` for `<path>`) should be: "identifies a two dimensional area by means of a series of pairs of numbers, each of which gives the x,y coordinates of a point on a line enclosing the area." If you agree, I'd open a ticket. 
5) These are the German translations: 
for gloss: Pfad
for desc: definiert eine beliebige Linie, die durch zwei oder mehr Punkte innerhalb eines <surface>-Elements verläuft.
for `@points`: beschreibt eine Linie innerhalb eines Begrenzungsrahmens, der durch das Elternelement definiert ist, über eine Reihe von zwei oder mehr Zahlenpaaren, die jeweils die X- und Y-Koordinaten eines Punktes dieser Linie angeben.

martindholmes:
Thanks @martinascholger. #2 looks like an excellent example; do you think it would be worth including it? I agree with your definition on #4. Should I put your translations into this branch?

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -347,19 +348,40 @@ used to define a polygon of any shape using this coordinate system:-->
...
+            <head>Smaller detail of p 49r from Bovelles highlighting two specific lines</head>
+         </figure>
+         <egXML xmlns="http://www.tei-c.org/ns/Examples">
+            <surface ulx="0" uly="0" lrx="443" lry="272">
+               <graphic url="facs-fig3.jpg"/>
+               <path points="74,73 171,244" xml:id="balan"/>
+               <path points="71,203 173,116" xml:id="dindan"/>
+            </surface></egXML>
+         
+         This is useful in cases where you need to link an annotation or explanation to a specific line on
```
raffazizzi: I would recommend avoiding using 'you' in the Guidelines. Change this to something like "This is useful for linking an annotation or explanation to..."

P5/Source/Specs/path.xml
```diff
@@ -0,0 +1,81 @@
...
+        path.
+      -->
+        <sch:let name="firstPair" value="tokenize( normalize-space( @points ), '&#x20;')[1]"/>
+        <sch:let name="lastPair"  value="tokenize( normalize-space( @points ), '&#x20;')[last()]"/>
+        <sch:let name="firstX"  value="xs:float( substring-before( $firstPair, ',') )"/>
+        <sch:let name="firstY" value="xs:float( substring-after( $firstPair, ',') )"/>
+        <sch:let name="lastX"   value="xs:float( substring-before( $lastPair, ',') )"/>
+        <sch:let name="lastY"  value="xs:float( substring-after( $lastPair, ',') )"/>
+        <sch:report test="$firstX eq $lastX  and  $firstY eq $lastY">The first and
+          last elements of your path are the same. If you want to specify a closed polygon, use
```
raffazizzi: Perhaps less important in schematron reports, but I would still prefer something slightly more formal, e.g.: "To specify a closed polygon, use the zone element instead"

P5/Source/Specs/path.xml
```diff
@@ -0,0 +1,81 @@
...
+        <graphic url="facs-fig3.jpg"/>
+        <path points="74,73 171,244"/>
+        <path points="71,203 173,116"/>
+      </surface>
+    </egXML>
+  </exemplum>
+    
+    <remarks versionDate="2018-04-20" xml:lang="en">
+      <p>The first and
+        last elements of your path are the same. If you want to specify a closed polygon, use
```
raffazizzi: Ditto

raffazizzi:
The review mechanism is super cool.

P5/Source/Guidelines/en/PH-PrimarySources.xml
martindholmes: I don't mind making that change, but to be honest I see nothing wrong with using "you" like this. What's the basis of the objection? Is there something in our styleguide about this?

P5/Source/Specs/path.xml
martindholmes: Again, no problem with the change, but I do tend to think that human-friendly language is more helpful than excessive formality.

P5/Source/Specs/path.xml
martindholmes: Ditto. :-)

martinascholger:
2) example: If you or anyone else thinks this might be useful, feel free to use it. I guess it might be too specific?
4) yes, please add the translations. I submitted a ticket for a clarification of `@points` in `att.coordinated `(#1789). If we change that accordingly, the German translation for `@points` in `<path>` is: beschreibt eine Linie über eine Reihe von zwei oder mehr Zahlenpaaren, die jeweils die X- und Y-Koordinaten eines Punktes dieser Linie angeben.

lb42:
I suggest https://goo.gl/images/b3q8dV

martinascholger:
It seems to me that we are close to finishing this?

martindholmes:
@martinascholger I'd like to incorporate your image or Lou's or both; not sure yet. But what we have so far can definitely be merged for this release, and then I can look more closely at the images.

martinascholger:
Great! Do you want to merge it yourself, @martindholmes ?

Issue #1765, opened 2018-04-09
Add figure to att.written


StefanDumont:
In the ["edition humboldt digital"](https://edition-humboldt.de) we sometimes have the case that someone other than the author inserts drawings in a manuscript (e.g. the recipient of a letter). This should, of course, be indicated. The attribute `@hand` (att.written) would be suitable, but it is not allowed in `figure` and its definition currently explicitly refers to "textual content". Nevertheless, it seems to me to make sense to define `figure` here as a class member. What does the TEI community think about it?

lb42:
You could use @resp I suppose

StefanDumont:
As far as I understand (correctly?) the [documentation](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/PH.html#PHHR), `@resp` is to encode the responsibility for the mark-up itself. That's how we used `@resp` already...

martindholmes:
+1 from me. Make `<figure>` a member of att.written, and tweak the definition of `@hand` so that it allows for drawing etc. as well as just writing.

sydb:
I like the idea, and my instinct is it should be implemented. I’d like to give someone who knows more about use of `<handNote>` in actual mss description the chance to object, first.

scstanley7:
@sydb are you thinking that the definition of `<handNote>` would need to be similarly modified if this change were implemented? I wouldn't want `<handNote>` to be an impediment to this solution!

StefanDumont:
@sydb Do you know somebody with `<handnote>` knwoledge, who can have a look on this issue? This would be great. And then, what are the next steps? I would be pleased to provide a pull request with the changes ...

lb42:
The desc of handNote doesn't preclude the use for @hand proposed here. I think the only change needed is to the desc for att.wrtten, which currently reads "provides an attribute to indicate the hand in which the textual content of an element was written in the source being transcribed". Delete the word "textual", make figure a member of att.written,  and you're all set.

sydb:
@scstanley7 — Yes, I suppose the definition of `<handNote>` should be tweaked to make this use explicit. This would probably mean a bit of added prose in [10.7.2 Writing, Decoration, and Other Notations](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/MS.html#msph2), and to the definition of the element. But whether that latter should go in `<desc>` or in `<remarks>` is not clear to me at the moment.

[...]

PR #1777, opened 2018-06-15
Add figure to att.written (see #1765)

StefanDumont:
This pull request handles the feature request described and discussed in ticket #1765 to allow @hand also in figure. Feedback, improvement requests etc. are very welcome. I think especially the question, what parts of the Guidelines should be updated, needs more discussion.


peterstadler:
Just tested locally (`make test`) without any problems …

jamescummings:
Looks good to me.

PR #1808, opened 2018-08-31
Make cleanup

duncdrum:
ran make and make clean and all seems to be in order

the jTei error is unrelated
```bash
[runjing] /tei/TEI/P5/Exemplars/tei_jtei.odd:2957:177: error: element "xsl:key" not allowed here; expected the element end-tag or text
```

### DIFF
--- a/P5/.gitignore
+++ b/P5/.gitignore
@@ -1,1 +1,2 @@
 local.mk
+repodate.xml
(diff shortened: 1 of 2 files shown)


duncdrum:
if the oNVDL files need to be downloaded and commited on every run of make, feel free to cherry-pick the first commit only.

peterstadler:
Thanks @duncdrum !

PR #1818, opened 2018-09-12
[jTEI] improved Schematron checks for cross-reference pointers

rvdb:


rvdb:
Thanks for the review, @sydb . Concerning the attribute checks: in a Schematron context with `querybinding="xslt2"`, it would be perfectly fine to have:

    <sch:rule context="(@corresp|@target|@from|@to|@ref|@rendition|@resp|@source)[not(ancestor::eg:egXML)][some $i in tokenize(., '\s+') satisfies starts-with($i, '#')]">

Yet, when I try to validate a document against a RelaxNG XML schema containing this rule as embedded Schematron, Oxygen returns following errors:

    Failed to compile stylesheet. 1 error detected.
    Error in expression (tei:author|tei:title)[ancestor::tei:teiHeader]: Unexpected token in pattern, found (

This indicates that the `@context` attribute in the resulting RelaxNG schema is interpreted as an XSLT1 expression, which indeed prohibits parentheses. 

Here's my question: is there a way to specify `querybinding="xslt2"` somewhere for Schematron rules embedded in RelaxNG (and if so, how can this be done in ODD)?

sydb:
Whoa! **That** should not be happening. (But then again, as I have pointed out for eons, the Schematron you get in the .rng output of the normal build process (e.g., teitorelax) is not at all the same as the Schematron you get in the .isosch output of extract-isosch; the latter does have a query binding of XSLT2, although it is hard-coded in the XSLT.) Thank you very much for the report. We need a new ticket (against the Stylesheets repo) for this. I plan to create one myself in an hour or two, unless you already have.

sydb:
Thank you _very_ much for your research on this, @rvdb.

I think avoiding `ParenthesizedExprP` (and other XSLT 3.0 constructs, if any) for now is a good idea. What to do in the big picture is tied up with what we do for the oXygen framework in the big picture, I guess.

PR #1820, opened 2018-09-17
[jTEI] removed XSLT 3 expressions

rvdb:
In theory the ODD now generates schemas that are backward-compatible to Oxygen-15.

(except that xsl:key inside schematron seems only to be supported from Oxygen-17 onward, for some reason)

(addresses remaining issues from https://github.com/TEIC/TEI/pull/1818#issuecomment-420638069)

### DIFF
--- a/P5/Exemplars/tei_jtei.odd
+++ b/P5/Exemplars/tei_jtei.odd
@@ -2755,7 +2755,7 @@
             <constraintSpec ident="jtei.sch-rendition" scheme="schematron">
               <constraint>
                 <sch:rule context="tei:rendition">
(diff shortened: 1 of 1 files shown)


rvdb:
@sydb thanks for the merge, and your thoughts. Apparently, non-capturing groups have only been [introduced](https://www.w3.org/TR/xpath-functions-30/#substantive-changes-2009-12-15) in XPath 3.0. (This had gone unnoticed to me before, since the latest Saxon processor didn't complain about XSLT 3.0 expressions, even when executed in an XSLT 2.0 context.)

In this commit, I have removed the non-capturing subgroups (by removing the XSLT 2.0-incompatible `?:` syntax from the parentheses), and indeed consequently had to update the capturing group reference numbers.

Issue #1797, opened 2018-07-23
More examples of `<path>` needed


martindholmes:
In issue #1453, I completed the basic work, but @martinascholger and @lb42 suggested some good example page-images worth including, and I didn't get around to doing that before the merge was done for the latest release. This ticket is for me to follow up after the release and do a better job of the examples.

martindholmes:
Working on an example based on the Sterne page now; I think that should be enough. I'm waiting for permission to use part of an image from Cambridge U as the example.

martindholmes:
New branch mdh_issue_1797 with a more complicated example now created. Pull request coming.

peterstadler:
Pull request merged, closing the ticket.
Thanks @martindholmes !

PR #1839, opened 2018-11-19
Mdh issue 1797: more complex example for `<path>` element

martindholmes:
As detailed in issue #1797, I've added a new more complex example of the `<path>` element. A build of the PH chapter shows it directly following the example here:

https://hcmc.uvic.ca/~mholmes/tei/issue_1797/PH.html#index-egXML-d52e92324


P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -380,7 +380,42 @@ used to define a polygon of any shape using this coordinate system:-->
             </surface></egXML>
          
          This is useful for linking an annotation or explanation to a specific line on
-         an object surface. Any number of coordinates can be included to specify lines which are not straight.
+         an object surface. Any number of coordinates can be included to specify lines 
+         which are not straight; this example shows the how the first of the famous
```
peterstadler: there is a double "the" methinks …
```suggestion
         which are not straight; this example shows how the first of the famous 
```

peterstadler:
is the zip file `sterne.zip` really necessary?

martindholmes:
@peterstadler TCW 20 says:

> If you use an image in your Guidelines change, you will need to add it to the git repository in the P5/Images directory. If you have asked for and received permission from a rights-holder to use the image, include all of the relevant correspondence in a zip file named the same as the image file (so for an image fred.png, include permission documents in a file called fred.zip).

That's why the sterne.zip file is there. We should have them for all images which are not public property.

martinascholger:
I like the example!

peterstadler:
> > If you use an image in your Guidelines change, you will need to add it to the git repository in the P5/Images directory. If you have asked for and received permission from a rights-holder to use the image, include all of the relevant correspondence in a zip file named the same as the image file (so for an image fred.png, include permission documents in a file called fred.zip).
> 
> That's why the sterne.zip file is there. We should have them for all images which are not public property.

Ah, thanks for the explanation. I didn't know about that!

peterstadler:
I think that can be merged, assigning @martinascholger to double check :)

martinascholger:
Already approved :-) This can be merged.

PR #1888, opened 2019-06-19
various small improvements to Schematron constraints

rvdb:


rvdb:
Thanks, @ebeshero , I've noticed my first commit contained some overly complex tests in two Schematron constraints, which I've greatly simplified in a following commit.

P5/Exemplars/tei_jtei.odd
```diff
@@ -2525,19 +2534,17 @@
                   </sch:report>
                 </sch:rule>
                 <sch:rule context="tei:note" role="warning">
-                  <sch:report test="not(@rend='inside.parenthesis') and following::text()[not(ancestor::tei:note)][1][matches(., '^\)')]" sqf:fix="note.rend.add">
+                  <sch:report test="following::text()[not(ancestor::tei:note)][1][matches(., '^\)')]">
```
peterstadler: Since you're checking for the beginning of the string, does it make sense to normalize-space it first?

P5/Exemplars/tei_jtei.odd
```diff
@@ -2525,19 +2534,17 @@
...
                     Footnotes may be placed before closing parentheses, though this is
-                    exceptionial. If this note's placement is intentional, please mark
-                    this with a @rend='inside.parenthesis' attribute. Otherwise, move
-                    it after the closing parenthesis.
+                    exceptionial. Please check if this note's placement is correct.
```
peterstadler: That should be "exceptional".

P5/Exemplars/tei_jtei.odd
```diff
@@ -2525,19 +2534,17 @@
...
                   </sch:report>
-                  <sqf:fix id="note.rend.add">
-                    <sqf:description>
-                      <sqf:title>Add a @rend='inside.parentheses' attribute to the &lt;note&gt;.</sqf:title>
-                    </sqf:description>
-                    <sqf:add match="self::tei:note" node-type="attribute" target="rend">inside.parenthesis</sqf:add>
-                  </sqf:fix>                  
                 </sch:rule>                
+              <sch:rule context="tei:note">
+                  <sch:assert role="warning" test="matches(normalize-space(), '[.?!]$')">
```
peterstadler: Don't you need to escape those characters, i.e. `[\.\?!]`?

P5/Exemplars/tei_jtei.odd
```diff
@@ -2969,7 +2976,9 @@
                 <sch:let name="apos.straight" value="''''"/>
                 <sch:let name="quotes" value="concat('[', $apos.straight, '&quot;]')"/>
                 <sch:let name="div.types.front" value="('abstract', 'acknowledgements', 'authorNotes', 'editorNotes', 'corrections', 'dedication')"/>
-                <sch:let name="tei.version.url" value="'http://www.tei-c.org/release/doc/tei-p5-doc/VERSION'"/>
+                <!-- http://www.tei-c.org/release/doc/tei-p5-doc/VERSION produces 
+                     429 "Too Many Requests" errors -->
+                <sch:let name="tei.version.url" value="'https://jenkins.tei-c.org/view/TEI%20master/job/TEIP5/lastStableBuild/artifact/P5/release/doc/tei-p5-doc/VERSION/*view*/'"/>
```
peterstadler: Better strip off the 'view/TEI%20master'-part: `https://jenkins.tei-c.org/job/TEIP5/lastStableBuild/artifact/P5/release/doc/tei-p5-doc/VERSION/*view*/`

P5/Exemplars/tei_jtei.odd
```diff
@@ -3262,26 +3271,26 @@
           </constraintSpec>
           <constraintSpec ident="jtei.sch-teiVersion" scheme="schematron">
             <constraint>
-              <sch:rule context="@target[matches(., '^https?://www\.tei-c\.org/release/doc/tei-p5-doc')]">
+              <sch:rule context="@target[matches(., '^https?://(www\.)?tei-c\.org/release/doc/tei-p5-doc')]">
```
peterstadler: probably nitpicking, but you might want to escape the dashes in the regex as well?

P5/Exemplars/tei_jtei.odd
rvdb: Sure, thanks. I'll change that.

P5/Exemplars/tei_jtei.odd
rvdb: Doh, well spotted!

P5/Exemplars/tei_jtei.odd
rvdb: Not necessarily: they are interpreted as normal characters (only `\`, `[`, and `]`, and -depending on the position- `^` and `-` have to be escaped inside a character class). All other characters are treated as literal characters inside a character class. I even thought they shouldn't be escaped inside a character class, but https://www.w3.org/TR/xmlschema-2/#nt-charClassEsc seems to explain that the usual meta-characters can occur escaped inside a character class as well. In other words: `[?]` and `[\?]` are equivalent.  I prefer `[.?!]` since I think it is more readable and less confusing than `[\.\?\!]` (with non-necessary character escapes). Unless there is a general preference for the latter expression?

P5/Exemplars/tei_jtei.odd
rvdb: Thanks, I see it works even without the `/*view*/` part, so I'll change to https://jenkins.tei-c.org/job/TEIP5/lastStableBuild/artifact/P5/release/doc/tei-p5-doc/VERSION.

P5/Exemplars/tei_jtei.odd
rvdb: Ah, I wasn't even aware the hyphen could be escaped (except inside a character class), since it's interpreted as a literal in this context. Again, `-` seems more readable (and less confusing than a non-necessary character escape) here, unless you disagree.

P5/Exemplars/tei_jtei.odd
peterstadler: No, if they don't need to be escaped, the shorter the better :) 
Thanks for the clarification, I didn't look that up

P5/Exemplars/tei_jtei.odd
peterstadler: you're right, again!

rvdb:
@peterstadler Thanks for your suggestions, I've applied the remaining ones in a new commit.

PR #1890, opened 2019-06-24
Testing odd2json output in multiple langs

raffazizzi:
Modified p5subset.json output to generate separate files for supported languages so that they will be in the Vault. 

Tested successfully locally.

### DIFF
--- a/P5/antbuilder.xml
+++ b/P5/antbuilder.xml
@@ -55,13 +55,40 @@
     <delete dir="Schema"/>
     <mkdir dir="Schema"/>
     
-    <xslt processor="trax" force="yes" style="${XSL}/odds/odd2json.xsl" in="p5subset.xml" out="p5subset.js">
-      <factory name="net.sf.saxon.TransformerFactoryImpl"/>
-      <param name="callback" expression="teijs"/>
-    </xslt>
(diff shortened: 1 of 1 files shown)


P5/antbuilder.xml
```diff
@@ -55,13 +55,40 @@
     <delete dir="Schema"/>
     <mkdir dir="Schema"/>
     
-    <xslt processor="trax" force="yes" style="${XSL}/odds/odd2json.xsl" in="p5subset.xml" out="p5subset.js">
+    <xslt processor="trax" force="yes" style="${XSL}/odds/odd2json.xsl" in="p5subset.xml" out="p5subset.json">
       <factory name="net.sf.saxon.TransformerFactoryImpl"/>
-      <param name="callback" expression="teijs"/>
```
peterstadler: Just to make sure: that `callback` param is not used anywhere (thus dropped in your PR)?

raffazizzi: Yes, this is a remnant of the first version of odd2json, which wrapped the data in a callback, JSONP style. This parameter has been dropped for a while.

peterstadler: Thanks for the clarification!

PR #1892, opened 2019-06-26
Measurement

### LINKED ISSUE #1707: How to encode measurement
I'm Naoki who gave a poster presentation at the TEI 2017 at UVic, on how to mark up **UNITS** that were not based on the metric system. It should be important for us to discuss measurement in a broad sense, because the problem on measurement implicates and represents cultural diversity. [A Wikipedia article](https://en.wikipedia.org/wiki/Imperial_and_US_customary_measurement_systems) may help us to understand the importance of discussing cultural uniqueness through measurement.

And, some of the members of TEI have already discussed how useful \<unit\> element is. Please see this https://github.com/TEIC/TEI/issues/1461

First of all, I would like to share my revised version of poster which omits the image of the historical source due to the license.
[TEI_2017_poster_for_github.pdf](https://github.com/TEIC/TEI/files/1476083/TEI_2017_poster_for_github.pdf)

The problem is that:
how should it be marked up?

**銅二千五百十六斤十両二分四銖**

which means 'Copper whose weight is 2516Kin, 10Ryo, 2Bu and 4Shu'.
It might be complicated to see what's happening, but you might understand it by considering some examples of British measurement, like 12yd, 2ft and 10in, which were also not based on the metric system.

Based on the discussion within the conference, there might be at least two solutions.

(i) Using only one \<measure\> element:
`<measure` commodity="銅" n="2516/10/2/4" unit="斤/両/分/銖" />

In current scheme, we can't use \@quantity to store multiple values.
Though, we can use any delimiter instead of using slash.

(ii) Using \<measure\> element to nest the other \<measure\> elements.
...Sorry, I have to get on the plane. Please see the poster file to find a second of the possible solutions!

ebeshero:
I've just asked about half of Council to help review this pull request and strongly recommend that more than one of us reviews it carefully. This measurement branch contains commits to resolve #1707 , introducing `<unitDecl>`, `<unitDef>`, a `<conversion>` element, and special attributes, including a new attribute class for `@formula`. This has been a long time coming and there's a lot of new documentation to review here in the TEI Header chapter! I hope I've written it all clearly, and expect that we'll need to make some changes. 

 I'll be very happy if we can merge it into dev: My changes don't seem to be introducing errors (From what I see, Paderborn Jenkins is just showing the same error with simplePrint that dev branch is experiencing: Here are the results: https://jenkins.tei-c.org/job/TEIP5-branch-measurement/lastFailedBuild/parsed_console/ ). If you approve these changes, we'll be ready to introduce a whole new set of tools for defining measurements and conversions introduced to us in Victoria and thoroughly refined and developed in Tokyo!


ebeshero:
Travis was faulting me for using `#chalder`, `#keel`, and #bushel` multiple times in our unitDecl conversion example, and I saw I used the wrong datatype. I've just pushed a correction.

ebeshero:
I'm not sure I understand the errors reported by Travis about `att.duration.iso`? I just made the `<conversion>` element be a member of `att.datable` not a member of `att.duration` so I'm confused...

ebeshero:
A question for whoever reviews this: Should we or should we not permit `<unit>` inside `<unitDef>`? Right now I've permitted it, but I think from reviewing the ticket that maybe we don't want it there.

And I expect this will look okay on Jenkins and not pass on Travis: I don't know what to do about its calling me out on the values for `@fromUnit` and `@toUnit`: I'm pretty sure the teidata.pointer datatype is not to blame...Thanks for any help here.

P5/Source/Specs/charDecl.xml
```diff
@@ -52,4 +52,4 @@ $Id$
   <listRef>
     <ptr target="#D25-20"/>
   </listRef>
-</elementSpec>
+</elementSpec>
```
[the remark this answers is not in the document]
ebeshero: I wonder if it came in when I was rebasing this branch...I'll see about reverting it.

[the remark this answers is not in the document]
ebeshero: And I'll see about altering the examples--I was worried that the issue was the use of keel, chalder, and bushel twice in the *same* example. (I like that example a lot because it shows why we'd want to use `unitDecl` in the first place: it shows differences in these units of measure over time periods.) I'll try to come up with something different for the others so we have some variety.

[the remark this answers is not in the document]
martindholmes: Naoki has some cool Japanese examples from his original work. Wouldn't it be nice to include them?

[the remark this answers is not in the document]
ebeshero: Yes—the code needs to be updated, though, and they don’t demonstrate the use of `@formula`. I was thinking about those too. The example I am featuring seems to be one of his too, though using English units.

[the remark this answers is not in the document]
ebeshero: Reverted charDecl.xml to its former state. Now looking into the examples...

ebeshero:
I have provided a wider range of international and historical examples of measurement unit declarations in these last commits, and I hope these may be merged into dev. But I'm sorry to say these builds are still failing, and if there's more to do on this branch that wouldn't be better to do in dev, I need some help to figure out what it is. Note that this branch has not been updated with dev's most recent changes this weekend--I attempted a rebase that didn't work out well (nearly lost several of my edits).

peterstadler:
No need to update from dev branch right now because there are no conflicts with the that branch.

Those tests still fail for measurement issues, see https://travis-ci.org/TEIC/TEI/builds/552582571

ebeshero:
@peterstadler Okay, I'm stumped here, and I'm sorry if this is something perfectly obvious. I don't understand what's causing the build failures on Travis. 
* One cluster of errors on the HD chapter SpecGrp entries for UnitDecl complains that I'm pointing to a nonexistent attribute and attribute class  (about `@formula`, on `att.formula`). But that is the attribute class and attribute that is most definitely present in Specs with this branch. What have I failed to do in order to introduce this attribute class and attribute? I'm pointing to it in the SpecGrp and that's where the error is raised. What else can I be doing to properly introduce an attribute class and its attribute? 
* A warning on [the new `<conversion>` element spec](https://github.com/TEIC/TEI/blob/measurement/P5/Source/Specs/conversion.xml) is a Schematron warning about the deprecation of empty content. But that warning seems to be for an empty `<content/>` element, which I do not have here. I don't know what to do about this, but I don't think this warning should apply here, because I'm using the `<empty/>` element.
```
    <WARNING>Schematron warning: The empty-content-deprecated construct becomes outdated on 2019-08-25 [Test: @validUntil cast as xs:date ge $advance_warning_period] Location: /TEI[1]/text[1]/body[1]/div[22]/div[6]/specGrp[1]/elementSpec[1]/constraintSpec[1]</WARNING>

``` 

* There is another persistent cluster of errors on the `ST` chapter about the supposed nonexistence of att.duration.iso. But `ST` was not edited in this branch. In case it was a rebase issue, I even copied in the ST chapter directly from dev just in case there was something problematic on the branch version of it. I saw no changes, so I don't think that was the problem. `att.duration.iso` is most certainly present. 

These are all errors or warnings that I do not understand. I need help with these in order to make progress on introducing `<unitDecl>` and its associated elements and attributes so we can resolve #1707. I'm sorry about my confusion if these are things that are obvious and ought to be easy to correct, but I'm really stuck here.

duncdrum:
@ebeshero I see the same `att.duration` test failures, i believe it has something to do with `detest.log` needing to be updated, but of course `make test` is automatically removing the newly generated `detest.log` to keep us on our toes.

ebeshero:
@peterstadler Okay! I found an `<include>` I'd missed in `HD` and repairing that seemed to resolve a bunch of errors (including, weirdly, the ones on `ST`. But I'm mystified about the persistent Schematron warning on empty `<content>` since I don't have that. Take a look, is the branch ready to move to `dev`?

duncdrum:
@ebeshero i find it very curious that both of us see the same travis failures unrelated to the code we are working on. Something about the `att.xxx` spec seems to be throwing a wrench into things.

P5/Source/Specs/unitDecl.xml
```diff
@@ -0,0 +1,47 @@
...
+    <memberOf key="att.datable"/>
+    <memberOf key="att.global"/>
+  </classes>
+  <content>
+    <elementRef key="unitDef" minOccurs="1" maxOccurs="unbounded"/>
+  </content>
+  <exemplum xml:lang="en">
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
+        <unitDecl>
+          <unitDef name="斤" xml:id="斤" type="weight">
```
peterstadler: These `@name` attributes are not declared in the elementSpec of `<unitDef>` (thus breaking the build).

ebeshero:
Aha--thanks for spotting that, @peterstadler: This was one of Naoki's examples I adapted from the ticket. I was about to supply `@name` to accommodate it, but now that I look at the example I'm not sure we really need that `@name` attribute...What do others think? We have a label (or model.descLike) element and an `@xml:id` and I think those should be sufficient, right? I think I'll just remove the `@name` now--but if anyone sees a need for it, we'll add it.

ebeshero:
Okay--maybe *this* time measurement branch will pass... 
@duncdrum I'm watching Travis now, but I think the mysterious error about the unrelated Guidelines chapter and attribute class (for me it was att.duration.iso in `ST`) went away when I fixed a real error in the actual chapter I was modifying (`HD`) where I'd forgotten to the `<include>` for `@formula`... I'm hoping with this last push I'm clear, but apparently Travis is *still* unhappy...

ebeshero:
@peterstadler Well, I'm logging experience in cryptic error messages, anyway. I removed `@name` from the `<egXML>` on the unitDecl spec, and then started seeing a new kind of error message on Travis about `att.lexicographic`. I figured I'd done something wrong there--needed something more in `HD` to indicate that `unitDef` is a member. I tried adding specDesc and specGrp include lines but that didn't help anything, so I returned `HD` to the way it was when I just edited out `@name` from the example code. 

On closer inspection of this persistent error message, it seems to be coming in a lot later in the build testing, right after the test to see whether the TEI website is up and running:
```
check that TEI web site is up and responding
TEISITE=`curl -s --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml`
BUILD: Testing files which demonstrate user extensions
xmllint --noout --valid extns.xml
../DTD/header.dtd:815: validity error : PEReference: %att.lexicographic.attributes; not found
 %att.lexicographic.attributes;
                               ^
Makefile:84: recipe for target 'test-extns' failed
make[1]: *** [test-extns] Error 4
make[1]: Leaving directory '/var/tei/P5/Test'
Makefile:197: recipe for target 'test' failed
make: *** [test] Error 2
```

I can't think of anything more to do here after experimenting--I mean, I could remove `<unitDef>` from `att.lexicographic`, but a) I wanted it to be a member for good reason and haven't heard any objections, and b) I don't think the unitDef spec is really the problem. So I'm going to stop experimenting and wait for another look.  What's going on with the att.lexicographic attributes here, and are we any closer to merging and getting rid of this old branch?

duncdrum:
@ebeshero yup i have to agree the error messages are next to useless in helping you to actually find the source of the problem. 

I ll say it one more time, build tools with test runners, build phases, test reporters, and dependency checking all build in do exist.

ebeshero:
WHEW! I can scarcely believe it, but I believe this is *really* ready to merge into dev now. Thanks for bearing with me in preparing all the code and examples for #1707. @peterstadler

peterstadler:
I just updated from current dev to include the build-fixing changes. Hopefully we'll be able to check your changes at https://jenkins.tei-c.org/view/various%20branches/job/TEIP5-branch-measurement/ soon

ebeshero:
@martinascholger Can you clarify about which description/which files? (There are a lot of `<desc>` elements in the egXMLs here). Are you saying that we're to be using `<gloss>` instead of `<desc>` inside `<unitDecl>`?

P5/Source/Specs/conversion.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?><!--
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+--><?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?><elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="header" ident="conversion">
+  <gloss versionDate="2019-06-25" xml:lang="en">describes a mathematical conversion from one system of measurement to another.</gloss>
```
martinascholger: The description is provided within the `<gloss>` element instead of `<desc>`;

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -1782,8 +1784,53 @@ more than one language, as in the following example:
...
+          <conversion fromUnit="#chalder" toUnit="#keel" formula="20" from="1421" to="1676"/>
+          <conversion fromUnit="#chalder" toUnit="#keel" formula="16" from="1676" to="1824"/>
+          <desc>Keel was a unit measuring weight of coal. It had been equal to 20 chalders from 1421 to 1676, and it was made to be equivalent to 16 chalders from 1676 to 1824.</desc>
+        </unitDef>
+        <unitDef xml:id="chalder" type="weight">
+          <label>chalder</label>
+          <placeName ref="#england"/>
+          <conversion fromUnit="#bushel" toUnit="#chalder" formula="32" from="1421" to="1676"/>
+          <conversion fromUnit="#bushel" toUnit="#chalder" formula="36" from="1676" to="1824"/>
+          <desc>Chalder was a unit measuring weight of coal. It had been equal to 32 bushels from 1421 to 1676, and it was made to be equivalent to 36 bushels from 1676 to 1824.</desc>
```
ebeshero: @martinascholger Are you referring to `<desc>` elements like this one?

P5/Source/Specs/att.formula.xml
```diff
@@ -0,0 +1,46 @@
...
+<!--
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" type="atts" ident="att.formula">
+  <desc versionDate="2019-06-25" xml:lang="en">provides attributes for defining a mathematical formula.</desc>
```
ebeshero: @martinascholger Or does this `<desc>` need to be a `<gloss>`?

martindholmes:
It's a bit late to be saying this, perhaps, but `@formula` doesn't look right to me. If it's simply a number by which fromUnit is to be multiplied in order to get its equivalent in toUnit, then it should be called "multiplier". But if it's intended to hold an actual formula, then its datatype is too simple. I may be misunderstanding because I've missed Council discussions, though.

ebeshero:
@martinascholger Thanks--I just pushed up a revision to convert that `<gloss>` to a `<desc>` for `<conversion>`, and I rephrased it, too: https://github.com/TEIC/TEI/pull/1892/commits/67633073c8c69d2d5a28faa70ff5011ef2d2e171
 
@martindholmes I was thinking the same thing as I was working on the ticket. The examples I have are all about augmentation--a multiplication factor, which did seem to vary historically. I kept the datatype simple as I was working and also because I wasn't sure how we'd want to express something like a conversion between Fahrenheit to Celsius which needs a real formula rather than just a factor. But we did all agree in Tokyo to use `@formula`, as described on the ticket. I would like there to be a richer range of values here, but I wasn't sure how to define it. If I remember correctly, when we spoke of more complex processing the idea was that we're supplying a numerical value that someone can use in an XSLT process if they want to convert between units. But can we be storing a complex mathematical formula in our data model? How is that expressed?

P5/Source/Specs/att.formula.xml
```diff
@@ -0,0 +1,46 @@
...
+  <attList xmlns:rng="http://relaxng.org/ns/structure/1.0">
+    <attDef ident="formula" usage="opt">
+      <gloss versionDate="2019-06-25" xml:lang="en">holds a numeric value designated for use in a mathematical calculation.</gloss>
```
martinascholger: This `<gloss>` contains a description. I think it is not necessary to provide a `<gloss>` in this case.

ebeshero:
Okay-- I pushed up just a change from a `<gloss>` to a `<desc>` on the conversion elementSpec, and the Travis PR build is displeased (but the push build is okay). Any idea what's up? (humph)

P5/Source/Specs/att.formula.xml
ebeshero: @martinascholger So I should change this to a `<desc>` as well? But I might need to modify this anyway--if we're following up on @martindholmes 's idea here. I have been wondering if we want to take a closer look at this new attribute class and attribute anyway...

martinascholger: Also in att.formula.xml: the second sentence of the `<remarks>` contains a word duplication - from from

P5/Source/Specs/conversion.xml
ebeshero: Done! :-)

martindholmes:
@ebeshero My memory from Tokyo is that we had an operator followed by an operand:

*20 
/12

and so on. So all that would be needed would be a tweak to the datatype to give it a regex like:

[\*/\+\-][\d\.]+

(not tested, off the top of my head)

so then you could express multiplication, division, and plus or minus. I can't think of a case that can't be expressed by multiplication, though, so if no-one else can either, I would suggest changing the name. Naoki's original proposal had `@factor`, which is closer to what this is doing; the change to `@formula` must have been prompted by a desire to enable more complex calculations, but if the datatype prevents that, I would suggest `@multiplier` or `@factor`, with `@formula` waiting in the wings until someone comes up with a genuine use-case that can't be handled by a plain multiplier.

ebeshero:
@martindholmes I was just looking at the `teidata.numeric` type that I applied here: it's pretty close, isn't it? 
```
(\-?[\d]+/\-?[\d]+)

```

It looks to me like that permits a ratio like `-9/5` but not something like a multiplication operand, or addition/subtraction etc. If we wanted to do a temperature conversion from Fahrenheit to Celsius we'd want to subtract 32 first, and then multiply by 5/9. (I was messing around with that on my calculator while working on this ticket to see if I could turn that into a single numerical multiplier and my math maybe wasn't good enough... but I'd want to write the formula algebraically anyway, like this:
```
(X - 32) * 5/9 
```
I think that's a little more complicated than what we have in our examples. I wonder if we should go back to `@factor` as you suggest, and hang on to `@formula` until we can write a good datatype for it? I'm worried that's a rabbit hole, though...can we ever figure out the different combinations of factors that can go into the text string that makes a mathematical formula? 

Or should we just make `@formula` be very open-ended as `teidata.text` so people can set what they want here?

ebeshero:
Another note on this: I wonder if we were thinking of making `@formula` do the work of a multiplication factor, *and* potentially be expanded into a more complex actual formula? In which case, what shall we do with our datatype?

martindholmes:
@ebeshero I'd completely missed that fact that F/C temperature conversion is the perfect example for this situation. I would say that any implementation that doesn't handle this specific conversion is inadequate; we should keep `@formula` but it should handle this case, which would make a perfect example. The question is how to represent the incoming value (the fromUnit value) in the equation. I have a vague memory that we said it should be specified as "fromUnit", didn't we? so (fromUnit - 32)*(5/9). That's pretty unambiguous. The regex could be as simple as allowing parentheses, operators, digits, periods and "fromUnit", in any configuration, at least for the moment.

ebeshero:
@martindholmes See, what I did was maybe a bit dodgy: I was thinking of Fahrenheit to Celsius but didn't know what to do about the datatype, so I craftily snuck in a simpler conversion, from Celsius to Kelvin. Hmmm. In that case it's just simple subtraction, but I only included a number. What you're doing is *much* more precise. But should we say:
```
(@fromUnit - 32)*(5/9)
```
Do we ever *do* that in TEI? That is, apply an xpath expression inside an attribute value?

ebeshero:
And would we be inventing a special datatype just for this attribute class? I have no problems with this, as it really does extend our expressive capacity. I've never invented a datatype for the Guidelines before...

martinascholger:
There is a datatype teidata.xpath

martindholmes:
@ebeshero @martinascholger I seem to remember having this exact discussion before, long ago. Did we not do this in Tokyo? I think we even finally settled on the idea of an XPath expression. If we do use `@fromUnit`, then we'll have to decide whether it's enough just to invoke it (in other words, we assume that the context is the `<conversion>` element) or whether we need to do something like `parent::conversion/@fromUnit`, on the assumption that the context is the `@formula` attribute. But I think I prefer the former; there's nothing particularly to gain from the latter as far as I can see.

ebeshero:
@martindholmes @martinascholger I don't think we (I) properly noted it in on the ticket in Tokyo, but as we're sorting this out now, the `teidata.xpath`  datatype seems exactly right, because it is already designed to be able to express relationships (and an XPath *can* just be a number or a simple calculation, too). I think we'll want an `<egXML>` to accompany this, perhaps our Fahrenheit to Celsius conversion after all! Where should it go? On `att.formula` itself? Or in the specs for `unitDef`  or in the TEI Header explanation?

ebeshero:
Also, we should really summarize this on the ticket for #1707, shouldn't we?

martindholmes:
Yes, I think we should summarize on the ticket. The Fahrenheit/Celsius example should go wherever it's most visible, because it's the most universally understandable one; perhaps that's on unitDef. I wonder if we can come up with another compelling one for the header chapter code, though. Perhaps Fahrenheit to Kelvin, although that's basically the same thing.

martindholmes:
Also let's not forget that the XPath division operator is "div", not slash...

sydb:
Sorry, I have not read entire thread. But it sounds like you are looking for examples of unit conversion that are not simple multiplication other than °F↔°C or currency.
1. degrees, minutes, seconds →radians `( $deg + ( $min div 60 ) + ( $sec div ( 60 * 60 ) ) ) div ( 180 div $pi )`
2. gradians →radians `$gons div ( 200 div $pi )`
3. blood pressure →estimated mean arterial pressure `$diastolic + ( ( $systolic + $diastolic ) div 3 )`

duncdrum:
As I recall @sydb  noted security concerns about executing xpath, but `teidate.xpath` seem the way to go for `@formula`

ebeshero:
@sydb and @martindholmes : A question about the datatype! We're talking about using `teidata.xpath` for `@formula` (as I currently just have `teidata.numeric` and we really want this to be more capable of expressing a real formula). Syd, in your examples (thank you!) you are using the `$` to signal a variable, and I remember something like this in our discussion in Tokyo, but wasn't sure what to do with it in defining the attribute datatype. Just now we were talking of a literal XPath inside `@formula` that would explicitly point to the value in `@fromUnit`, maybe just like `formula = (parent::conversion/@fromUnit - 32) * (5 div 9)`. But that looked a little awkward--and we thought maybe `formula = (@fromUnit - 32) * (5 div 9)` would be simpler, but maybe the dollar sign signaling is the better way? `formula = ($fromUnit - 32) * (5 div 9)`. The point, though, is that `@formula` shows you how to get from `@fromUnit` to `@toUnit`, and we'd want to be demonstrating that in the formula. Is the dollar sign the way to go here?

ebeshero:
Also, as we're thinking of more examples, remember that key to `unitDecl` is the curation of *nonstandard* unit data that we can't easily look up via ISO. I think it's okay to have a few examples of standard conversions maybe for `@formula` itself though.

sydb:
- I love the idea of using XPath
- I do not now remember why in Tokyo I mildly preferred variables
- BUT, the paths do get a bit complex, no?
  - Won’t every attribute value need to be cast into a number, e.g.? (Unless you know you are using XSD and a Schema-aware processor, right?)
  - If the datatype of the attribute being used in the formula is teidata.numeric, you have the problem that a number might be expressed as "22/7", which would lead to XPath failures
- AND, more importantly, processing an XPath is very scary stuff. It is perfectly reasonable for you to process the XPaths on `@formula` attributes in your own project by just evaluating them. But if you are accepting files that have `@formula` from outside your own system (as an archive, repository, or service like TAPAS might), you need to thoroughly examine the XPath for security attacks before you execute it. Doing so is a bit over my head.

ebeshero:
@sydb I'm thinking that if we're wanting to suggest people can enter a number, a ratio (5 div 9), or a formula, the teidata.xpath is the right datatype for `@formula` because it's the most flexible. However, it's not without caveats that we have to carefully document and exemplify: As long as you type in `5 div 9` you're okay, but no `5/9`. Entering a number by itself should be just fine, so factors will work. Referring to an attribute by name on the same element and casting it to a number should be okay, just a little verbose. I think a point of confusion I have about this is whether we *need* to cast to a number or if we're really using the XPath to express the formula--and someone else writing XSLT or XQuery to process a file with a `<conversion>` element in its `<unitDecl>` up the TEI Header will be the ones to cast it to a number and use it for calculations. If that's the case, we don't need to do the casting, do we?

ebeshero:
If we use teidata.xpath for our datatype, it sounds like we'll want to include a caveat intended for people working with the code--something along the lines of, "Be very careful with this, undersatnding others processing your XPath expose themselves to security vulnerabilities." But I don't think that should be a reason against describing a relationship between variables using XPath, should it?

sydb:
> Also, as we're thinking of more examples, remember that key to `unitDecl` is the curation of _nonstandard_ unit data that we can't easily look up via ISO. I think it's okay to have a few examples of standard conversions maybe for `@formula` itself though.

Could always exemplify conversion of [FFF](https://en.wikipedia.org/wiki/FFF_system) units.

- furlongs per fortnight →m/s:  `@unitFrom cast as xs:decimal * 0.0001663`
- furlongs per fortnight →miles per hour: `@unitFrom cast as xs:decimal * 0.000372`

When I was a lad my father was one of the mathematicians who made up problems for the local high school math leauge. One problem was to calculate the exact time at which the minute hand and the second hand coincide for the first time after, say, 13:00. At my suggestion he made the problem harder (or easier, depending upon the disposition of the contestant) by using a clock that measured not hours and minutes, but rather decidays (2.4 hours or 2 hours 24 mins or 144 mins) and centidays (0.24 hours or 14.4 minutes or 14'24").

- decidays → hours: `@fromUnit cast as xs:decimal * 144 div 60`
- minutes → decidays: `@fromUnit cast as xs:decimal div 144`
- centidays → seconds: `@fromUnit cast as xs:decimal * 0.24 * 60 * 60`

I work in Boston, so personally I love the [smoot](https://en.wikipedia.org/wiki/Smoot). But for TEI purposes I think it is a bit US-centric, having been invented as an MIT prank. (Although, to be fair, Smoot himself later became president of the ISO.)

Could exemplify the [potrzebie](https://en.wikipedia.org/wiki/Potrzebie) system. I’m not very familiar with it, but it was invented by Donald Knuth, upon whose concept of literate programming ODD is based.

martindholmes:
I don't know that we need to worry enormously about actual processability; any give project can write its own code for that. We're providing a method of documenting an equivalence, not acting on it. Ditto for the "risk" of XPath in your code; the same risk presumably pertains for regular expressions -- you might invoke a regular expression processor that has a bug that does rm -rf /, but that's not really our problem. Any processor you use might act in an unexpected way on any particular piece of TEI you have.

The issue of casting is interesting. We're documenting an equivalence between two abstract units (`@fromUnit` and `@toUnit`), and the attributes hold the names of those units, not some value of them. So `@fromUnit cast as xs:decimal ` doesn't really make sense; you're  casting a string name to a decimal value. So I'm inclined to think that's a red herring.

sydb:
> If we use `teidata.xpath` for our datatype, it sounds like we'll want to include a caveat intended for people working with the code--something along the lines of, "Be very careful with this, understanding that others processing your XPath expose themselves to security vulnerabilities." But I don't think that should be a reason against describing a relationship between variables using XPath, should it?

The caveat (which I think is very important and should be written very strongly) is **not** for people writing the XPaths, but for people who plan to _process_ the XPaths someone else has written. That someone else might be an Evil Genius bent on using the fact that you are executing an arbitrary XPath as a way of bringing down your system, or worse.

That said, I had not realized (at least, not when I wrote that) that the `@fromUnit` was a stand-in, rather than the value of an attribute. Maybe that’s why I prefer dollar signs. But take everything I say with a grain of salt until I have read the entire thread. :-)

martindholmes:
@sydb I think that's why you preferred variable names in Tokyo. I think it's a reasonable convention to say: $fromUnit = a quantity of @fromUnit things.

I have the feeling we said all of these things before and eventually arrived at the same conclusion. I wonder if it's documented in the minutes from Tokyo.

ebeshero:
I do remember mention of the `$` notation, yes, but the discussion was fast-paced and ethereal at the time. I don't think we explicated it fully in terms of examples as we're doing now--and I'm glad we're doing that.

martindholmes:
@sydb and me looking at this all in more detail find a big issue: there's no exemplified mechanism for pointing from a `<unit>` in the text to a `<unitDef>` in the header.  So in this example in unit.xml:

~~~~~
  <exemplum xml:lang="en">
      <p>Light travels at <num value="3E10">3×10^10</num>
        <unit type="rate" unit="cm/s">
          <unit type="space">cm</unit> per <unit type="time">second</unit>
        </unit>.</p>
    </egXML>
</exemplum>
~~~~~

the `@unit` attribute needs to be pointing to a `<unitDef>` OR another attribute needs to be provided/exemplified (`@unitRef`?) to make that connection. Without it, there is no way to specify that a unit you have marked in the text is an instance of a particular unit defined in a `<unitDef>`. This is a showstopper.

ebeshero:
@martindholmes Good catch, but I also think we can work out the pointing mechanism as part of this package, in time for the release, as well as work on `@formula`. I'm packing for a trip right now or I'd work on it immediately.

ebeshero:
@sydb @martindholmes Any reason we can't just use `@ref` on `<unit>` in the usual way to point to an `@xml:id` in the usual way?

martindholmes:
@ebeshero `<unit>` doesn't have `@ref`, does it? It would need to be added to att.canonical. And `@unitRef` follows the existing pattern used by other `@*Ref` attributes that point at specific ids. Also, we want to make this a single pointer, surely, whereas `@ref` would allow multiple pointers.

ebeshero:
@martindholmes @sydb I've been traveling to Utrecht for DH 2019, so I've been quiet on this a moment. But now I'm looking in on the spec for `<unit>` and right, we didn't assign it to `att.canonical`--and I expect we were waiting to set up `@unitRef`. Thanks for walking me through that, and I agree with  adding `@unitRef`, mainly because we want to limit the referencing to one and only one value. It definitely helps that we have precedent, as @martindholmes observes: I'm looking right now at `att.cReferencing` and that seems to be something like what we want here. 

Shall we just go ahead and create it, as an attribute assigned to a class like `att.unitReferencing`,  or `att.uReferencing`? I might want some help with this if we want to get it ready for this release, which I think we'd like to do.

sydb:
1. I forgot to post here, but I agree w/ @martindholmes that a dedicated `@unitRef` is somewhat better than using the `@ref` from att.canonical.
2. I think the new `@unitRef` should go in att.measurement, no?
3. Should it be mutually exclusive with `@unit`, though?
4. If it were to get a class of its own it would be something like att.cReferencing, but `@cRef` has completely different semantics and is `teidata.text`. We want `teidata.pointer` with Schematron that checks that it points to a `<unit>`, no?

ebeshero:
@sydb Yes--that makes sense to put it in `att.measurement`, though it works differently from `@unit`, `@quantity` and `@commodity`. I guess, right, it might not make sense to have an `@unit` as well as an `@unitRef` so we'd make it so you'd have one or the other (depending on whether you have a `<unitDecl>` to point to. I'd agree we want `teidata.pointer` for `@unitRef` (I'm not sure why `@cRef` doesn't do that, but that's a rabbit hole...), and we'd want a Schematron to make sure it resolves--to an `@xml:id` on `<unitDef>`, surely.

sydb:
### Rabbit hole

Because the value of `@cRef` is a canonical reference, not a URI or pointer. It (in theory) can be converted to a URI by the algorithms specified in `/TEI/teiHeader/encodingDesc/refsDecl`. E.g. `<ref cRef="tdg5:confdates">confdates</ref> is a useful element`, given something like
~~~~~xml
<cRefPattern matchPattern="tdg5:([a-z]+)"
  replacementPattern="https://tdg.docbook.org/tdg/5.0/$1.html">
~~~~~
is essentially the same as `<ref target="https://tdg.docbook.org/tdg/5.0/confdates.html">confdates</ref> is a useful element`.

martinascholger:
I've noticed that `<unitDef>` is currently a member of `model.global` and therefore can be used at any point within the text. That is not intended, right?

ebeshero:
@martinascholger Putting `<unitDef>` in `model.global` was in Naoki's proposal--see https://github.com/TEIC/TEI/issues/1707#issuecomment-345683240 . And I think it's there (and we didn't question it) because we think of `<unitDef>` potentially being used in a standoff way in a separate TEI document. (It occurs to me, then, that we may want a `<listUnit>` to match...)

jamescummings:
Re: model.global -- I really think we should be resistant to putting things in model.global which aren't needed globally. I can't think of a reason why it shouldn't just be somewhere specific in the header (and then yes, added to standoff if/when that appears). The only reason Naoki added it to model.global is he wasn't sure what the best model class to add it to (see https://github.com/TEIC/TEI/issues/1707#issuecomment-345683240) and the truth is it needs no model class.  

The originally proposed unitDecl was intended, I believe, as the equivalent to the listUnit @ebeshero suggests above. (Since it isn't really a list of things but a set of declarations.) unitDef does not need a model class because it is only used in one place, namely inside unitDecl. If it is to be used in standoff if/when that arrives, then it is unitDecl that will be used there, not unitDef directly.

Looked through everything else quickly and didn't spot anything glaring in need of change.

martindholmes:
I agree with @martinascholger and @jamescummings; unitDef belongs in unitDecl and nowhere else (for the moment at least). It's possible that `<unit>` needs a broader availability so that you can use it without `<measure>`, but that's a question for another day.

sydb:
Oooh! Something I disagree with @martindholmes about!

> It's possible that `<unit>` needs a broader availability so that you can use it without <measure>, but that's a question for another day.

I think it is a question for today, and that the answer is “yes”, `<unit>` should be broadly available. Just because I want to foreground units in my encoding does not mean I should be forced to foreground measurements. Besides, there are real conversations that discuss units without measurements. “Would you prefer a bag of galleons, sickles and knuts, or a hefty bar of latinum?” (Well, OK, that’s fake — but you still might have to encode it or something like it.)
Heck, half the conversations we have had about `<unit>` involve units without measurements. :-)

Without thinking about it much, my instinct is that `<unit>` should be a member of model.measureLike.

ebeshero:
@sydb Well `<unit>` is already in `model.measureLike` since we established it last year--so that's all right. :-) Let's see if I can work in a nice `@unitRef` now...

ebeshero:
Okay, I've set up `@unitRef` in `att.measurement` and wrote in a Schematron constraint with a `@role="warning"`  that `@unit` should not be used in the presence of `@unitRef`. I think that takes care of getting an `@unitRef` in place, though I didn't write Schematron to absolutely check that `@unitRef` is pointing at a `//unitDecl/@xml:id` somewhere (because what if it's in another document?) The language is there to guide people in the usage...

martindholmes:
@ebeshero That looks fine to me, although I don't actually think I agree with the Schematron warning. I don't see why people shouldn't use both strategies at the same time if they want to (for example, while transitioning from one to the other).

sydb:
> @ebeshero That looks fine to me, although I don't actually think I agree with the Schematron warning. I don't see why people shouldn't use both strategies at the same time if they want to (for example, while transitioning from one to the other).

While I am not 100% convinced either way that `@unit` and `@unitRef` should be mutually exclusive, in my mind the fact that someone might want them both temporarily while transitioning is not an argument for their being valid together.

martindholmes:
@syd If only one person who is not 100% sure is proposing the Schematron rule, I don't think that's enough justification for its being added right now. I think that should probably be a separate ticket, where Council gets to imagine possible scenarios that might warrant the use of both.

ebeshero:
Hmm. Okay--so I should just drop the Schematron rule? How about I leave it in, but commented out for now, and post a ticket about it?

Meanwhile I have set `@formula` to `teidata.xpath` am absolutely working some units of the Potzrebie system into examples...which means I'm adding a source to the Bibl...

sydb:
Yes, @martindholmes, except I think you have it backwards. We should start more restrictively, and if someone comes up with a scenario that requires both, submit a ticket to remove the restriction.

martindholmes:
@sydb OK, if that's Council's general principle, then I withdraw my objection.

sydb:
Well, I don’t know if it _is_ Council’s general principle, but since it is much easier to relax constraints than add them, I’m of a mind that it _should_ be. :-)

ebeshero:
Okay, I'm going to compromise here between @sydb and @martindholmes. We really did earlier say a few times that we liked the idea of making @unitRef be exclusive of @unit, an alternate way of referencing. But we just want to gently nudge people in that direction, so maybe we'll just set the Schematron rule to "info," with a little explanation of our line of thinking. Okay? Meanwhile I need to figure out how I broke the branch (sigh).

ebeshero:
@martindholmes @sydb @martinascholger @jamescummings @peterstadler  I think I've addressed everything we've raised here on the ticket. I've added `@unitRef` and changed `@formula` to `teidata.xpath`. And I implemented lots of examples, and had some fun with the "whatmeworry" units in the Potrzebie system (and added its source from that 1957 issue of Mad Magazine to BIB). I've been hunting for typos (corrected those that @martinascholger spotted. Take a read through it from the HD chapter to the specs and let's see if this is ready? Thanks, all!

martinascholger:
Great @ebeshero! Looks good to me, I think this is ready.

ebeshero:
As we are going into a release, I'd like to have this merged into dev before I make the release branch. I believe we've responded to all the requests now? Is there anything we missed?

ebeshero:
Jenkins is reporting green on this branch now. Anyone want to pull this in to dev?

PR #1895, opened 2019-06-29
bugfix(readme): fix broken tei-c.org link

duncdrum:
Final link on readme was broken.
checked all links on readme, this should be an automated test imv, but i leave that up to council.


peterstadler:
Thanks for the PR!

Issue #1805, opened 2018-08-28
Suggestion for new uniHan element


duncdrum:
[The UAX #38 unicode han database (unihan)](https://www.unicode.org/reports/tr38/)

is an important resources for folks encoding ideographic (as the Unicode standards uses the term) texts. Eg. It expresses, among other things, sorting rules, variant relations, dictionary references, and most importantly for the TEI imv, it contains normalized references to different national standards. 

Currently encoders can use
```xml
<localName type='some-national-standard'>some-name</localName>
<value>some value</value>
<!-- or -->
<localName type='unihan'> kRSKangXi </localName>
<value>120.5</value> 
<!-- or -->
<localName type='uax38'> kRSKangXi </localName>
<value>120.5</value> 
...
```

Of the ~90 properties only a hand full apply to  `unicodeName` proper, only in the rarest of cases, do uniHan properties apply to texts not using CJK(V). Since the use cases and audiences are fairly distinct, and since the UC is treating them separately, I'd like to suggest that tei follows UC, and adds a new `uniHan` element, to make these properties available in a structured fashion. 

The alternatives would be to allow `@type` on `unicodeName` and make it a closed list. But this seems a dangerous path, leading to how about `@subtype`? Similarly adding all possible values as legal contents to `unicodeName` seems a bad idea, since they don't apply equally to all users of unicodeName, so they shouldn't be forced to dig through 200 values just to find the one they need. Crucially, this would weaken the close alignment between unicode property names as the UC understands the term and as TEI would then implement it. Not all unihan properties are part of the ucd rnc schema; see #1804 .

The content model of the suggested element would closely mirror that of `unicodeName`, be part of the gaiji modul, and appear in `charProp`. In essence, the following should become valid tei. 
```xml
<charProp>
  <unicodeName>General_Category</unicodeName>
  <value>Lo</value>
  <uniHan>kRSKangXi</uniHan>
  <value>120.5</value>
</charProp>
```

martindholmes:
Wouldn't the example be:

~~~~~
<charProp>
  <unicodeName>General_Category</unicodeName>
  <value>Lo</value>
</charProp>
<charProp>
  <uniHan>kRSKangXi</uniHan>
  <value>120.5</value>
</charProp>
~~~~~

Would you imagine using it under both `<char>` and `<glyph>`, or only one of them?

duncdrum:
Good catch, fixed the listing, yes I'd say both `glyph` and `char` keeping things the way they work now.

[...]

PR #1897, opened 2019-06-30
Unihan phase 1

duncdrum:
This builds on #1896 and is ready for feedback and discussion, I ll remove the WIP, once phase 0 has been reviewed and merged.
As it stands both `@value` and `@name` are required on the three new elements.  
Since automation of definition generation is for a later phase i have made `@version` recommended, with a semi-open list.
My main question is how to go about the deprecation and in how far i should switch out the old prose and examples in the documentation already? But deprecation stuff is phase 2. 
The idea is that phase 1 could be included in a release, without folks really noticing much about it. i think this pretty much achieves that.

see #1805


duncdrum:
i can run `make test` locally but travis catches an actual problem,  I'm not sure where to add the reference to the new attribute class i just created. Also if any css wizard has some suggestions for how to get the eqivalent of `overlow: scroll` working for the spec pages without introducing new divs it would be much appreciated.

duncdrum:
could someone point me to the documentation for how i can grab a new detest log to replace the one in expected result?

martindholmes:
What we usually do for the detest log is to diff the one generated in the build process against the one in expected-results; if the only changes we see are those we expect from the code change, and the changes are all correct, then we copy the generated one over the expected-results one and commit.

It's crude, but we haven't been able to come up with a better approach. The important thing is to pay close attention when doing the diff, in case something unexpected sneaks past.

duncdrum:
what command do you use to make the log stay around after diffing? `make test` removes the generated log immediately so i can't check it.

martindholmes:
I thought it was only deleted if the test completed successfully. Are you running locally, or just looking at Travis?

duncdrum:
locally using docker
```
Validate ODD file using RELAX NG schema
rnv detest.rnc detest.xml > detest2.log 2>&1
Makefile:93: recipe for target 'detest' failed
make[1]: [detest] Error 1 (ignored)
Validate XML file using Schematron
java -Xmx752m -jar ../Utilities/lib/saxon9he.jar -ext:on detest.xml detest.xsl > detest3.log 2>&1
grep "\[xslt\]" detest1.log | sed 's/Processing.*detest.odd//' | sed 's/Loading.*message.isosch.xsl//' > detest.log
sed 's/.*detest.xml:/detest.xml:/' detest2.log >> detest.log
sed 's/.*detest.xml:/detest.xml:/' detest3.log >> detest.log
diff detest.log expected-results/detest.log
rm -f detest*log
```

travis does catch something, with a slighly more meaningful error message, but i also want to add things to detest...

duncdrum:
i think the command should be `rm -f detest*.log`?

martindholmes:
That line has been unchanged for six years, and it was Sebastian who last touched it, so I wouldn't quickly assume that it's wrong. :-)

duncdrum:
yeah red herring, i changed it to
```
	rm -f detest1.log detest2.log detest3.log
```
so i could keep my detest.log, but when i compare the newly generated one with the one in  in expected results i see two identical files, Yet:
`Makefile:93: recipe for target 'detest' failed`

P5/Source/Guidelines/en/WD-NonStandardCharacters.xml
```diff
@@ -1356,11 +1356,16 @@ The selection and combination of modules to form a TEI schema is described in
 <include xmlns="http://www.w3.org/2001/XInclude" href="../../Specs/charName.xml"/>
 <include xmlns="http://www.w3.org/2001/XInclude" href="../../Specs/charProp.xml"/>
 <include xmlns="http://www.w3.org/2001/XInclude" href="../../Specs/charDecl.xml"/>
+<!-- duncdrum: Why should / can I not include the att class here??  -->
+<!-- <include xmlns="http://www.w3.org/2001/XInclude" href="../../Specs/att.gaijiProp.xml"/> -->
```
duncdrum: If I include this here i get two name and two value properties on the elements which leads odd validation errors (two name attribtes) and to display issues with the html spec page.

duncdrum:
now that i can reproduce the CI failure i m going to clean this up

Issue #1833, opened 2018-10-31
New element for grouping notes: `<noteGrp>`


joeytakeda:
Stemming from a discussion on the TEI-L (https://listserv.brown.edu/archives/cgi-bin/wa?A2=TEI-L;c694031.1810), I would like to propose a new element for grouping together notes: `<noteGrp>`. [Thanks to Peter Boot for making the original suggestion for this element and for the others on that chain for their helpful recommendations and suggestions!]

The `<noteGrp>` element would be similar to the `<spanGrp>` element insofar as it would simply be a wrapper for grouping together series of notes. Usually this will be for the purposes of attachment (i.e. all of the children `<note>`s in `<noteGrp>` all attach to the same place in a document defined by the `@target` and `@targetEnd` attributes on `<noteGrp>`), but it could also simply be for providing a convenient way to group together notes of a similar type. 

## Classes

`<noteGrp>` would be a member of all the same classes as the `<note>` element:

```
<classes>
                  <memberOf key="att.global"/>
                  <memberOf key="model.noteLike"/>
                  <memberOf key="att.placement"/>
                  <memberOf key="att.pointing"/>
                  <memberOf key="att.typed"/>
                  <memberOf key="att.written"/>
                  <memberOf key="model.correspDescPart"/>
                  <memberOf key="model.correspActionPart"/>
                  <memberOf key="model.correspContextPart"/>
               </classes>
```

## Content

`<noteGrp>` can only contains a sequence of `<note>` elements (this is the same content model as `<spanGrp>`):

```
               <content>
                  <elementRef key="note" minOccurs="0" maxOccurs="unbounded"/>
               </content>
```

## Attributes

The `<noteGrp>` would also have the same attributes as `<note>`. I would suggest creating a new attribute class (att.noteAttachment or something) and include the `@anchored` and `@targetEnd` attributes defined within this new attribute class. 

But since the `@targetEnd` attribute is not recommended, however, I would also suggest that, if desired, the `@to` and `@from` attributes on `<span>` and `<app>` are harmonized into a class called `att.pointingRange` and that class be added to `<note>` and `<noteGrp>` instead of the creation of `att.noteAttachment`. However, this is a somewhat separate concern that may or may not be necessary to implement, depending on what Council thinks about `@targetEnd`.

## Summary

A new element for grouping notes that can go wherever notes can go; it will contain 0 or more `<note>` elements (following `<spanGrp>`'s content model); it will also bear the same attributes as the `<note>` element.

[...]

PR #1920, opened 2019-09-15
implement noteGrp

tuurma:
addresses #1833 

please could I ask for a review, particularly on

* [x] GL phrasing (thanks @martindholmes)
* [x] class membership specification

TODO

* [x] attributes specification


martindholmes (commented):
I would expand this:

"or presenting a short and more elaborate form of a note"

to this:

"or presenting a short form alongside a more elaborate form of a note"

martindholmes (commented):
There's a missing close parenthesis at the end of the para beginning "When alternative..."

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -2608,6 +2608,33 @@ effective than the free form <gi>note</gi> element; these elements are
...
+
+ <p>When alternative versions of the same note, e.g. in different languages or
+   presenting a short and more elaborate form of a note it is possible to group
+   such notes within the <gi>noteGrp</gi> element. Such notes would make use of
+   attributes such as <att>xml:lang</att> or <att>type</att> to differentiate their
+   field of application while sharing the same point of attachment (either implicit
+   in case of inline notes or explicitly specified in a <att>target</att> attribute.
+  </p>
+
+  <p>Simple example below demonstrates grouping of shortened and full version of
```
martindholmes: "The simple example below..." (adding The) 

and also: "shortened and full versions" (adding s)

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -2608,6 +2608,33 @@ effective than the free form <gi>note</gi> element; these elements are
...
+    the same note, perhaps shorter meant for media with space constraints.
```
martindholmes: ...where the shorter version might be intended for use in contexts with space constraints.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -2608,6 +2608,33 @@ effective than the free form <gi>note</gi> element; these elements are
 available when the module for simple analysis is selected (see section
 <ptr target="#AISP"/>).
  </p>
+
+ <p>When alternative versions of the same note, e.g. in different languages or
```
sydb: Sentence needs to be rewritten, possibly as multiple sentences. (It boils down to “When alternative versions of the same note … it is possible to group such notes within …”.)

Perhaps something like
~~~xml
A text may have multiple alternative versions of the same note. For
example the same annotation expressed in multiple languages, or
both an extensive note and a short form for different audiences. In
such cases multiple <gi>note</gi> elements can be grouped within a
<gi>noteGrp</gi> element.
~~~

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -2608,6 +2608,33 @@ effective than the free form <gi>note</gi> element; these elements are
...
+   presenting a short form alongside a more elaborate form of a note it is possible to group
+   such notes within the <gi>noteGrp</gi> element. Such notes would make use of
+   attributes such as <att>xml:lang</att> or <att>type</att> to differentiate their
```
sydb: Two thoughts: first, to be more specific about which notes we’re talking about; and second, wouldn’t the common `@target` typically be on the parent `<noteGrp>`, i.e. inherited by all the child `<note>`s?

If so, perhaps tweaking the wording to something like:
~~~xml
Typically the <gi>note</gi> elements within a <gi>noteGrp</gi> would
be differentiated by use of attributes such as <att>xml:lang</att> or
<att>type</att>, while sharing the same point of attachment (either
implicitly in case of inline notes, or explicitly specified via a
<att>target</att> attribute, which may be specified on the
<gi>noteGrp</gi> itself).</p>
~~~

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -2608,6 +2608,33 @@ effective than the free form <gi>note</gi> element; these elements are
...
+  </p>
+
+  <p>The simple example below demonstrates grouping of shortened and full versions of
+    the same note, where the shorter version might be intended for use in contexts with space constraints.
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:lang="en">
+      <p>... reuerendos dominos archiepiscopum et canonicos Leopolienses
+        in duplicibus Quatuortemporibus
+        <noteGrp>
+          <note type="short">Quatuor Tempora, so called dry fast days.</note>
+          <note type="full">Quatuor Tempora, so called dry fast days (Wednesday, Friday and Saturday)
```
sydb: Missing comma after “Friday”.

P5/Source/Specs/noteGrp.xml
```diff
@@ -0,0 +1,138 @@
...
+    <memberOf key="att.pointing"/>
+    <memberOf key="att.typed"/>
+    <memberOf key="att.written"/>
+    <memberOf key="model.correspDescPart"/>
+    <memberOf key="model.correspActionPart"/>
+    <memberOf key="model.correspContextPart"/>
+  </classes>
+  <content>
+    <sequence>
+        <classRef key="model.headLike" minOccurs="0"/>
```
sydb: I am thinking that if 1 heading is allowed, more than 1 should be allowed. Thus `<classRef key="model.headLike" minOccurs="0" maxOccurs="unbounded"/>`, no?

P5/Source/Specs/noteGrp.xml
```diff
@@ -0,0 +1,138 @@
...
+        <elementRef key="desc" minOccurs="0" maxOccurs="unbounded"/>
+        <alternate minOccurs="1" maxOccurs="unbounded">
+          <elementRef key="note"/>
+          <elementRef key="noteGrp"/>
+        </alternate>
+    </sequence>
+  </content>
+  <attList>
+    <attDef ident="anchored" usage="opt">
```
sydb: Rather than repeating all these remarks, should we perhaps move `@anchored` to its own class? Same goes for `@targetEnd`, I suppose.

P5/Source/Specs/noteGrp.xml
```diff
@@ -0,0 +1,138 @@
...
+      <remarks versionDate="2017-06-25" xml:lang="de">
+        <p>Dieses Attribut wurde beibehalten, um eine Abwärtskompatibilität zu gewährleisten; bei einem
+          zukünftigen Release der Guidelines kann es entfernt werden. Der empfohlene Weg, um auf einen
+          Bereich zu verweisen, ist die Verwendung der <ident>range</ident>-Funktion von XPointer, wie in
+          <ptr target="#SATSRN"/> beschrieben.</p>
+      </remarks>
+    </attDef>
+  </attList>
+  <exemplum xml:lang="la">
+    <p>In the following example, two notes specifying the content of the annotation relating to the
```
sydb: Confusing sentence. How about “In the following example, there are two notes in different languages, each specifying the content of the annotation relating to the preceding fragment of text.”?

P5/Source/Specs/noteGrp.xml
```diff
@@ -0,0 +1,138 @@
...
+      preceding fragment of text in different languages are grouped within <gi>noteGrp</gi>:</p>
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
+      <p>... Parwum vadit theoloneum, tamen reuerendos dominos archiepiscopum et canonicos Leopolienses
+        necnon episcopum in duplicibus Quatuortemporibus
+        <noteGrp>
+          <note xml:lang="en">Quatuor Tempora, so called dry fast days (Wednesday, Friday and Saturday)
```
sydb: Missing same comma after “Friday”.

sydb (commented):
I think the class membership specification looks fine.
I think it would be a good idea to have an example that used `@target`.

P5/Source/Guidelines/en/CO-CoreElements.xml
tuurma: agreed, thank you

P5/Source/Specs/noteGrp.xml
tuurma: Thanks Syd for focusing on this! I had a look at other *Grp elements and only `spGrp` allows any headings (but it does allow multiple too). 
I think any headings make sense only for transcribing (e.g. printed) sources, but I believe for such purposes we were planning a `listNote` element and `noteGrp` would be something. So perhaps I'll remove model.headLike from the noteGrp altogether?

P5/Source/Specs/noteGrp.xml
tuurma: Not sure what's the procedure. Never done it myself.

P5/Source/Specs/noteGrp.xml
sydb: I think @hcayless’ “don’t put it in unless requested” philosophy would definitely have us leave `<head>` out for now. And while I often think particular cases override that philosophy, I do not think this one does.

P5/Source/Specs/noteGrp.xml
sydb: [Outline of procedure sent to @tuurma via Slack.]

tuurma:
I think it's good to go. Thanks @sydb for assistance

P5/Source/Specs/noteGrp.xml
```diff
@@ -0,0 +1,57 @@
...
+          <elementRef key="note"/>
+          <elementRef key="noteGrp"/>
+        </alternate>
+    </sequence>
+  </content>
+  <exemplum xml:lang="la">
+    <p>In the following example, there are two notes in different languages, each specifying the content of 
+      the annotation relating to the same fragment of text:</p>
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
+      <p>... Parwum vadit theoloneum, tamen reuerendos dominos archiepiscopum et canonicos Leopolienses
```
hcayless: Incredibly pedantic comment; feel free to ignore: "Parwum" might be what the source has, but I'm suspicious, because the correct Latin form would be "Paruum" or "Parvum" (if your orthography distinguishes 'u' and 'v'—"reuerendos" doesn't, since it's not "reverendos"). Latin doesn't have a 'w', though all sorts of weirdness can happen in later Latin :-).

tuurma: To satisfy your sweet Latin tooth, I'm no palaeographer ;-) 

image

tuurma: There is another case in the same document they seem to be using w (3rd line gwerrarum). I'm happy to shorten the example though, to avoid controversy

image

hcayless: I think it is "parvum". But I agree that's "gwerrarum" in the other example, :-)

I am not a paleographer either. What is this text?

sydb:
@tuurma has implmeneted #1833.

Issue #1823, opened 2018-10-01
teiCorpus: make guidelines and data model consistent


mccaskey:
All the guidelines, all the examples, and the note in the spec for the element say that a `<teiCorpus>` element contains (“must contain” in the note) `<TEI>` elements. But the data model allows `<teiCorpus>` to contain no `<TEI>` elements and a `<text>` element instead.

jamescummings:
I think the Guidelines are correct (and should be made more clear) and I think the addition of model.resourceLike to the content model as a whole here is wrong. If one is going to use teiCorpus to then have some alternating teiHeader and text elements in it, it is far better to wrap those in a TEI element and hardly any problem. 

I believe this is in error when we were making changes to membership of model.resourceLike and added text to it in release 3.0.0. Compare: 
http://www.tei-c.org/Vault/P5/3.0.0/doc/tei-p5-doc/en/html/ref-model.resourceLike.html
and
http://www.tei-c.org/Vault/P5/2.9.0/doc/tei-p5-doc/en/html/ref-model.resourceLike.html

While its use inside TEI makes sense, its use in the content model of teiCorpus doesn't. (IMHO).

sydb:
I think I agree w/ @jamescummings, here. I think adding model.resourceLike+ to the before-TEI portion of the content model (which was done in 2.4.0) made sense, but
 a) The content model should have been changed to `teiHeader, model.resourceLike*, ( TEI | teiCorpus )+` — that is, the `<TEI>` (or nested `<teiCorpus>`) bit should always be required, not optional as we have now, and maybe
 b) a Schematron rule should have said “no `<text>` child of `<teiCorpus>`”.

However, @jamescummings, note that you can’t have alternating `<teiHeader>` and `<text>` elements inside a `<teiCorpus>`, even with this current bad content model.

peterstadler:
I admittedly never(?) used `teiCorpus` but always thought of it as a pure wrapper for `TEI` or `teiCorpus` elements – and this is how I understand the remark: "Must contain one TEI header for the corpus, and a series of TEI, one for each text."

Yet, in [2.4.0](http://www.tei-c.org/Vault/P5/2.4.0/doc/tei-p5-doc/readme-2.4.0.html) it was decided to "allow a `<teiCorpus>` to have `<facsimile>` and `<sourceDoc>` children, like `<TEI>` (per [FR 456](https://sourceforge.net/p/tei/feature-requests/456/))". 

[...]

PR #1922, opened 2019-09-15
change teiCorpus content model

peterstadler:
as per Council face2face at Graz #1823.

The one thing I couldn't solve is to allow for interleaved `<TEI>` or `<teiCorpus>`. Those have to be put at the end now. 
The underlying problem of creating an ambiguous content model was already imminent with the old content model once you designed a schema without a member of `model.resourceLike`.

### DIFF
--- a/P5/Source/Specs/teiCorpus.xml
+++ b/P5/Source/Specs/teiCorpus.xml
@@ -36,25 +36,17 @@
   <content>
     <sequence>
       <elementRef key="teiHeader"/>
-      <alternate>
-        <sequence>
-          
-            <classRef key="model.resourceLike" minOccurs="1" maxOccurs="unbounded"/>
-          
-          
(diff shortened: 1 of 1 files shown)


sydb:
@peterstadler — I think I would prefer to see this work done in the standOff branch, no? Can you take a look at the [teiCorpus tagdoc there](https://github.com/TEIC/TEI/blob/sydb-standOff/P5/Source/Specs/teiCorpus.xml) (in particular commit 10991f7b7eb5184c588e5943532189eac7ff1577) and see if you think it addresses #1823 or not?

martinascholger:
@peterstadler @sydb can be close this PR?

sydb:
@martinascholger: I think so. The content model of `<teiCorpus>` is now 
~~~~~
( teiHeader, model.resource*, model.describedResource+ )
~~~~~
thus requiring 1 or more of `<TEI>` or `<teiCorpus>`.

Issue #1830, opened 2018-10-22
<derivation> should allow <bibl> in its content model


martindholmes:
`<derivation>` can be used to specify that a text is a translation from another text:

`<derivation type="translation">`

but there's no easy way to specify what it's a translation of. I think the content model of `<derivation>` should allow `<bibl>` so that information about the original text can be supplied.

lb42:
The spec for derivation says that info about the sources from which a text is derived should be supplied in the sourceDesc. Which makes more sense to me

emylonas:
Council F2F with @martindholmes agree that the bibliography for a translation or other type of derivation information belongs in `<sourceDesc>`.  It will be helpful to add an example showing this to `<sourceDesc>` and assign to Martin to add the example that caused this issue.

martindholmes:
Pull request issued.

martindholmes:
This is now just waiting on pull request https://github.com/TEIC/TEI/pull/1924.

tuurma:
Happy to merge unless requested reviewers would like to have a say @jamescummings @emylonas @martinascholger

martinascholger:
This has been merged, so I'm closing this.

PR #1924, opened 2019-09-16
Fixed example added for issue #1830.

martindholmes:
Please check that this makes sense based on your memory of the ticket.


peterstadler:
pinging @jamescummings, @emylonas, @martinascholger to review!

PR #1940, opened 2019-10-17
Removed anchored='false' attribute from examples in SA.html, Using Po…

### LINKED ISSUE #1932: Example error in 16.1.2 Using Pointers and Links (first example following Dunciad image)
The first example following the Dunciad image reads:
```xml
<l>(Diff'rent our parties, but with equal grace</l>
<l>The Goddess smiles on Whig and Tory race,</l>
<l>
 <note type="imitation" place="bottom"
  anchored="false">
  <bibl>Virg. Æn. 10.</bibl>
  <quote>
   <l>Tros Rutulusve fuat; nullo discrimine habebo.</l>
   <l>—— Rex Jupiter omnibus idem.</l>
  </quote>
 </note>'Tis the same rope at sev'ral ends they twist,
</l>
<l>To Dulness, Ridpath is as dear as Mist)</l>
```
*****
First, the value of 'anchored="false"' is simply wrong. What else could Verse 283-84. or the prose "With equal grace Our Goddess smiles on Whig and Tory race" be other than anchors? 

Second, the anchors are simply lost from the note, although one could record a decision to not capture anchors but that should be mentioned in the text here, lest casual readers become confused. 

Third, the note confused two notes, the first one being the two anchors plus Virg. Æn. 10., which is one entire note, with "Tros Rutulusve fuat; nullo discrimine habebo. -- Rex Jupiter omnibus idem." which is a second note, albeit also found in Virg. Æn. 10.

This will be clearer if you consult the image of the 1829 edition here: http://grubstreetproject.net/works/T5549?display=image&page=162

Which clearly anchors both notes to V. 283-84 (plus the prose anchor) and uses "idem." to tie the second note to the reference Virg. Æn. 10.

In summary, there should be two distinct note elements, both anchored by verse and prose and containing distinct content. I'll file separate comments on the other examples but in general they follow a similar theme.

pdurusau:
Removed anchored='false' attribute from examples in SA.html, Using Pointers and Links

### DIFF
--- a/P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
+++ b/P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
@@ -162,7 +162,7 @@ of notes, one headed <q>Remarks</q> and the other
 <q>Imitations</q>.</note>
 <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:lang="mul" source="#SAPTEG-eg-3"><l>(Diff'rent our parties, but with equal grace</l>
 <l>The Goddess smiles on Whig and Tory race,</l>
-<l><note type="imitation" place="bottom" anchored="false">    
+<l><note type="imitation" place="bottom">    
     <bibl>Virg. &#xC6;n. 10.</bibl>
     <quote>
 <l>Tros Rutulusve fuat; nullo discrimine habebo.</l>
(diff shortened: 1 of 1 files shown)


sydb:
Hi @pdurusau! Long time no see! Good to hear from you. But I am curious about what you have against those `@anchored` attributes. You are not suggesting these notes are, in fact, anchored, are you? Or are you just thinking the extra attribute is not helpful in this discussion about how one connects an annotation with that which is annotated?

pdurusau:
Sorry for the long absence, trying to recapture some TEI skills for a lexicon project. 

Yes, the notes are anchored, else how to you explain Verse 283-284 at the beginning of the footnote (in the image of the text to be encoded)? Isn't that an anchor? Granting the anchor isn't captured by the encoding in question, which is an encoding choice, but in the text being encoded, the anchor does in fact exist. Yes? 

Removing the anchored="false" corrects the mis-impression that no anchor existed in the text being encoded. The omission of anchoring text should be mentioned in the encoding description if not elsewhere as well.

sydb:
I approve. Merge away!

Issue #1769, opened 2018-04-30
teidata.pointer equivalent to move/@where


joeytakeda:
The Map of Early Modern London is continuing to encode mayoral pageants, which took place across a number of places throughout London. These documents detail both the performances at particular sites as well as how the entire show moved from one place to another. We would like to be able to encode these movements using the `<move>` element and link these to our database of places, but currently `move/@where` is defined as teidata.word. Is there an equivalent to denote a particular place defined by an entity? For example:

```
<move where="locations.xml#place1"/>
```
Can there be a way to denote a place via `@where` that is a pointer (note that `event/@where` *is* a teidata.pointer and requires pointing to a `<place>` element)?  I understand the utility of `move/@where` being teidata.word, so could there be a new attribute--maybe something like `@wherePtr` --that allowed a `<move>` to point to a defined place?

lb42:
Is it the case that your @where attributes will *always* take a URL value? If so, you could just redefine @where in your ODD with a teidata.pointer value. This would mean your texts are still unquestionably TEI conformant -- since teidata.pointer values are a subset of teidata.word values -- and it would be obvious to any user of your data what you were doing. Adding a new attribute would also work, but seems like more work/potential confusion.

martindholmes:
@lb42 using teidata.word to store pointers is OK within the project itself, but it isn't great for interchange; it would be difficult for a downstream user to determine whether something is a word or a pointer, unless an obvious protocol prefix is there (which it wouldn't be in this case -- just a private URI scheme).

lb42:
@martindholmes I think you are misunderstanding my point: a "downstream user" (whatever that is) should look at the ODD for the project to determine how an attribute's value should be interpreted. My suggestion is to make explicit there that your @where s are pointers.

martindholmes:
@lb42 You have more confidence in the willingness of future users of a file to investigate its background than me. :-) And the "downstream user" may not be human anyway.

lb42:
Well, if we are not allowed to assume that future users of a file might look at its associated schema, why are we bothering to provide schemas at all?

ebeshero:
This seems to be parsing things rather finely. I've got a question about the nature of subsets: @lb42 mentions that " teidata.pointer values are a subset of teidata.word values". What issues are there with interchange when we refine datatypes of attribute values to be a subset classification? I'm not sure I'm following why the refining of datatype in the schema is problematic for interchange.

lb42:
[...]

PR #1958, opened 2020-01-16
Adding where to new att.locatable to address #1769

ebeshero:
I've moved `@where` to a new `att.locatable` and I'm hoping we can look at / improve my new explanation of it. I'm also hoping we can add some translations to other languages here.

### DIFF
--- a/P5/Source/Guidelines/en/ND-NamesDates.xml
+++ b/P5/Source/Guidelines/en/ND-NamesDates.xml
@@ -815,6 +815,7 @@ target="#CONA"/>).  The tag -->The
         of one or other of the above three classes. Generic elements for state, trait, and event are also defined: <specList>
           <specDesc key="state"/>
           <specDesc key="trait"/>
+          <specDesc key="event"/>
           <specDesc key="listEvent"/>
           <!-- PB, 19-04-2012
     <specDesc key="listState"/>
--- a/P5/Source/Guidelines/en/ND-NamesDates.xml
+++ b/P5/Source/Guidelines/en/ND-NamesDates.xml
@@ -815,7 +815,6 @@ target="#CONA"/>).  The tag -->The
         of one or other of the above three classes. Generic elements for state, trait, and event are also defined: <specList>
           <specDesc key="state"/>
           <specDesc key="trait"/>
-          <specDesc key="event" atts="where"/>
           <specDesc key="listEvent"/>
           <!-- PB, 19-04-2012
     <specDesc key="listState"/>
--- a/P5/Source/Specs/att.locatable.xml
+++ b/P5/Source/Specs/att.locatable.xml
@@ -7,8 +7,25 @@ $Date$
 $Id$
 -->
 <?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
-<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" type="atts" ident="att.locatable">
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" xmlns:sch="http://purl.oclc.org/dsdl/schematron" module="tei" type="atts" ident="att.locatable">
   <desc versionDate="2020-01-15" xml:lang="en">provides attributes that specify locational information.</desc>
+<constraintSpec ident="att-locatable-where" scheme="schematron">
+    <constraint>
+      <sch:pattern id="testing-if-anyURI">
+      <sch:rule context="tei:*[@where[contains(string(), ':')]]">
+        <sch:assert test="matches(tokenize(@where/string(), ':')[1], '^[A-Za-z0-9+.-]+$')" role="warning">The value of the where attribute will need to conform to the xsd:anyURI datatype. If one or more colons are used, the first colon must only be preceded by one or more simple alphanumeric characters, the plus, minus, or period symbols: [A-Za-z0-9+.-]+.
+          Finally, if the # is present it must only appear once.</sch:assert>
+      </sch:rule>
+      <sch:rule context="tei:*[@where[contains(string(), '%')]]">
+        <sch:let name="tokens" value="tokenize(@where, '%')[position() gt 1]"/>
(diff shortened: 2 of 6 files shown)


ebeshero:
Okay, we're really passing build tests now. Let's take a look and see if this is what we want to add to the Guidelines.

P5/Source/Guidelines/en/ND-NamesDates.xml
```diff
@@ -815,7 +815,6 @@ target="#CONA"/>).  The tag -->The
         of one or other of the above three classes. Generic elements for state, trait, and event are also defined: <specList>
           <specDesc key="state"/>
           <specDesc key="trait"/>
-          <specDesc key="event" atts="where"/>
```
peterstadler: why is `<event>` removed completely?

P5/Source/Specs/att.locatable.xml
```diff
@@ -0,0 +1,58 @@
...
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" type="atts" ident="att.locatable">
+  <desc versionDate="2020-01-15" xml:lang="en">provides attributes that specify locational information.</desc>
+  <attList xmlns:rng="http://relaxng.org/ns/structure/1.0">
```
peterstadler: this rng prefix namespace declaration seems to be an unnecessary copy/paste leftover?

P5/Source/Specs/att.locatable.xml
```diff
@@ -0,0 +1,58 @@
...
+    <attDef ident="where" usage="opt">
+      <desc versionDate="2020-01-15" xml:lang="en">The <att>where</att> specifies a location by using a URI to another resource.</desc>
```
peterstadler: maybe better: "by using a URI to point at another resource"?

P5/Source/Specs/att.locatable.xml
```diff
@@ -0,0 +1,58 @@
...
+                <l>HIſtory, Truth, and Vertue ſeeke by name,</l>
+                <!--...-->
+              </lg>
+            </sp>
+          </div>
+        </egXML>
+      </exemplum>
+ 
+  <remarks versionDate="2020-01-15" xml:lang="en">
+    <p>This attribute class provides <att>where</att> for use in pointing to a URI containing a canonical name or information about the locational context of an <gi>event</gi> or a <gi>move</gi> on stage. It can be used for pointing out blocking cues, as in stage left, right, or center, where these locations are defined in the <gi>teiHeader</gi> or an external source. It may also be used to point to a canonical <gi>place</gi> defined in a <gi>listPlace</gi>.</p>
```
peterstadler: maybe: "… for use in pointing to a resource containing a canonical …"

P5/Source/Specs/move.xml
```diff
@@ -173,7 +104,7 @@ $Id$
         </castList>
       </performance>
       <!-- ... -->
-      <stage type="entrance"><move who="#bellaf" type="enter" where="L" perf="#perf1"/> Enter
+      <stage type="entrance"><move who="#bellaf" type="enter" where="#L" perf="#perf1"/> Enter
```
peterstadler: I think we'll need an example for resolving the `where="#L"` as all other targets of references ("#bellaf" and "#perf1") are provided.

P5/Source/Specs/move.xml
```diff
@@ -203,7 +134,7 @@ $Id$
         </castList>
       </performance>
       <!-- ... -->
-      <stage type="entrance"><move who="#zh-tw_bellaf" type="enter" where="L" perf="#perf1"/>
+      <stage type="entrance"><move who="#zh-tw_bellaf" type="enter" where="#L" perf="#perf1"/>
```
peterstadler: see above for resolving the `where="#L"`. 
And it just occurred to me, that this should probably be 'localized' as well, i.e. `where="#zh-tw_L"` – which is also true for the `@perf` here!

peterstadler (changes requested):
the bibliography shows a lot of whitespace changes. Is it somehow possible to reset these and only add the needed changes?

P5/Source/Specs/att.locatable.xml
martindholmes: The wording is a little weird here too: "The `@where` specifies...". Should be "the `<att>where</att>` attribute specifies...", no?

P5/Source/Guidelines/en/ND-NamesDates.xml
ebeshero: @peterstadler Oops! Thanks for catching that--I was just intending to remove the atts.

P5/Source/Specs/att.locatable.xml
ebeshero: Yep--good catch. I'm removing it now.

P5/Source/Specs/att.locatable.xml
ebeshero: Thanks--just checking these off as I edit them. :-)

P5/Source/Specs/att.locatable.xml
ebeshero: Yes--that's better! Edited.

ebeshero: I'm also adding a bit more to the remarks to make the drama context a little clearer for `<move>`.

ebeshero:
@peterstadler @martindholmes @martinascholger @TEITechnicalCouncil Thanks for your help with this PR! I've made some changes as we discussed to expand the examples, so we now have some lists of stage movements for `@where` to point to. I've also made some revisions to the wording. How does this look? Also, can I get some help with figuring out how to do something *like* a deprecation of the old way of doing things, like we discussed in our meeting? @sydb : I'm trying to remember how we do this.

ebeshero:
Edited the above to separate this part out: **If** we decide this is okay to merge into dev, can I get some help with figuring out how to do something *like* a deprecation of the old way of doing things, like we discussed in our meeting? @sydb : I'm trying to remember how we do this.

ebeshero:
Just an update on this branch--I think we resolved the requested changes now, but there's an outstanding issue (not on the reviews I think?) about how we talk to people about their use of the new datatype on `@where` which might invalidate existing values. I could use a little help with that, but I'll try something on it later today.

ebeshero:
VF2F decision: We are thinking of better ways to handle the expression of stage movement, and think `att.locatable` is probably not the best approach to take now. Closing and not merging.

PR #1974, opened 2020-02-12
Datatype for values that may be value lists or URIs.

hcayless:
As discussed in Council meeting yesterday, the problem of move/@where could be addressed by a datatype that permits either words from a value list or URIs. This pull request implements that feature.


hcayless:
Linking this to #1769

martinascholger:
@ebeshero could you please have a look at this - thanks!

peterstadler:
I like 👍

ebeshero:
Since I'm the last one on the reviewer list, shall I pull it in?

PR #1983, opened 2020-03-24
Citation Structures

### LINKED ISSUE #1957: FR: Declarative Citation Structure
The TEI element [`<cRefPattern>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-cRefPattern.html) defines a mechanism for resolving "canonical" references into machine-actionable URIs. While it permits automated resolution of these references, it does not provide for the automated *construction* of references. So while TEI documents can declare how citations should be treated, they can't actually declare how those citations should be made. This means systems that might want automatically to do things with a document's citation structure, like chunk a large document into smaller pieces or generate a table of contents are at a disadvantage.

We (@PonteIneptique and myself) are proposing two new elements to allow documents to declare their citation structure and to specify how structural information may be extracted from them:

`<citeStructure>` permits an application to address and extract parts of the TEI document declaring its structure,
and can provide hints on how to construct canonical references. Instead of telling an application how to
resolve  references, as `<cRefPattern>` does, this tells the application how to *build* them using
the TEI document as a data source.
        
`<citeStructure>` has three required attributes, `@unit`, a label specifying what kind of unit this level
of structure refers to, e.g., "book", "line", or "section"; `@match`, an XPath that tells you how to
find each unit in of the given type in the document; `@use`, an XPath, relative to the element(s) matched
by `@match` that gives a mechanism for uniquely identifying the element(s) referenced by `@match`. This could
be an `@xml:id`, or a labeling attribute like `@n`, or even a positional reference like `position()`. It should
be possible to construct an XPath that will retrieve an element referenced by `<citeStructure>` by constructing
a predicate on the XPath in `@match` using the value for a given element of the XPath in `@use`.
             
`<citeStructure>` also has an optional attribute, `@delim`, which can be used as a hint for the construction of 
canonical references, e.g. "1.2" for the poem `<div n="1">` and the `<l n="2">`. In those references,
the value of `@delim` should be assumed to come before the referenced unit. So, in example 2 (again, assuming a 
reference style like "1.2"), the line `<citeStructure>` is the one that has a `@delim`.
             
`<citeStructure>`s may nest, and a relative XPath in `@match` is assumed to start from the element(s) matched
by the parent `<citeStructure>`. See example 2 below. 
             
`<citeData>` provides a way for a document to declare how certain types of data can be extracted from the
structures mapped out by `<citeStructure>`. In examples 1 and 2, a Dublin Core Title can be obtained for each
front section or poem by getting the `<head>` element from each.
             
`<citeData>` has two required attributes, `@property` (analogous to `@property` in RDFa), a URI, and `@use`,
a relative XPath exactly like `@use` on `<citeStructure>`. 

```xml
<encodingDesc>
    <refsDecl>
        <!-- Example 1: Units of the introductory front matter may be cited, e.g. "Intro. 1" -->
        <citeStructure unit="front"
            match="//front/div[@type='introduction']/div[@type='section']" 
            use="@n" delim="Intro. "> 
            <!-- The thing(s) matched by ../@match have titles at the location ./head. 
                 Assume the prefix "dc" has been declared using a <prefixDef> elsewhere.
                 Multiple <head>s in the matched <div> would mean multiple dc:titles being
                 extracted for that div.-->
            <citeData property="dc:title" use="head" /> 
        </citeStructure>
        <!-- Example 2: Poems may be cited as, e.g. "1", lines as, e.g. "1.5" -->
        <citeStructure unit="poem"
            match="//body/div[@type='edition']/div[@type='textpart']" use="@n">
            <citeData property="dc:title" use="head" />
            <citeStructure unit="line" 
                match=".//l[parent::div or parent::lem]" use="@n" delim=".">
            </citeStructure>
        </citeStructure>
    </refsDecl>
</encodingDesc>
```
The following example is the equivalent of the one given for `<cRefPattern>` in https://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html#SACR and could be used (with a different algorithm) both to resolve citations of the type "Matt 5:7" and/or to generate a set of resolvable citations at any level.
```xml
<encodingDesc>
  <refsDecl>
    <citeStructure unit="book" match="//body/div" use="@n">
      <citeStructure unit="chapter" match="div" use="position()" delim=" ">
        <citeStructure unit="verse" match="div" use="position()" delim=":"/>
      </citeStructure>
    </citeStructure>
  </refsDecl>
</encodingDesc>
```

hcayless:
See #1957 for the relevant feature request and discussion


hcayless:
Build is failing because citeStructure in examples isn't valid against the dev nvdl, which, of course, doesn't know about it yet. Another circular dependency!

P5/Source/Specs/citeData.xml
```diff
@@ -0,0 +1,40 @@
...
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="header" ident="citeData">
+  <gloss versionDate="2020-02-03" xml:lang="en">citation data</gloss>
+  <desc versionDate="2008-01-13" xml:lang="en">specifies how information may be extracted from citation structures.</desc>
```
ebeshero: @hcayless You probably want to change the `@versionDate` on `<desc>`.

P5/Source/Specs/citeData.xml
```diff
@@ -0,0 +1,40 @@
...
+  <classes>
+    <memberOf key="att.global"/>
+    <memberOf key="att.citeStructurePart"/>
+  </classes>
+  <content>
+    <empty/>        
+  </content>
+  <attList>
+    <attDef ident="property" usage="req">
```
ebeshero, line 21: @hcayless I wonder whether `@property` ought to be required or just `rec` or `opt`. I don't have any strong objections to its being required, but it demands that people encoding citation structures must work with RDF. I can imagine someone wanting to specify citation structures (for say, building a table of contents within a project) who isn't interested in specifying an RDF property for a specific item they want to pull.

P5/Source/Specs/citeStructure.xml
```diff
@@ -0,0 +1,72 @@
...
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="header" ident="citeStructure">
+  <gloss versionDate="2020-02-03" xml:lang="en">citation structure</gloss>
+  <desc versionDate="2008-01-13" xml:lang="en">declares a structure and method for citing the current document.</desc>
```
ebeshero: @hcayless Change the `@versionDate` on `<desc>` here, too.

P5/Source/Specs/citeStructure.xml
```diff
@@ -0,0 +1,67 @@
...
+        <dataRef key="teidata.xpath"/>
+      </datatype>
+    </attDef>
+    <attDef ident="unit" usage="opt">
+      <gloss versionDate="2020-02-03" xml:lang="en">unit</gloss>
+      <desc versionDate="2020-02-03" xml:lang="en">describes the structural unit indicated by the <gi>citeStructure</gi>.</desc>
+      <datatype>
+        <dataRef key="teidata.enumerated"/>
+      </datatype>
+      <valList type="open">
```
ebeshero, line 73: @hcayless Good that it's an open `valList`. In addition to the suggestions below, maybe we could add: 
- "volume"
- "section"
- "entry"
- "letter"

ebeshero (changes requested):
I wonder whether the `@property` attribute on `<citeData>` (pointing to an RDF property) should really be recommended or optional, instead of required. We're basically saying if you're going to use `<citeData>`, you must be ready with RDF properties, and I think we could walk that back a little and just *recommend* their use. What do others think?

P5/Source/Specs/citeData.xml
hcayless: `@property` must be required, because otherwise `<citeData>` won't be able to say what kind of data it represents. But it shouldn't be required to be an RDF property. It might map to a category in a taxonomy or a SKOS Concept, etc.

hcayless:
I have integrated @ebeshero's suggestions and merged in the dev branch so that there should be no barriers to merging this.

P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
```diff
@@ -1213,6 +1213,47 @@ to <code>$</code> during processing.</p>
     references in non-TEI documents (such as plain text
     documents), possibly as part of their conversion to TEI.</p>
 </div>
+  <div type="div4" xml:id="SACRCS">
+    <head>Citation Structures</head>
+    <p>The equivalent structure to the set of <gi>cRefPattern</gi>s in <ptr target="#SACRWE"/> would be:
```
sydb: This needs some introductory prose before jumping into the equivalent structure.
I think some of the verbiage from the last “paragraph” of this section (i.e., “One advantage `<citeStructure>` … tables of contents.”) could usefully be moved here.

P5/Source/Specs/att.citeStructurePart.xml
```diff
@@ -0,0 +1,23 @@
...
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" xmlns:rng="http://relaxng.org/ns/structure/1.0" module="tei" type="atts" ident="att.citeStructurePart">
+  <desc versionDate="2016-03-15" xml:lang="en">provides attributes for selecting particular elements
+    within a document.</desc>
+  <attList>
+    <attDef ident="use" usage="req">
+      <gloss versionDate="2020-02-03" xml:lang="en">use</gloss>
+      <desc versionDate="2020-02-03" xml:lang="en">supplies an XPath expression relative to the context given in <att>match</att>,
```
sydb: My first instinct is no, this is not an XPath expression but rather an XPath selection pattern. (The expression "2+79" does not make sense as the value of `@use` — and, if I understand correctly (50% chance) a hard-coded string would never occur, because it would be the value of `@delim` instead.)

P5/Source/Specs/citeData.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="header" ident="citeData">
```
sydb, line 10: Why does this element exist? Would not minor modifications of `<equiv>` keep things slimmer and be more appropriate?

P5/Source/Specs/citeStructure.xml
```diff
@@ -0,0 +1,76 @@
...
+      <datatype>
+        <dataRef name="string" restriction=".+"/>
+      </datatype>
+      <remarks versionDate="2020-03-24" xml:lang="en">
+        <p><att>delim</att> must contain at least one character.</p>
+      </remarks>
+    </attDef>
+    <attDef ident="match" usage="req">
+      <gloss versionDate="2020-02-03" xml:lang="en">match</gloss>
+      <desc versionDate="2020-02-03" xml:lang="en">supplies an XPath expression which returns a sequence of nodes which are
```
sydb: Not an XPath expression but rather an XPath selection pattern. Should also give a citation to XPath 3.0 spec. The description of `@match` of att.scoping says
~~~xml
supplies an XPath selection pattern using the
syntax defined in <ptr target="#XSLT3"/> which
identifies a set of nodes …
~~~
which seems apropos.

P5/Source/Specs/citeStructure.xml
```diff
@@ -0,0 +1,76 @@
...
+      </datatype>
+      <remarks versionDate="2020-03-24" xml:lang="en">
+        <p><att>delim</att> must contain at least one character.</p>
+      </remarks>
+    </attDef>
+    <attDef ident="match" usage="req">
+      <gloss versionDate="2020-02-03" xml:lang="en">match</gloss>
+      <desc versionDate="2020-02-03" xml:lang="en">supplies an XPath expression which returns a sequence of nodes which are 
+        citable structural components. The expression may be absolute (beginning with <code>/</code>) or relative. 
+        <att>match</att> on a <gi>citeStructure</gi> without a <gi>citeStructure</gi> parent must be an absolute XPath. If it is
```
sydb, line 47: Why no `<constraintSpec>` to enforce this? (I am not sure that `not( starts-with( ., '/') )` is sufficient, but am checking.)

sydb (commented):
Kinda hasty review, to try to give you some time, @hcayless.

ebeshero (approved):

P5/Source/Specs/citeData.xml
```diff
@@ -0,0 +1,40 @@
...
+  </classes>
+  <content>
+    <empty/>        
+  </content>
+  <attList>
+    <attDef ident="property" usage="req">
+      <gloss versionDate="2020-02-03" xml:lang="en">property</gloss>
+      <desc versionDate="2020-02-03" xml:lang="en">A URI indicating a property definition.</desc>
+      <datatype>
+        <dataRef name="anyURI"/>
```
peterstadler, line 25: While I'd always advise for using URIs, I wonder if we should enforce it here. In your own example in #1957 you put in "dc:title" as property – and I think that's ok since this mechanism can potentially be used to _label_ all sorts of features of a document.

P5/Source/Specs/citeStructure.xml
```diff
@@ -0,0 +1,76 @@
...
+    <memberOf key="att.citeStructurePart"/>
+  </classes>
+  <content>
+    <sequence>
+      <elementRef key="citeData" minOccurs="0" maxOccurs="unbounded"/>
+      <elementRef key="citeStructure" minOccurs="0" maxOccurs="unbounded"/>
+    </sequence>
+  </content>
+  <attList>
+    <attDef ident="delim" usage="opt">
```
peterstadler, line 25: `@delim` is only optional on the outermost `<citeStructure>`, otherwise it's impossible to parse the input, right? One might think of using different character sets for the `@use` attribute (to construct a citation string like "A3µ" without any delimiters) but I think that's even more complicated?

So, I think what I'm trying to say is that we need a Schematron constraint that asserts `@delim` on nested `<citeStructure>`s

P5/Source/Specs/citeStructure.xml
```diff
@@ -0,0 +1,76 @@
...
+      <gloss versionDate="2020-02-03" xml:lang="en">match</gloss>
+      <desc versionDate="2020-02-03" xml:lang="en">supplies an XPath expression which returns a sequence of nodes which are 
+        citable structural components. The expression may be absolute (beginning with <code>/</code>) or relative. 
+        <att>match</att> on a <gi>citeStructure</gi> without a <gi>citeStructure</gi> parent must be an absolute XPath. If it is
+        relative, its context is set by the <att>match</att> of the parent <gi>citeStructure</gi>. </desc>
+      <datatype>
+        <dataRef key="teidata.xpath"/>
+      </datatype>
+    </attDef>
+    <attDef ident="unit" usage="opt">
```
peterstadler, line 67: In the original FR in ticket #1957 the `@unit` attribute was a required attribute. I prefer it being optional because it's merely a label with no functionality, right?

If it's only for documentary purposes, what about adding `<desc>` to the content model of `<citeStructure>`?

peterstadler (changes requested):
Thanks for the nicely worked out ticket and PR!

My major concern is about the de facto required attribute `@delim` on `<citeStructure>`, and well, you'll need to rebase (or update) this PR to bring it up to date with current dev

P5/Source/Specs/citeData.xml
hcayless: I anticipated some objections to citeData, but not this one. I don't think they're really the same. `<equiv>` could of course be surgically altered to do the job we meant for `<citeData>`, but a) `<equiv>` gives me the creeps and b) it's definition would have to be changed: equivalence to its parent element is not at all what `<citeData>` provides. It specifies how additional metadata about the section could be extracted, which is not equivalence at all.

P5/Source/Specs/citeStructure.xml
hcayless: Agreed. Will do.

P5/Source/Specs/citeStructure.xml
hcayless: @peterstadler do you mean instead of using `@unit`? I'd like to keep it, because it's an enumerated list, but I've no objection to adding `<desc>` as well.

P5/Source/Specs/citeData.xml
hcayless: Well, "dc:title" is (most likely) a URI. If I used it, I'd have a `<prefixDef>` explaining how to expand it.

hcayless:
I've made the changes requested or responded to them, so I contend this is ready to be merged.

PR #1996, opened 2020-05-06
align `teidata.version` with Semantic Versioning Specification, closes #1993

### LINKED ISSUE #1993: discrepancy of P5 version number vs definition of version attribute of TEI
The version number of the _TEI Guidelines_ is (at least nowadays) stored in ./p5.xml at /TEI/teiHeader/fileDesc/editionStmt/edition/ref[2]. The value therein is copied from ./VERSION.

The current contents of that file (in the dev branch) is “4.1.0a”. Makes sense. The current release is 4.0.0, and we are anticipating the release of 4.1.0 in August or thereabouts by referring to the development branch copy as an alpha. IIRC, we plan to switch that to beta at the time when we go into “refrigeration” mode, and thus will use “4.1.0b” for a couple of weeks; and then the new release will be “4.1.0”. All well and good.

BUT the definition of the `@version` attribute of `<TEI>` is teidata.version, which in turn is defined as a token that matches "[\d]+(\.[\d]+){0,2}", and gets that definition from the Unicode standard. Thus the ‘a’ or ‘b’ at the end of our version number (which I submit could just as well be an ‘α’ or ‘β’) for P5 cannot be used on the `@version` attribute of `<TEI>` (or `<teiCorpus>`).

This strikes me as bad on general principles. But in specific, it causes a problem for building tei_customization which wants to use the version number of the p5.xml that was used as its source as both the `@version` of its own `<TEI>` and the `@version` of a TEI document that conforms to it. (And, I daresay, all Exemplars should do that. But that’s a different ticket.)

Possible solutions:

1. Stop using ‘a’, ‘b’, ‘α’, or ‘β’ in our version numbers. I don’t like this for a variety of reasons, most obviously that it means the version number of an in-development version looks like a released version number.
2. Strip the ‘a’, ‘b’, ‘α’, or ‘β’ off before using the VERSION number on `@version` of tei_customizatoin. I don’t like this because that means tei_customization is lying about its own pedigree.
3. Change the definition of `@version` of `<TEI>`, or `@version` of att.versioned to allow for an optional ultimate ‘a’, ‘b’, ‘α’, or ‘β’. I like this, but it does involve some work. First, we would have to decide on whether or not to keep `@version` of `<TEI>` and `<unicodeName>` the same (allowing a trailing letter where none should be allowed in the case of `<unicodeName>`), or have `<unicodeName>` inherit then modify the attribute so it is further restricted. Second, we would no longer be able to rely soley on the Unicode definition of a version number.

peterstadler:
This PR updates the regex used for defining the value of `teidata.version` to match the Semantic Versioning Specification. In fact, it's a slight modification of the provided regex at https://semver.org to work around the special flavor of XML Schema Regular Expressions (see https://www.regular-expressions.info/xml.html).

Additionally, the english prose was updated and some 'wrong'  `@version` attributes removed from the Guidelines source.


P5/Source/Specs/teidata.version.xml
```diff
@@ -9,21 +9,20 @@ See the file COPYING.txt for details
...
   <desc versionDate="2013-11-20"
          xml:lang="en">defines the range of attribute values which may be used to
-  specify a TEI or Unicode version number.</desc>
+    specify a version number following the Semantic Versioning Specification.</desc>
   <desc xml:lang="fr"
          versionDate="2007-06-12">définit la gamme des valeurs d'attribut
   exprimant un numéro de version TEI.</desc>
   <content>
-      <dataRef name="token" restriction="[\d]+(\.[\d]+){0,2}"/>
+    <dataRef name="token" restriction="(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-((0|[1-9]\d*|\d*[\-a-zA-Z][\-0-9a-zA-Z]*)(\.(0|[1-9]\d*|\d*[\-a-zA-Z][\-0-9a-zA-Z]*))*))?(\+([\-0-9a-zA-Z]+(\.[\-0-9a-zA-Z]+)*))?"/>
```
sydb, line 17: I have not hand-parsed the regular expression and seen that it makes sense. I did test it against all 71 test cases, and it correctly flags the first 31 as valid and the remaining 40 as invalid. So this looks like the correct W3C regular expression for SemVer.
So hurrah! to @peterstadler for a job well done. I think he should submit this to the SemVer folks, but hope he’ll coordinate with me, as I may have an entirely different regexp for them, as well.
That said, I think we should be much more restrictive for `@version` of `<TEI>` and `<teiCorpus>`. Something more akin to `4\.1(\.(0|[1-9]\d*))?(-(alpha|beta))?(\+([\-0-9a-zA-Z]+(\.[\-0-9a-zA-Z]+)*))?`.

P5/Source/Specs/teidata.version.xml
```diff
@@ -9,21 +9,20 @@ See the file COPYING.txt for details
...
+  <remarks versionDate="2020-05-04"
             xml:lang="en">
       <p>The value of this attribute follows the pattern specified by
-    the Unicode consortium for its version number (<ptr target="http://unicode.org/versions/"/>). A version number
-    contains digits and fullstop characters only. The first number
-    supplied identifies the major version number. A second and third
-    number, for minor and sub-minor version numbers, may also be
-    supplied.
+        the Semantic Versioning Specification (<ptr target="https://semver.org"/>). A version number
+    contains digits only and consists at least of three parts separated by a dot.
```
sydb, line 23: 1. “consists at least of three parts” should probably be “consists of at least three parts”.
2. But in truth (as alluded to above), I am not sure why a user should be **required** to specify the patch number in the `@version` of `<TEI>` or `<teiCorpus>`.

sydb (commented):
See individual comments.

duncdrum:
I have to say the question if TEI (guidelines and stylesheets) is an api according sem-ver definition is not straight forward to me. Actually i m leaning more towards saying yes. 

IIRC anything that is build metadata should be sorted alphabetically when determining version precedence, no?

sydb:
Maybe it’s me … I am certainly an old dog in a new trick game. But I cannot see TEI as an API. (In fact, I just yesterday got a virtual earful of how it is problematic that there is no Web API that allows one to get snippets of the source of the _Guidelines_.) An API “defines the kinds of calls or requests that can be made, how to make them, the data formats that should be used, the conventions to follow, etc.” [Wikipedia]. What call or request can a software system make to TEI?

As for build metadata, I wish that were so, @duncdrum, but what the SemVer spec says is “Build metadata does not figure into precedence”.

But I think we want build metadata included in our little universe, although we may want to go ahead with this ticket first, and then add it later.

raffazizzi:
VF2F subgroup thinks that it is appropriate to say that a TEI version number follows the Semantic Versioning Specification. 
Before we merge, however, could @peterstadler address @sydb's other comments?

sydb:
Council SVF2F NA group thinks:
- Still pending @peterstadler to make mods requested by me (per @raffazizzi)
- Still need to decide whether or not the value of `@version` should be required to match the (beginning of) the version of the schema
- Still need prose to explain what it means to be conforming to SemVer

sydb:
Council SVF2F NA group still thinks the 3 items listed on 07 May should be addressed. For 2nd bullet point we suggest that there be a soft requirement. I.e., prose that says “they should match”, not an actual schema test that will fail if they don’t.
@peterstadler should consider himself poked about the other two bullet points.

ebeshero:
Council meeting 2022-04-08: Add language to indicate that for the TEI Guidelines we’re inspired by the semantic versioning conventions but we’re applying our own TEI conventions to it (not following the custom for identifying “major” / “minor”.)

sydb:
Note: commit 713184682 did nothing except merge in the 664 commits that had been made to 'dev' branch in the interim.

sabineseifert:
Council at VF2F 16 March 2024:

Discussion
- Main problem: versioning system (major-dot-minor-dot-patch followed by optional ‘a’ or ‘b’, or supposedly ‘α’, ‘Β’, “alpha”, or “beta”) is not permitted by the datatype used for `TEI/@version` and `teiCorpus/@version`
- datatypes for other version numbers need rationalization as well
- `teidata.version` should allow for multiple variations
- Should the TEI provide a datatype that is not used by any actual element or attribute defined by the Guidelines themselves?

Council decision: Development of a comprehensive system for datatypes of version numbers
- An uber-datatype (e.g. `teidata.version` or `teidata.versionIndicator` or `teidata.versionNumber`) should be defined as an alternation of various specific datatypes which match particular kinds of version numbering systems. E.g.:
  - `teidata.semVer` would match the major-dot-minor-dot-patch-optional-dash-details kinda thing (`TEI/@version`, `teiCorpus/@version`)
  - `teidata.calVer` would match date-like version numbers (e.g.: 20240315)
  - `teidata.unicodeVer` would match theirs (`att.gaijiProp`)
  - `teidata.version.versionNumber` (e.g. 4.7.0)
- `teidata.version` would be defined as something like
```
<dataSpec ident="teidata.version">
<content>
<alternate minOccurs="1" maxOccurs="1">
        	 <dataRef key="teidata.version.semVer"/>
        	 <!-- follows semantic versioning syntax, see https://semver.org/ -->
        	 <dataRef key="teidata.version.calVer"/>
        	 <!-- follows calendar versioning syntax, see https://calver.org/ -->
        	 <dataRef key="teidata.version.unicodeVer"/>
        	 <!-- an enumerated list of Unicode versions (currently in att.gaijiProp)-->
        	 <dataRef key="teidata.version.versionNumber"/>
        	 <!-- for backwards comopatability -->
    	 	</alternate>
	 </content>
</dataSpec>
```
- original issue [#1993](https://github.com/TEIC/TEI/issues/1993): value of `@version` on `TEI/teiCorpus` will be changed to `teidata.version.semVer`
- SB: a TEI schema should permit only 1 possible value on `TEI/@version` or `teiCorpus/@version`, namely that schema’s version number. That schema number might be derived by appending the `schemaSpec/@ident` to the version number of the schema pointed to by `@source`. [This would be a new ticket.]

tuurma:
Buenos Aires F2F: action on @sydb to update the existing ticket with a redirect to SB’s branch (and corresponding ticket) and then close this PR

sydb:
Per the Buenos Aires F2F, Council is taking a different, more expansive approach, and this PR is to be closed. See [today’s comment](https://github.com/TEIC/TEI/issues/1993#issuecomment-2505132827) on the issue itself.

P.S. I personally think this branch should be kept, as the regex @peterstadler created for SemVer is worth keeping, as, IIRC, it is better than the one they supply.

Issue #1661, opened 2017-06-30
Need <surplus> in <subst>


cmm2209:
I have a manuscript in which the scribe abandons a word partway through (without indicating it should be deleted) when there's an error, and then writes the correct word immediately after.  It would be great to have the surplus element be available within the subst element.

lb42:
I am not sure I understand this. A subst element is defined as combining something signalled as a deletion with something signalled as an addition. What you seem to be describing is something that you think should be deleted but which is not signalled as such -- which is what surplus is for. Why not just use surplus (as in the example in the Glines)?

hcayless:
Because the correction is there. The error has been abandoned, but not erased. The implication is that the two are connected, as a deletion and correction are, but there's no deletion.

lb42:
OK. I suppose I'm reluctant to consider the word that follows the surplus as an addition, rather than just part of the text flow. If I am dealing with text like ths "something wib wibble something" I am quite comfortable marking the "wib" as a  "surplus" because it makes no sense to leave it there unmarked, but I am less comfortable marking the "wibble" as anything special. If I want to show that "wib" has been *corrected* to "wibble" why not use <corr> and <orig> ?

ebeshero:
Council F2F agrees after extensive discussion that we should permit one or more `<surplus>` or `del>` together with `<add>` in `<subst>`.

peterstadler:
@cmm2209 if you could provide an illustration for your example that would make a great addition to the Guidelines.

ebeshero:
I'm just reopening this ticket so it's more visible for developing an example.

peterstadler:
… and I purposely left it closed 😬

ebeshero:
Well, we can reopen if we get an example posted later. I guess we should close it ahead of the Guidelines release right now. Sorry.

cmm2209:
Thanks for working on this tweak, and sorry for the delayed response!  The real-world cases I’m working with are all in medieval Latin, which I presume wouldn’t be so helpful as an English one.  Here’s a possible example that I created:
Madness has in our age 
<subst>
<surplus>beca</surplus> 
<add>become</add>
</subst> 
some sort of lost truth.

sydb:
@cmm2209: Medieval Latin is fine, but this is even better. Do you have a source citation for it, or is it made up?

sydb:
(Because if it is just made up, shouldn’t it be
~~~xml
Truth has, in our age, 
<subst>
[...]

PR #2005, opened 2020-06-10
add surplus to subst

peterstadler:


peterstadler:
pinging @sydb and @ebeshero because I think this is good to go into the next release!

ebeshero:
@peterstadler Looks good to me! Nice work with the Schematron.

martinascholger:
@peterstadler - looks good to me. @sydb, do you approve this?

peterstadler:
> In as far as it goes, this update seems absolutely fine. No reason not to merge. (And I am impressed that @peterstadler caught the needed change in expted_results/, too :-)

Thanks :)

> That said, wouldn’t it be nice to have at least one example that uses `<surplus>` in `<subst>`? 

I find it hard to invent a good example and none was provided at the ticket. We might ask for one but IMHO it's not a blocker.

> And maybe some prose describing what it’s for in PHSU (11.3.1.5, Substitutions)? 

Tried to do so in e83cc60, borrowing @cmm2209's words.

> And maybe change the `<desc>` of `<substJoin>` to mention superfluous text? (I also notice that description is missing its Oxford comma.) E.g.:
> > identifies a series of possibly fragmented additions, deletions, superfluous text, or other revisions on a manuscript that combine to make up a single intervention in the text

Thanks again. Duly added! And I also found it necessary to alter the  `<desc>` of `<subst>` in this direction.

@ebeshero and @sydb may I ask you once more to have a look at these changes?

sydb:
> I find it hard to invent a good example and none was provided at the ticket. We might ask for one but IMHO it's not a blocker.

I agree it is not a blocker, but I don’t want us to forget, either. So my suggestion is to merge w/o example, but post a request for example(s) to the ticket (#1661) itself.

Will try to review prose soon. Thanks @peterstadler!

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -1326,6 +1326,13 @@ in the Eliot typescript mentioned above, there are --> several
                of the <soCalled>overlay</soCalled> levels implicit in the manuscripts of James
                Joyce's <title>Ulysses</title>. A fuller and more complex way of supporting such an
                approach is discussed in <ptr target="#PH-changes"/>. </p>
+            
+            <p>A special case of a substitution may consist of a superfluous word or phrase that is
```
sydb, line 1330: Good. Suggested mildly modified version:
~~~xml
<p>A special case of a substitution may consist of a superfluous word
or phrase that is silently replaced by some addition. E.g. a scribe
abandons a word (without indicating it should be deleted), and then
writes a replacement word immediately after. Here the encoder may
interpret this as an <soCalled>unmarked</soCalled> deletion which can
then be combined with a corresponding addition to a substitution.</p>
~~~

peterstadler:
revised the prose according to @sydb's proposal. @sydb what do you think of @ebeshero's [argument](#pullrequestreview-468017960) about the description of `<substJoin>`?

> However, I'm not sure I agree w/ @sydb that we needed to change anything in the desc for `substJoin.xml`. I think that one makes it unnecessarily difficult to puzzle through in the context of an element spec, _unless_ you're reading it in context with the Primary Sources chapter, as mostly I'm not sure people do. But doesn't the original desc carry the meaning we need for substJoin? I think the idea of superfluous text is implied sufficiently in "other revisions on a manuscript" besides additions / deletions.

sydb:
Eh. I am not convinced either way (on whether `<desc>` of `<substJoin>` should mention superfluous text or not), so let’s leave it as is (as @ebeshero recommends) for now.

peterstadler:
hopefully final ping for @ebeshero to have a look and merge if possible

Issue #567, opened 2015-08-26
Allow <relation> to occur intermittently within parent elements


TEITechnicalCouncil:
Currently in the guidelines, `<relation>` is allowed to occur within the following:

```
core: listBibl
namesdates: listEvent listNym listOrg listPerson listPlace listRelation
```

There are many uses for `<relation>` that enable users to define types of relationships between these given types of data. This tool is very useful for modeling data in ways that are consistent with ontological information and conversion to RDF.

However, even though `<relation>` is allowed to occur in any of the above elemental contexts, the way the content model is defined (below) it is not currently permitted to occur intermittently amongst the `<place>`, `<event>`, `<person>`, etc.  elements whose relationship to one another it is defining. Thus it is effectively only useable as a standoff element.

```

  (
     model.headLike*,
     ( model.placeLike | listPlace )+,
     ( relation | listRelation )*
  )

```

While use of `<relation>` as a standoff element may be acceptable for those using automated processes to gather and structure pre-existing data, it is not feasible for those who need to manually encode such data, as it requires a cumbersome and untenable process of scrolling back and forth between the different sections of a document.

While this may require added complexity in defining the content model, doing so would allow the element to reach its full functional potential and would almost definitely increase its usage.

What I propose is to change the model to allow for such uses of `<relation>` as in the following example:

```

         <listPlace>

            <place n="681" xml:id="Öst.">
               <placeName xml:lang="de">Österreich</placeName>
               <placeName xml:lang="de" full="abb">Öst.</placeName>
               <location>
                  <geo corresp="#gis_region_id-862"/>
               </location>              
            </place>

            <relation name="vaterregion" active="#Öst." passive="#Kä #OÖ #Bgl. #NÖ #Sa. #St. #Tir. #W. #Vlbg."/>

           <place n="181" xml:id="OÖ">
              <placeName xml:lang="de">Oberösterreich</placeName>
              <placeName xml:lang="de" full="abb">OÖ</placeName>
              <location>
                 <geo corresp="#gis_region_id-735"/>
              </location>
           </place>

            <place n="101" xml:id="Kä">
               <placeName xml:lang="de">Kärnten</placeName>
               <placeName xml:lang="de" full="abb">Kä</placeName>
               <location>
                  <geo corresp="#gis_region_id-708"/>
               </location>
            </place> 
            .....

```

Original comment by: @iljackb

raffazizzi:
[...]

PR #2021, opened 2020-08-06
Work on 567

sydb:
This change implements 567 by changing the content models of the following elements as per the description on the ticket. This fix has a non-backward compatible change: a member of model.milestoneLike is no longer allowed as the only child of `<listBibl>`. I consider the fact that such was allowed a corrigible error. (For supporting evidence I note that there are no examples of it in the _Guidelines_, nor am I aware of any prose discussing what the BLEEP it would mean.)

I did not remove `<listRelation>` from model.biblPart, per my [comment on the ticket of 2017-11-20](https://github.com/TEIC/TEI/issues/567#issuecomment-345599582), to wit that:

- the (only) effect would be to remove `<listRelation>` as a possible child of `<bibl>`;
- that would obviously break backward compatibility;
- `<bibl>`s can clearly have relations to at least other bibliographic items (n.b. `<relatedItem>`), if not people, places, and things.

Thus I think if we want to remove `<listRelation>` from model.biblPart, that would require another ticket and a separate discussion.

I did not change any examples that had `<relation>` as a direct child of `<listSomething>`, because there were none.

Elements changed:

- `<listBibl>`
- `<listEvent>`
- `<listNym>`
- `<listObject>`
- `<listOrg>`
- `<listPerson>`
- `<listPlace>`
- `<listRelation>`


lb42:
A bibl containing only milestone like elements and text is entirely possible. You might want to indicate formatting only. Please think again.

sydb:
Sorry, @lb42, that was my fault … the comment should have said “listBibl”, not “bibl”. While you are obviously correct that `<bibl>` should be allowed to have text & milestones (one wonders why not text & model.global, though), I think you will agree that a `<listBibl>` should have at least one bibl-like-thing in it.
I will see if I can edit the comment above …

P5/Source/Specs/listBibl.xml
```diff
@@ -34,22 +35,21 @@
...
+        <elementRef key="relation" minOccurs="1" maxOccurs="1"/>
+        <elementRef key="listRelation" minOccurs="1" maxOccurs="1"/>
+      </alternate>
+      <sequence minOccurs="1" maxOccurs="unbounded">
+        <classRef key="model.biblLike" minOccurs="1" maxOccurs="unbounded"/>
         <alternate minOccurs="0" maxOccurs="unbounded">
-          <elementRef key="relation"/>
-          <elementRef key="listRelation"/>
+          <classRef key="model.milestoneLike" minOccurs="1" maxOccurs="1"/>
+          <elementRef key="relation" minOccurs="1" maxOccurs="1"/>
```
peterstadler, line 49: Just out of curiosity: the addition of `minOccurs="1" maxOccurs="1"` to various `elementRef` is purely cosmetic (for clarity?), right?

peterstadler (approved):
changes look appropriate and my local build of the schemas was successful – thanks @sydb

P5/Source/Specs/listBibl.xml
sydb: Yes, that is my recollection. I sometimes think `@minOccurs` and `@maxOccurs` should not have defaults, but rather should be required.

Issue #1998, opened 2020-05-12
permit multiple seriesStmts in the TEI header


ebeshero:
Connected to #1870, Council recognizes that it would be helpful to permit multiple `<seriesStmt>`s in the TEI header, for cases where an item belongs to multiple series. We invite discussion here in case people have use cases, examples, or concerns.

martindholmes:
In our LEMDO project, we are planning for a situation in which the same edited play is published in multiple anthologies managed by different teams -- one default LEMDO anthology and at least one other specialist anthology, such as a Shakespeare-focused collection -- and also as a print publication. It would be really useful to have multiple `<seriesStmt>` elements for this.

sydb:
Main concern is whether or not `<seriesStmt>` should become a member of att.declarable, too, then. (In the use case we have imagined — this TEI document is a member of multiple different series — it is not necessary or even helpful to have `@default`. But many (if not most) other repeatable elements in the header are members of att.declarable so that different bits of the TEI document or corpus can say “that is the part of the header that applies to me”. So I am wondering if we allow multiple `<seriesStmt>s`, won’t people expect it to be usable the same way? … I can already hear @hcayless saying “maybe, but no one has asked for it”. : - )

ebeshero:
I think I can implement this one for our next release. A question for @martinascholger and @sydb : `<seriesStmt>` (zero or one) is permitted in `<biblFull>` and `<fileDesc>`. I think this ticket is *only* to address the `<seriesStmt>` contained by `<fileDesc>` , correct?

ebeshero:
Or, shall we permit multiple `<seriesStmt>` elements in `<biblFull>` as well?

martindholmes:
I think biblFull should mirror teiHeader, shouldn't it?

ebeshero:
[...]

PR #2024, opened 2020-08-09
allowing multiple seriesStmts in TEI Header for issue #1998

ebeshero:
Here's a fairly simple resolution to issue #1998, which I think should be ready for the upcoming 4.1.0 release. May I ask someone to review this and make sure it's okay to merge to dev?

### DIFF
--- a/P5/Source/Guidelines/en/HD-Header.xml
+++ b/P5/Source/Guidelines/en/HD-Header.xml
@@ -692,7 +692,7 @@ individual items may or may not be numbered.</item>
 or other items, similar in character and issued in sequence.</item>
 <item>A separately numbered sequence of volumes within a series or
 serial.</item></list>
-The <gi>seriesStmt</gi> element may contain a prose description or one
+A <gi>seriesStmt</gi> element may contain a prose description or one
 or more of the following more specific elements:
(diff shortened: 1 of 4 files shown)


martindholmes (commented):
＠ebeshero Looks perfect to me!

sydb:
Looks good to me, too. However, I am not fond of the use of the verb “applied” in HD:
> Multiple <seriesStmt> elements may be applied if …

I think “supplied” would be much better. Perhaps “used” or “encoded”.

ebeshero:
@sydb Okay! I'll push that change now.

peterstadler (approved):
Looks good to me!

lb42:
I haven't checked but presumably setiesStmt is now a member of att. Declarable?

Reluctantly using Outlook for Android<https://aka.ms/ghei36>

peterstadler:
> I haven't checked but presumably setiesStmt is now a member of att. Declarable? 

Yes

hcayless (approved):
Looks good to me.

lb42:
I just skimmed through the discussion on this ticket, and was a bit 
surprised by the implication that there might be sub-components of the 
TEI Header which were repeatable but were not members of att.declarable. 
Since the default assumption for any descriptive component of the header 
is that it concerns the whole of the associated text, I am finding it 
hard to think  of such a case.

However, that's no reason to disapprove of this proposed change.

martinascholger (approved):

PR #2031, opened 2020-09-05
Sydb 1921

### LINKED ISSUE #1921: Remove <schemaSpec> from model.divPart and add to small set of elements
schemaSpec should only be allowed in a small number of places. Currently it is allowed in:

> core: item note q quote said stage
> drama: castList epilogue performance prologue set view
> figures: cell figure
> header: change encodingDesc handNote licence scriptNote
> msdescription: accMat acquisition additions collation condition custEvent decoNote filiation foliation layout musicNotation origin provenance signatures source summary support surrogates typeNote
> namesdates: occupation
> tagdocs: specGrp
> textcrit: lem rdg
> textstructure: argument body div div1 div2 div3 div4 div5 div6 div7 epigraph postscript
> transcr: metamark

I would say that this should only be:
> encodingDesc, body, div

This enables it everywhere people actually use it, for example, in a back/div/schemaSpec but removes it from stupid places like 'provenance', 'epigraph', or dangerously 'specGrp'. This is because schemaSpec claims membership of model.divPart and it should not and just be added directly to div and body (keeping model.econdingDescPart is fine, it is needed there for some uses).

This issues specifically forces council to consider the ridiculousness of the willy-nilly application of classes through laziness.  I do *not* think that there needs to be any backwards compatibility concerns and it is a corrigible error with no deprecation needed. *If* someone comes forward and says "Oh no, my schemaSpecs inside 'metamark' are broken!", or whatever then either they should be corrected as to where to put them or council could add that in the next release. This would also be a good thing to show that the TEI is tightening the ridiculous places some elements are allowed.

sydb:
Reviewer(s):
The first two commits change nothing but whitespace. It is the third commit that makes the substantive changes. So what you want to do is compare the 3rd commit to the 2nd commit, not to the base, or you will be driven insane by the whitespace changes. (I tried to commit all the whitespace changes separately just to protect your sanity. I’m afraid I was not entirely successful, but I hope it still works. :-)  I hope gitHub lets you do that, but if it does not, `git diff -U0 a9f5b74..2268f3e` should do the trick.


peterstadler:
this is about #1921

sydb:
This PR is now blocked (at Council’s request) so that we can have a deprecation period. I will be creating a hand-deprecation in a different branch (“sydb-1921-deprecation” or some such) and a PR for that.

sydb:
The PR for deprecation is #2041.

ebeshero:
At VF2F May 2021: We think with @sydb that we should leave this pull request where it is for future reference once the deprecation ends, which will be 2021-10-23.

sydb:
@jamescummings, @lb42, @martindholmes, and @HelenaSabel —

Per our deprecation schedule, this PR should be folded in today or tomorrow. (And, correspondingly, #2041 should be backed out.)

I do not mind doing the work, but it should be reviewed by _someone_ else, eh?

martindholmes:
I don't know that the deprecation PR should be reversed. Wouldn't that make it appear in the history as though it was never deprecated?

sydb:
I did not mean “backed out” in any technical or `git` sense, just that we should remove it somehow so that a user doesn’t get 2 separate error messages for the same thing. I have not looked (doing encoder training almost all day), but I suspect that means just removing a `<constraintSpec>` from the `<schemaSpec>` tagdoc.

martindholmes:
With @sydb trying to get P5 build working again: this PR is ready for merging and will enable us to remove the deprecated Schematron rule which is failing the build.

martindholmes (approved):
Checked and ready for merge.

Issue #1919, opened 2019-09-15
category, taxonomy, and joinGrp should not have model.glossLike in their content models


jamescummings:
Elements category, taxonomy, and joinGrp have model.glossLike in their content model which gets them silly things like altIdent (rather than say altIdentifier). I think this is an accidental error confusing with wanting to provide the ability to gloss things in these elements with model.glossLike which is used in ODD writing. 

I think that taxonomy, category, and joinGrp should have model.descLike, gloss, and possibly equiv, but not altIdent. Optionally, one might want to consider providing altIdentifier on taxonomy and category to provide additional idno elements.

sydb:
Council VF2F subgroup thinks that removing `<altIdent>` from the content of those three non-TD elements is probably a good idea. Not high priority, as it only affects those using the TD module. However, we are not sure that the model changes @jamescummings suggests is the right way to do this. (Nor that it is the wrong way.) For example, it might be better to make a new class for “things that can go at the start of TD specification elements”, and reserve use of model.glossLike for transcription elements.
Note also that there are some TD elements that probably should not have `<altIdent>`, either; `<modelGrp>`, e.g.

martinascholger:
@sydb - can we close the issue?

sydb:
Well, no one (including Mr. Jenkins) has complained, so I guess so.

sydb:
What?
No, we can’t close this issue, the commit mentioned above (e13a62f) was made to special-purpose branch, not to dev.
I will issue a pull request as soon as I’ve merged current dev into it and tested it locally.

sydb:
OK, passed my local tests and awaiting on Mr. Travis. Please see PR #2032, and close this once that has been approved and merged in.

sydb:
Associated PR merged.

PR #2032, opened 2020-09-05
remove altIdent from some silly places

sydb:
This PR consists of two commits in order to try to make it easier to review. 
5e9988f has just whitespace changes; e13a62f has the actual content model changes.

P5 built and tested locally OK for me.

P.S. This is “take 2” because on my first attempt I messed up the git part of it, not the content model part. (I think I created new 1919 branch while I was still on 1921 branch, not dev, thus getting too many commits. Sigh.)


ebeshero (approved):
@sydb After poking around #1919 and making sure I wasn't confusing `<altIdent>` with `<altIdentifier>`, I fully approve this pull request. 

To help explain this to our new Council people (@lujessica), these elements were being defined as holding `model.glossLike` which includes `<altIdent>` as well as `<gloss>` and `<equiv>`. But `<altIdent>` is not appropriate for these three elements because it's intended for ODD-writing use in an elementSpec: `<altIdent>` is indicating an alternative identifier for a customized element. By using `model.glossLike` in the content model for these elements (`<category>`, `<joinGrp>`, and `<taxonomy>` which are not part of tagDocs/ODD writing), we are just encouraging misuse of `<altIdent>`. It's confusing enough that we have `<altIdent>` vs. `<altIdentifer>` as totally different elements anyway. This should help make things a little easier for people.

lujessica (approved):

PR #2041, opened 2020-10-25
deprecation of locations of schemaSpec

sydb:
Use a routine constraint and remarks in the tagdoc to limit where schemaSpec is allowed, with the intent of schema-enforcing it one year from now.

Reviewers: I have generated a [tagdoc](https://bauman.zapto.org/~syd/temp/TEItmp/ref-schemaSpec.html) for you.

I think some tweaks in the wording may be in order, but could be merged into dev first.

### DIFF
--- a/P5/Source/Specs/schemaSpec.xml
+++ b/P5/Source/Specs/schemaSpec.xml
@@ -46,7 +46,7 @@ $Id$
     </sequence>
   </content>
   <constraintSpec scheme="schematron" ident="deprecate_schemaSpec_in_bizarre_places">
-    <desc>enforces the depecration of <gi>schemaSpec</gi> as a child
+    <desc>enforces the deprecation of <gi>schemaSpec</gi> as a child
     of anything other than <gi>front</gi>, <gi>body</gi>,
     <gi>back</gi>, <gi>div</gi>, or any of the numbered division
     elements; it should be removed when the class memberships and
--- a/P5/Source/Specs/schemaSpec.xml
+++ b/P5/Source/Specs/schemaSpec.xml
@@ -7,7 +7,11 @@ $Date$
 $Id$
 -->
 <?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
-<elementSpec xmlns="http://www.tei-c.org/ns/1.0" xmlns:rng="http://relaxng.org/ns/structure/1.0" module="tagdocs" xml:id="SCHEMASPEC" ident="schemaSpec">
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0"
+             xmlns:rng="http://relaxng.org/ns/structure/1.0"
+             xmlns:sch="http://purl.oclc.org/dsdl/schematron"
+             xmlns:teix="http://www.tei-c.org/ns/Examples"
+             module="tagdocs" xml:id="SCHEMASPEC" ident="schemaSpec">
   <gloss versionDate="2007-07-04" xml:lang="en">schema specification</gloss>
   <gloss versionDate="2007-12-20" xml:lang="ko">스키마 명시</gloss>
   <gloss versionDate="2008-04-06" xml:lang="es">especificación de esquema</gloss>
@@ -18,7 +22,7 @@ $Id$
   <desc versionDate="2007-05-02" xml:lang="zh-TW">建立一個符合TEI標準的模型以及該模型文件。</desc>
   <desc versionDate="2008-04-05" xml:lang="ja">TEI準拠のスキーマや文書を示す．</desc>
   <desc versionDate="2007-06-12" xml:lang="fr">génère un schéma conforme à la TEI et la
-			documentation qui l'accompagne.</desc>
+                        documentation qui l'accompagne.</desc>
   <desc versionDate="2007-05-04" xml:lang="es">genera un esquema TEI-conforme y la documentación relativa.</desc>
   <desc versionDate="2007-01-21" xml:lang="it">genera uno schema TEI-conforme e la relativa documentazione</desc>
   <classes>
@@ -31,20 +35,40 @@ $Id$
   </classes>
   <content>
     <sequence>
-      
-        <alternate minOccurs="0" maxOccurs="unbounded">
-          <classRef key="model.glossLike"/>
-          <classRef key="model.descLike"/>
-        </alternate>
-      
-      
-        <alternate minOccurs="0" maxOccurs="unbounded">
-          <classRef key="model.oddRef"/>
-          <classRef key="model.oddDecl"/>
-        </alternate>
-      
+      <alternate minOccurs="0" maxOccurs="unbounded">
+        <classRef key="model.glossLike"/>
+        <classRef key="model.descLike"/>
+      </alternate>
+      <alternate minOccurs="0" maxOccurs="unbounded">
+        <classRef key="model.oddRef"/>
+        <classRef key="model.oddDecl"/>
+      </alternate>      
     </sequence>
   </content>
+  <constraintSpec scheme="schematron" ident="deprecate_schemaSpec_in_bizarre_places">
+    <desc>enforces the depecration of <gi>schemaSpec</gi> as a child
+    of anything other than <gi>front</gi>, <gi>body</gi>,
+    <gi>back</gi>, <gi>div</gi>, or any of the numbered division
+    elements; it should be removed when the class memberships and
+    content models are altered so that this restriction is enforced by
+    closed schema.</desc>
+    <constraint>
+      <sch:assert test="  parent::teix:egXML
+                        | parent::tei:encodingDesc
+                        | parent::tei:front
+                        | parent::tei:back
+                        | parent::tei:body
+                        | parent::tei:div
+                        | parent::tei:div1
+                        | parent::tei:div2
+                        | parent::tei:div3
+                        | parent::tei:div4
+                        | parent::tei:div5
+                        | parent::tei:div6
+                        | parent::tei:div7">WARNING: use of deprecated construct — the schemaSpec element will no longer be a valid child of <sch:value-of
+                        select="name(..)"/> as of 2021-10-23; instead, it should be a child of “encodingDesc”, “front”, “body”, “back”, or a division element.</sch:assert>
+    </constraint>
+  </constraintSpec>
   <attList>
     <attDef ident="start" usage="opt">
       <desc versionDate="2010-05-14" xml:lang="en">specifies entry points to the schema, i.e. which patterns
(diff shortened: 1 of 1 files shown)


martindholmes:
@sydb Looks good, but "depecration" should be "deprecation". I also don't think it's worth adding the numbered divisions; they're a bit of a hangover from many years ago, really. I would also add a context to the Schematron rule, just it case it gets separated from its elementSpec; I don't like relying on the container to provide the context for the Schematron. But only the typo matters.

sydb:
@martindholmes  —

Thank you!

1. [typo] Good catch; fixed; will push momentarily
1. [divN] Perhaps, but Council did not say to disallow `<schemSpec>` from numbered divs, and at this point it would be more work to reverse it
1. [Schematron context] Why do we have provided context capability if we don’t use it?

martindholmes:
@sydb Re #3: We don't use it *properly*, though, because the Stylesheets fail to process Schematron expressed at the attribute level properly. My own preference is to use the container as an organizin g principle -- put the constraint where you would expect to find it, and want to see it rendered in the Guidelines -- but be clear about the context. That way, if you move a Schematron rule, it won't break so easily at the processing stage.

sydb:
Hmmm… food for thought.

P5/Source/Specs/schemaSpec.xml
```diff
@@ -31,20 +35,40 @@ $Id$
...
+      </alternate>
+      <alternate minOccurs="0" maxOccurs="unbounded">
+        <classRef key="model.oddRef"/>
+        <classRef key="model.oddDecl"/>
+      </alternate>      
     </sequence>
   </content>
+  <constraintSpec scheme="schematron" ident="deprecate_schemaSpec_in_bizarre_places">
+    <desc>enforces the depecration of <gi>schemaSpec</gi> as a child
+    of anything other than <gi>front</gi>, <gi>body</gi>,
```
martinascholger: please add <encodingDesc> here as well.

P5/Source/Specs/schemaSpec.xml
```diff
@@ -31,20 +35,40 @@ $Id$
...
+        <classRef key="model.descLike"/>
+      </alternate>
+      <alternate minOccurs="0" maxOccurs="unbounded">
+        <classRef key="model.oddRef"/>
+        <classRef key="model.oddDecl"/>
+      </alternate>      
     </sequence>
   </content>
+  <constraintSpec scheme="schematron" ident="deprecate_schemaSpec_in_bizarre_places">
+    <desc>enforces the depecration of <gi>schemaSpec</gi> as a child
```
martinascholger: "This assertion" or "This" enforces ....

martinascholger (changes requested):

ebeshero:
@martinascholger : VF2F subgroup finds that far more often than not, the `<desc>` begins with a verb, taking the name of the tagdoc specification as the implied subject. So we do not think we should make this change. We may want to go the other way and change the few deviant descs to begin with a verb consistently.

martinascholger:
Pro consistency :-)

P5/Source/Specs/schemaSpec.xml
martinascholger: what I wanted to say here is please add `<encodingDesc>`

sydb: Good catch! Thank you. Fixing now …

P5/Source/Specs/schemaSpec.xml
sydb: No, our usual practice is that the description starts with a verb, as if the `@ident` were the subject.

martinascholger (approved):

jamescummings:
For some reason I'm just seeing this now.

Why the heck do we want to deprecate schemaSpec from appearing in encodingDesc? We worked really hard to *put* it there just a couple years ago. It is a natural place for those people who wish to embed a copy of their ODD in their document instance.

sydb:
Either you are misreading, @jamescummings, or we have a definite problem here. `<schemaSpec>` is (eventually) only to be **allowed** in `<encodingDesc>`, `<front>`, `<body>`, `<back>`, `<div>`, or `<divN>`. _I_ think that is what both the XPath test and content of the Schematron imply. (But the whole point of the review, of course, is to check that I got it right.)

P5/Source/Specs/schemaSpec.xml
```diff
@@ -31,20 +35,40 @@ $Id$
...
   </content>
+  <constraintSpec scheme="schematron" ident="deprecate_schemaSpec_in_bizarre_places">
+    <desc>enforces the deprecation of <gi>schemaSpec</gi> as a child
+    of anything other than <gi>encodingDesc</gi>, <gi>front</gi>, <gi>body</gi>,
+    <gi>back</gi>, <gi>div</gi>, or any of the numbered division
+    elements; it should be removed when the class memberships and
+    content models are altered so that this restriction is enforced by
+    closed schema.</desc>
+    <constraint>
+      <sch:assert test="  parent::teix:egXML
```
jamescummings: Should this be ancestor::teix:egXML ?

jamescummings:
Ah, apologies. I was reading it the wrong way around -- preventing it being in that location. Mea culpa. This happens all the time with SchemaTron for me. All looks good to me then, except I wonder if the first xpath pattern shouldn't be ancestor::teix:egXML rather than parent?

sydb:
Ooh. _Really_ good point, @jamescummings. No, I don’t think the test should be just ancestor::teix:egXML, which says just “don’t complain if this is an example”. The truly right thing to do, seems to me, is to deprecate use of `<schemaSpec>` outside those particular places whether in an example or not. Thus
~~~xml
      <sch:assert test="             parent::teix:egXML
                        | parent::tei:encodingDesc  | parent::teix:encodingDesc
                        | parent::tei:front         | parent::teix:front
                        | parent::tei:back          | parent::teix:back
                        | parent::tei:body          | parent::teix:body
                        | parent::tei:div           | parent::teix:div
                        | parent::tei:div1          | parent::teix:div1
                        | parent::tei:div2          | parent::teix:div2
                        | parent::tei:div3          | parent::teix:div3
                        | parent::tei:div4          | parent::teix:div4
                        | parent::tei:div5          | parent::teix:div5
                        | parent::tei:div6          | parent::teix:div6
                        | parent::tei:div7          | parent::teix:div7 ">WARNING: use
      of deprecated construct — the schemaSpec element will no longer be a valid child
      of <sch:value-of select="name(..)"/> as of 2021-10-23; instead, it should be a
      child of “encodingDesc”, “front”, “body”, “back”, or a division element.</sch:assert>
~~~
But this, of course, is predicated on the Schematron being fired when the element being tested is `<teix:schemaSpec>`, which I think is how this works for checking example validity, but would have to either re-read the code or test to be sure.

sydb:
I have not only changed the Schematron constraint so that it directly checks the parent element whether it is in the tei: or teix: namespace, I have also:
- changed the context of the constraint to be either `<tei:schemaSpec>` or `<teix:schemaSpec>`
- explicitly made the constraint `"nonfatal"`
- added a `@validUntil`
- added a `<desc type="deprecationInfo">`
- wordsmithed the `<remarks>`

Reminder to reviewers: You can read the current [tagdoc](https://bauman.zapto.org/~syd/temp/TEItmp/ref-schemaSpec.html) rather than generate it for yourself, if you like. I have also put the [deprecations](https://bauman.zapto.org/~syd/temp/TEItmp/DEPRECATIONS.html) page up.

jamescummings:
> Ooh. _Really_ good point, @jamescummings. No, I don’t think the test should be just ancestor::teix:egXML, which says just “don’t complain if this is an example”. The truly right thing to do, seems to me, is to deprecate use of `<schemaSpec>` outside those particular places whether in an example or not.

Hrmmm. I'm not sure, on principle, whether I agree. Wouldn't that imply that examples should contain valid TEI? We know from the fact that egXML has a `@valid` attribute that many (ok, some) might want to construct examples based on invalid TEI. (For example, to demonstrate what not to do....).  

But if this is nonfatal then it just ends up as a warning right? In that case it seems perfectly pragmatic to have. The minor annoyance of having it for those who really really want to do this is offset by the benefit of warning all those doing this accidentally.

jamescummings:
Oh and every time you refer to it you say: 

"encodingDesc, front, body, back, or a division element" 

(or something like that order). 

Is it that order in specific for a reason? the 'encodingDesc' seems misplaced. I suppose it is header elements first then the others, then divisions inside those.  But still looks strange to have encodingDesc first to me. I'm not arguing for an alphabetically-sorted list... that would be 'back, body, division element, encodingDesc, or front'  which just looks even sillier. Instead I might have said 'front, body, back, encodingDesc, or a division element'. The logic being there are three types of component (front, body, back), (encodingDesc), and (division elements) and for some reason it looks better that way around to me. 

This is a minor aesthetic quibble which you can ignore. Otherwise looks good to go!

sydb:
Addressing @jamescummings’ thoughts in prev 2 posts.

## but `<egXML>` is allowed to be invalid

Again, good point. I thought about this for a moment and figured I really should have `ancestor::teix:egXML[@valid eq 'false']` as one of the “this is OK” tests on the `<sch:assert>`. But then my brain broke when I thought about valid="feasible", and like the moron I am I just forgot about "false". But then thought that this is just a warning, and even in the case where you assert an example is marked as invalid it may not be a bad idea to remind you that this particular structure will cause a problem in the future.

That said, if you had an example (`@valid` defaulting to "true") that raised this warning and thought “OK, I know that now; I want that message to go away” and thus changed `@valid` to "false", you would be annoyed when it did not go away.

Sigh. I honestly don’t know what to do.

## order of elements in list of OK parents

You have surmised completely correctly that I think of this as three sets of elements — `<front>`, `<body>`, & `<back>`; `<encodingDesc>`; and divisions — and ordered them from “top to bottom”. But even as I did so I was waffling on whether `<encodingDesc>` should be 1st or middle in that list. So given that we have your vote for middle, and no one has objected, I am going to change it. … done in 580a89368.

P5/Source/Specs/schemaSpec.xml
```diff
@@ -45,28 +45,38 @@ $Id$
...
+			  | parent::tei:div           | parent::teix:div
+			  | parent::tei:div1          | parent::teix:div1
+			  | parent::tei:div2          | parent::teix:div2
+			  | parent::tei:div3          | parent::teix:div3
+			  | parent::tei:div4          | parent::teix:div4
+			  | parent::tei:div5          | parent::teix:div5
+			  | parent::tei:div6          | parent::teix:div6
+			  | parent::tei:div7          | parent::teix:div7"
+		    role="nonfatal">WARNING: use of deprecated construct — the “schemaSpec” element will no longer be a valid child of “<sch:value-of
+                        select="name(..)"/>” as of 2021-10-23; instead, it should be a child of “encodingDesc”, “front”, “body”, “back”, or a division element.</sch:assert>
```
martinascholger: I wonder if the date should be changed to February 2022?

sydb: Only if we defer to 4.3. From the beginning this was supposed to be a short (6–12 month) deprecation period.

martinascholger: ah, ok.

peterstadler (approved):
have tested locally (`make clean validate test`) and it builds ok. Nothing else that wasn't already addressed

PR #2053, opened 2020-10-30
type= on interpLike elements

### LINKED ISSUE #2012: revamp att.interpLike
This class is one reason for still having a couple of elements (interp interpGrp span spanGrp) just having `@type` and not being members of `att.typed`. Could we make a quick fix:

- delete `@type` from att.interpLike
- make `<interp>`, `<interpGrp>`, `<span>` and `<spanGrp>` members of att.typed

sydb:
Turns out we can’t fix #2012 the way we think it should be fixed because of Stylesheets bug 370. So here I have made an outright copy of the type= attribute we wanted to have in att.interpLike (and thus inherited by interp, interpGrp, span, and spanGrp) directly on interp, interpGrp, span, and spanGrp. I have included comments to explain what was going on, thus to make it easy to do the right thing when bug 370 is fixed. Each of this comments contains the string U+0023 370 for easier finding. (Although there may be a few false positives.)

### DIFF
--- a/P5/Source/Specs/spanGrp.xml
+++ b/P5/Source/Specs/spanGrp.xml
@@ -1,10 +1,15 @@
-<?xml version="1.0" encoding="UTF-8"?><!--
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
 Copyright TEI Consortium. 
 Dual-licensed under CC-by and BSD2 licences 
 See the file COPYING.txt for details
 $Date$
 $Id$
---><?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?><elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="analysis" ident="spanGrp">
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0"
+             xmlns:xi="http://www.w3.org/2001/XInclude"
+             module="analysis" ident="spanGrp">
   <gloss versionDate="2005-01-14" xml:lang="en">span group</gloss>
   <gloss versionDate="2007-12-20" xml:lang="ko">범위 집단</gloss>
   <gloss versionDate="2007-05-02" xml:lang="zh-TW">文字段群組</gloss>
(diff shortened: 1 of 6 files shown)


martindholmes:
@sydb The detest.log looks like it also needs changing (@default-is-in-list-req vs @default-NOT-in-list-req). Is it better to do that here, or after the merge into dev?

sydb:
I think mildly better to fix here. Will try to look at it tonight.

sydb:
No, of course it should be fixed in dev, the problem is because of a commit in dev that changed an error message, but did not change the corresponding test. I plan to commit a fix in minutes.

sydb:
OK, now I’m confused. I fixed detest.log in dev branch, tested it, and pushed it. I then switched to sydb-2012 and pulled. I then tried a `git merge -m "..." dev`, but it said I was already up-to-date. Which makes no sense, I _know_ there was a change to dev in the last few minutes, I made it: two lines of the file P5/Test/expected-results/detest.log were updated. Those updates are reflected on GitHub in dev, but are not reflected in this branch (either locally or GitHub). Any ideas what I might be doing wrong, here?

martindholmes:
＠sydb did you pull on dev locally? (git checkout dev, git pull, git checkout sydb-2012, git merge dev)?

sydb:
I had thought so, but apparently not. Thank you @martindholmes !

sydb:
Sorry for the delay, folks. But as you can see I have now fixed my ridiculous oversight. This PR is ready for review.

martinascholger (approved):

P5/Source/Specs/spanGrp.xml
```diff
@@ -22,14 +27,72 @@ $Id$
...
+  <attList>
+    <!--
+        The following <attDef> is a copy of the one in the
+        specifications of <interpGrp>, <span>, and <spanGrp>. It
+        *should* be part of the specification of att.interpLike rather
+        than in the specifications of the individual members of that
+        class. However, a bug in the Stylesheets means that won't work
+        at the moment. (See Stylesheets issue #370.)
+    -->
+    <attDef ident="type" usage="rec" mode="change">
```
peterstadler: the other specs have `mode="replace"` here. This might work, but actually it should be `mode="add"` since we are not inheriting from a class (`att.typed`) anymore?

peterstadler (approved):
ran `make clean validate test` locally and it works :)

Otherwise I have just a small remark about the `@mode` which should be aligned

P5/Source/Specs/spanGrp.xml
sydb: Good question. Short answer is “idunno”. Will try to test over next few hours.

sydb: - "add" causes duplicate definitions of `@type` (one with, one without the value list).
- "change" produces exactly the same results as "replace" **except** that the sample values list is dropped from the RELAX NG (and thus XSD) schema.
- Thus it seems "replace" is what we want.

Good catch, @peterstadler. Fixed in f0a2de56a.

PR #2059, opened 2020-11-16
Remove default value from attDef/@ns

### LINKED ISSUE #606: attDef/@ns has wrong default value
`<attDef>`/`@ns` has a default value of http://www.tei-c.org/ns/1.0, the TEI namespace. However, all TEI attributes are unprefixed, and therefore in no namespace (according to my reading of the spec, which says "the namespace name for an unprefixed attribute name always has no value" (http://www.w3.org/TR/REC-xml-names/#defaulting). So I believe the default namespace for TEI attributes should be an empty string.

Original comment by: @martindholmes

### LINKED ISSUE #2058: attDef/@ns still has the wrong default value
The discussion in #606 ended in the decision to remove the default value from attDef/@ns. But the guidelines still define a defaultVal for this attribute:

https://github.com/TEIC/TEI/blob/9e6d7fb88717edf3befe52f83aceb315ec8298a7/P5/Source/Specs/attDef.xml#L142

dmj:
Fix #2058
Fix #606

### DIFF
--- a/P5/Source/Specs/attDef.xml
+++ b/P5/Source/Specs/attDef.xml
@@ -139,7 +139,6 @@ $Id$
       <desc versionDate="2007-01-21" xml:lang="it">indica lo spazio del nome al quale appartiene
       l'attributo</desc>
       <datatype><dataRef key="teidata.namespace"/></datatype>
-      <defaultVal>http://www.tei-c.org/ns/1.0</defaultVal>
     </attDef>
   </attList>
   <exemplum xml:lang="en">


sydb (approved):
As I implied on the ticket, I think it is a no-brainer that we remove the `P5\Source\Specs\attDef.xml/elementSpec/attList/attDef[@ident eq 'ns']/defaultVal` (using a [FOXpath](https://archive.xmlprague.cz/2017/files/xmlprague-2017-proceedings.pdf#page=173) expression, hope you don’t mind). The only question is whether or not this requires deprecation. I think it is a (very) corrigible error, and deprecation is not needed. But I am often in the minority on this issue.

martindholmes:
I agree that this doesn't need deprecation.

martindholmes (approved):

PR #2068, opened 2020-12-04
Issue 2036 creating constraints on versionDate

MegJBrown:
@sydb and I have constrained versionDate to dates after 1970 and (in P5odds) equal to or before Tomorrow.

### DIFF
--- a/P5/Source/Guidelines/en/ST-Infrastructure.xml
+++ b/P5/Source/Guidelines/en/ST-Infrastructure.xml
@@ -1253,6 +1253,8 @@ entities in the XML DTD fragments.--></p>
           <specDesc key="teidata.duration.w3c"/>
           <!-- see ND <specDesc key="teidata.duration.iso"/> -->
           <specDesc key="teidata.temporal.w3c"/>
+          <!-- temporary working placement -->
+          <specDesc key="teidata.temporal.working"/>
           <!-- see ND <specDesc key="teidata.temporal.iso"/> -->
           <specDesc key="teidata.truthValue"/>
           <specDesc key="teidata.xTruthValue"/>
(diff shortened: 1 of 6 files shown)


martindholmes:
Looking at the regex:

`(19[789][0-9]|[2-9][0-9][0-9][0-9]).*`

I wondered if there should be a start anchor (^) at the beginning, and if you could use a quantifier for the repeated bit:

`^(19[789][0-9]|[2-9][0-9]{3}).*

Also, I'm not sure what the outer parentheses are doing. I'm probably missing something obvious, though.

sydb:
[test.rnc.txt](https://github.com/TEIC/TEI/files/5683722/test.rnc.txt)
@martindholmes asks three questions:

1. anchor at begininng (`^…`)?
1. quantifier instead (`[0-9]{3}` instead of `[0-9][0-9][0-9]`)?
1. why parens?

## 1. Anchor?

No! Because XSD  “implicitly anchors all regular expressions at the head and tail” starting with an anchor would mean that, in order to match the pattern, the value of the attribute would have to start with ‘^’ (U+005E, CIRCUMFLEX ACCENT). But since the value has to be an xs:date (or xs:dateTime), which has to start with a digit, **no** value would ever be valid. ☹ 

## 2. Quantifier?

Yes, I think `[0-9]{3}` and `[0-9][0-9][0-9]` will match exactly the same thing, and the former is a bit easier to read. I suspect I shy away from using quantifiers just out of brain-spillover with respect to curly braces in XSLT. But they do not even need to be escaped here in RELAX NG land.

## 3. Parens

The parens just serve as grouping for the alternation. (I wrote it as `(A|B).*`, but you could also say `A.*|B.*`. Same difference.)

sydb:
In a376bb2, just changed regular expressions to use quantifiers and fixed 2 incorrect `@versionDate` attributes that our new test caught!

Issue #2049, opened 2020-10-29
on altIdent


sydb:
(Note: This issue wraps in the problem initially posted in #1859, the problem later discovered in #1859, and one of the last comments of #1919, namely that “there are some TD elements that probably should not have `<altIdent>`, either; `<modelGrp>`, e.g.”.)

# Problems

The problems with `<altIdent>` are:
* It is still allowed in some places where perhaps it should not be.
* Its content model is silly: it should be `<dataRef key="teidata.xmlName"/>`.
* If the content model is changed, it could no longer be used is as a child of `<moduleSpec>` (with `@type` of "FPI"). Which would be a good thing, as its use there is simply incorrect — it does not match the definition of `<altIdent>` at all. But we need to find another way to represent the FPI of each module (which is needed for DTD support). Towards the end of this comment please find a table of the identifier of each module along with the heading of its chapter and its FPI.

# Plan

So my suggested plan of action is, roughly, as follows.

1. Decide on an alternate encoding of FPIs. My current favorite is to use an `<idno type="FPI">` exactly where we currently use `<altIdent type="FPI">`.
2. Decide which elements should, and should not, have `<altIdent>` in their content model.
3. Write Schematron constraint to manually deprecate use of `<altIdent>` in places identified in step 2 as no-nos.
1. Write another Schematron constraint to ensure that all `<altIdent>`s are xsd:NCName (the equivalent of teidata.xmlName) EXCEPT those that are a child of `<moduleSpec>` and (or or?) have a `@type` of "FPI".
4. Wait 6 months to a year.
5. During this time, if even 1 person asks for it, write a stylesheet to convert `moduleSpec/altIdent` to whatever we decided in step 1.
6. During this time modify all Stylesheets so that they can read and process both the method agreed in step 1 and the current method for finding the FPI. (That is, either will work. There are 62 occurrences of "altIdent\b" in 10 Stylesheet files, although a few of those are comments.)
1. Modify content models, classes, etc., so that `<altIdent>` is not allowed in the silly places identified in step 2.
1. Remove manual deprecation Schematron.
1. Modify content model of `<altIdent>` itself (to `<dataRef key="teidata.xmlName" minOccurs="1" maxOccurs="1"/>`)
1. Remove constrain-to-xsd:NCName Schematron.
1. Change our own use of `moduleSpec/altIdent` to whatever we decided in step 1.
1. Wait another 6 months to a year.
[...]

PR #2069, opened 2020-12-04
Warn if an attDef with a non-colonized @ident has a @ns attribute

dmj:
As pointed out in [1] attributes with an unprefixed ("non-colonized") name are in the empty namespace as opposed to unprefixed element names that are in the default namespace. A @‍ns attribute on an attDef with a non-colonized @‍ident thus has no effect. It would bind the default namespace to the value in @‍ns but it would not put the attribute in this namespace.

[1] https://github.com/TEIC/Stylesheets/issues/237#issuecomment-285476102

### DIFF
--- a/P5/Source/Specs/attDef.xml
+++ b/P5/Source/Specs/attDef.xml
@@ -39,6 +39,13 @@ $Id$
         <elementRef key="remarks" minOccurs="0" maxOccurs="unbounded"/>
     </sequence>
   </content>
+  <constraintSpec ident="attDefNsNCIdent" scheme="schematron">
+    <constraint>
+      <sch:rule context="tei:attDef[@ns]">
+        <sch:report test="not(contains(@ident, ':'))" role="nonfatal">Attribute: This definition of the @<sch:value-of select="@ident"/> attribute declares a namespace URI and uses an unprefixed name. By definition unprefixed attributes are in no namespace. The declared namespace URI will not have an effect.</sch:report>
+      </sch:rule>
+    </constraint>
+  </constraintSpec>
   <constraintSpec ident="attDefContents" scheme="schematron">
     <constraint>
       <ns xmlns="http://purl.oclc.org/dsdl/schematron" prefix="teix" uri="http://www.tei-c.org/ns/Examples"/>


hcayless:
Summarizing the state of things: we want to do this in tandem with fixing the processing problem, so we're not acting on it immediately. The solution implicitly proposed here is that if a namespace is being set with `@ns`, then it should have a prefix set on the ident, and this prefix will be bound to the namespace. This isn't the only possible solution, but if we go with it, then we'll implement it in the Stylesheets and merge this PR.

ebeshero:
Council SVF2F May 2021: We need to think about this. Questions and issues to follow up on:

*  How should we define an attribute within a specified namespace? We should use the `@ns` attribute, but should `@ident` have a prefix, or should it be inherited from the root element of the ODD, or via the `{Qname}` notation? 
* Is it really actually best to ensure we use a colon?
* We may need to define a new attribute for declaring the prefix. Or how do we declare the prefix so that we may refer to it, so that there's no name collision with a TEI attribute, for example? 
*  How is this related to Stylesheets ticket:  https://github.com/TEIC/Stylesheets/issues/237 ?

sydb:
For SVF2F discussion of 2022-04-02: An  [ODD file](https://github.com/TEIC/TEI/files/8400833/NAtst.zip) to play with. It demonstrates (what we hope is) a solution to the problem w/o any changes to TEI or Stylesheets.

raffazizzi:
Council decides to reject the PR because relying on namespace prefixes isn't a reliable solution. Prefixes are user defined and may be altered by XML processors at any time, while a prefix in `@name` wouldn't be altered and may break an ODD.

The solution linked by @sydb above, which relies on `<altIdent>` to allow `@name` to refer to the desired `@indent`, works for the time being to address the issue raised. Council will also open a new ticket to investigate a better solution that may, for example, introduce `@ns` to elements such as `<attDef>`.

ebeshero:
@sydb We just realized the build error problem is due to the positioning of the `sch:let` in the constraintSpec.

dmj:
Is there an issue that tracks the intended use of ```altIdent```?

ebeshero:
Hi @dmj : I think this is probably the ticket you are looking for: https://github.com/TEIC/TEI/issues/2049

sydb:
@dmj — No, I don’t think there is an issue that tracks the intended use of `<altIdent>`, just there is no issue to track the intended use of `<gloss>`. (The one @ebeshero links to is one about _removing_ a particular use case for `<altIdent>`; it doesn’t really talk about how to use it.)
But it is an interesting question as to whether or not the _Guidelines_, or perhaps some user documentation about ODD, should discuss or demonstrate how to do this.

PR #2075, opened 2020-12-09
Update SA-LinkingSegmentationAlignment.xml

bwbohl:
The first example for left() pointer – as far as I understand the pointing scheme – should read "second lb" instead of "first lb"

### DIFF
--- a/P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
+++ b/P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
@@ -887,9 +887,11 @@ which is either an <rs>XPATH</rs> as defined above or an
 <rs>IDREF</rs>, the value of an <att>xml:id</att>
 occurring in the document addressed by the base URI in effect
 for the pointer.</p>
-<p>Example: the pointer <code>#left(//gap[1])</code> 
+<p>Example: the pointer <code>#left(//supplied)</code>
 indicates the point between the first <code>lb</code> and the first
-  <code>gap</code> in the <ref target="#SATSXP-ex">example</ref> above.</p>
+<code>supplied</code> in the <ref target="#SATSXP-ex">example</ref> above.</p>
+<p>Example: <code>#left(//gap[1])</code> indicates the point immediately before
+the first <code>gap</code> element in line two and the string <code>si</code>.</p>
 <p>Example: <code>#left(line1)</code> indicates the point immediately before
 the <code><![CDATA[<lb n="1"/>]]></code> element.</p></div>
 


hcayless:
This is indeed an error, but the wrong solution, I'm afraid. I think the point (sorry) of the example was to indicate a location between two adjacent elements. The example should therefore either use `#left(//supplied)` and continue to reference the point immediately after the first `<lb>` and before the `<supplied>`. Or, if we keep `#left(//gap[1])`, it should say the point is between the letters `si` and the `<gap>` in line two.

bwbohl:
I think an example should be added in order to have both the cases you describe:
a) pointing exactly between two adjacent element tags
b) fix the description of the on  changed to precisely name the location between the tag and the `si` string

bwbohl:
updated the section accordingly, please have a second look @hcayless  or @sydb

P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
```diff
@@ -887,9 +887,11 @@ which is either an <rs>XPATH</rs> as defined above or an
 <rs>IDREF</rs>, the value of an <att>xml:id</att>
 occurring in the document addressed by the base URI in effect
 for the pointer.</p>
-<p>Example: the pointer <code>#left(//gap[1])</code> 
+<p>Example: the pointer <code>#left(//supplied)</code>
```
sydb, line 890: Concern 1 (with respect to this PR): the supplied XPath selects 0 or more nodes. There happens only to be 1 `<supplied>` in the example, so it happens to return only 1. But wouldn’t it be better to use an XPath that returns only 0 or 1 node? Previously paths like `//supplied[1]` where used, which I think is better. (But see below.)

Concern 2 (not part of this PR — if folks agree, we should make a separate ticket): The XPath `//duck[1]` does not select the first `<duck>` in document order, it selects _all_ `<duck>`s for which `position()` is 1, i.e. that are the first `<duck>` child of their parent. (The path `(//duck)[1]` selects the first `<duck>` in document order.) So while it is correct in the context of the example, it is not likely to be that useful in the context of an entire document. Thus I am thinking we should either exemplify a less fragile XPath, here, or at least explicitly mention that the example XPath is intended as an example to be thought of with respect to the XML fragment above, and not likely to be useful in actual documents.

P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
```diff
@@ -887,9 +887,11 @@ which is either an <rs>XPATH</rs> as defined above or an
...
 indicates the point between the first <code>lb</code> and the first
-  <code>gap</code> in the <ref target="#SATSXP-ex">example</ref> above.</p>
+<code>supplied</code> in the <ref target="#SATSXP-ex">example</ref> above.</p>
```
sydb, line 892: Problem that is not part of this PR: This entire section is poorly encoded (e.g., the “supplied” in this line is encoded as `<code>` rather than `<gi>`, the ID “SATSXP-ex” is attached to an anchor, not to the example, etc.). I am about to open another ticket to list (some of) these problems. Would you, @bwbohl, prefer I mark the new ticket as blocked by this one (thus avoiding conflict worries), or would you prefer to trust git to make merging relatively painless?

P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
```diff
@@ -887,9 +887,11 @@ which is either an <rs>XPATH</rs> as defined above or an
...
+<p>Example: <code>#left(//gap[1])</code> indicates the point immediately before
+the first <code>gap</code> element in line two and the string <code>si</code>.</p>
```
sydb, line 894: This is simply mis-written. (It boils down to “… the point immediately before A and the string B”, i.e. is missing the word “after” or “following”.) I think you mean something like
~~~xml
<p>Example: <code>#left(//gap[1])</code> indicates the point
immediately before the first <code>gap</code> element in line
two, and after the string <code>si</code>, which is
found immediately after the <tag>lb n="2"</tag>.</p>
~~~
That said, it might be clearer if written “after string B and before A”.

sydb (changes requested):
Thank you, @bwbohl.

P5/Source/Guidelines/en/SA-LinkingSegmentationAlignment.xml
bwbohl: agree with both concerns

hcayless:
Since this is essentially correcting a typo, I'm going to merge it and then add an additional qualifier to the path.

PR #2079, opened 2020-12-18
For Stylesheets 471:

sydb:
See [Stylesheets repo PR 481](https://github.com/TEIC/Stylesheets/pull/481) (which itself is about Stylesheets ticket 471, hence the name), in particular [today’s comment](https://github.com/TEIC/Stylesheets/pull/481#issuecomment-748269928) for details. 

Summary of changes herein:

 * Update CSS for new class "anchorlink";
 * Add new constraint to `<eg>` element in p5odds, so that in the Guidelines we will never have a leading or trailing newline in an `<eg>` (because the "bibliography Show all ⚓" line looks problematic if there is a trailing newline, and besides, the boxes look inconsistent when sometimes there is and someitmes there isn't);
 * Remove leading and trailing newlines from `<eg>`s in several chapters.

This PR should be considered in conjunction with Stylesheets PR 481.


npcole:
Looks good to me.  Shall I merge in?

martindholmes:
Perhaps wait for one more review?

PR #2091, opened 2021-01-29
add certainty to att.typed; close #1934

### LINKED ISSUE #1934: Categories of uncertainty
The TEI P5 Guidelines says (21.1.2):
> To record uncertainty in a more structured way, susceptible of at least simple automatic processing, the certainty element may be used.

It would be very desirable for purposes of automatic processing or visualization of uncertainty to be able to categorize annotations of uncertainty.

According to the book "Data Uncertainty and Important Measures" by Christophe Simon, Philippe Weber and Mohamed Sallak (John Wiley & Sons, 2018) and other literature, the epistemic uncertainty is most often divided into the following four categories:

- **Imprecision** corresponds to the inability to express the true value because the absence of experimental values does not allow the definition of a probability distribution or because it is difficult to obtain the exact value of a measure.

- **Ignorance** is related to the fact that information could have been incorrectly assessed by the person gathering or organizing the data. It is also possible that people, not fully sure about how to deal with data, ignore some information and generate uncertainty during the evaluation and decision processes.

- **Incompleteness** corresponds to the fact that not all situations are covered. Often it is impossible to know every possible option available.

- **Credibility** (discord) concerns the weight that an agent can attach to its judgment. This concept can be linked to that of biased opinions, which are related to personal visions of the landscape, which can make for significant variations between different groups and individuals, given their backgrounds.

We (https://providedh.eu/) postulate to add a new attribute "category" to the "certainty" element. It can be a semi-open list with the above mentioned categories. We do not want to limit others to this taxonomy of uncertainty.

tuurma:
use open list of values suggested by @mkozak in the OP

### DIFF
--- a/P5/Source/Specs/certainty.xml
+++ b/P5/Source/Specs/certainty.xml
@@ -22,16 +22,25 @@ $Id$
     <memberOf key="model.certLike"/>
     <memberOf key="model.global.meta"/>
     <memberOf key="att.scoping"/>
+    <memberOf key="att.typed"/>
   </classes>
   <content>
-    
       <alternate minOccurs="0" maxOccurs="unbounded">
         <classRef key="model.descLike"/>
         <classRef key="model.certLike"/>
       </alternate>
-    
   </content>
   <attList>
+    <attDef ident="type" usage="opt" mode="change">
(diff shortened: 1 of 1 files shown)


peterstadler (approved):
looks good to me :)

PR #2094, opened 2021-01-29
allow name in place; close #1484

### LINKED ISSUE #1484: person and place elements should allow 'name' child similar to org.
The org element allows nameLike elements as content, and in order to rationalise the named entity elements it would be useful if the name element was allowed in person and place. (Although when available it makes sense to use persName, placeName, and orgName in some customisations these may not be available.)

This would also benefit the TEI Simple customisation by allowing name instead of persName/placeName/orgName in person/place/org.

tuurma:
as requested in #1484 person and place elements should allow 'name' child similar to org 
person already does allow name, so only adding name to model.placeNamePart to cover places

### DIFF
--- a/P5/Source/Specs/name.xml
+++ b/P5/Source/Specs/name.xml
@@ -24,6 +24,7 @@
     <memberOf key="att.global"/>
     <memberOf key="model.nameLike.agent"/>
     <memberOf key="model.personPart"/>
+    <memberOf key="model.placeNamePart"/>
     <memberOf key="att.personal"/>
     <memberOf key="att.datable"/>
     <memberOf key="att.editLike"/>


peterstadler:
probably the easiest solution would be to add `<name>` directly to the content model of `<place>` like

```
element place
{
   att.global.attributes,
   att.global.rendition.attributes,
   att.global.linking.attributes,
   att.global.analytic.attributes,
   att.global.facs.attributes,
   att.global.change.attributes,
   att.global.responsibility.attributes,
   att.global.source.attributes,
   att.typed.attributes,
   att.editLike.attributes,
   att.sortable.attributes,
   (
      model.headLike*,
      (
         model.pLike*
       | ( model.labelLike | model.placeStateLike | model.eventLike | ***name*** )*
      ),
      ( model.noteLike | model.biblLike | idno | ptr | linkGrp | link )*,
      ( model.placeLike | listPlace )*
   )
}
```

PR #2096, opened 2021-01-30
allow name in place; close #1484

### LINKED ISSUE #1484: person and place elements should allow 'name' child similar to org.

tuurma:


tuurma:
@peterstadler I think you can press the button ;-)

Issue #2078, opened 2020-12-13
problems with match= (part of att.scoping)


sydb:
The `@match` attribute is defined as part of att.scoping.
1. The remarks are written as if it were defined in `<certainty>`.
2. It is defined as an XPath _expression_, but it should really only be a _pattern_ (v. 2) or _selection pattern_ (v. 3). (See below.)
3. Its definition refers to XPath v. 2.0. I’m thinking it should refer to [XSLT 3.0](https://www.w3.org/TR/xslt-30/#patterns) **selection pattern** instead.

#### below
From _XSLT 2.0 and XPath 2.0_ by @michaelhkay, 4th ed., pp. 681–2:

> A pattern is used in XSLT to define a condition that a node must satisfy in order to be selected.
>  … every pattern is a valid XPath expression, but not every XPath expression is a valid pattern. It wouldn’t make any sense to use the expression `2+2` as a pattern, for example — which nodes would it match? 

Note that patterns are defined by XSLT, not by XPath. (Used by other specs, too, though; in particular XProc and Schematron, I think.)

peterstadler:
@sydb is your point 2 ("defined as an XPath expression") also applicable to `teidata.xpath`? Is that also about _patterns_ not _expressions_?

ebeshero:
@peterstadler @sydb If I remember correctly, we worked with `teidata.xpath` on the `unitDecl`, unit and measure tickets, and I believe there we want to keep the XPath options a little more open to being expressions or patterns.

ebeshero:
VF2F: I will try to implement the small changes recommended here in time for upcoming release.

ebeshero:
With this pull request: https://github.com/TEIC/TEI/pull/2099 , I think I've addressed all the requested changes in this ticket.

sydb:
@peterstadler:
No, at the very least `@value` of `<param>` and `@formula` of att.formula (which @ebeshero alluded to) require full XPath expressions.
But if you are poking at the idea that teidata.xpath should be divided into teidata.xpath and teidata.selectionPattern (or whatever), seems reasonable to me.

PR #2099, opened 2021-01-30
updating match attribute to mention XSLT 3 specs, broaden reference b…

ebeshero:
…eyond certainty re #2078

### DIFF
--- a/P5/Source/Guidelines/en/CE-CertaintyResponsibility.xml
+++ b/P5/Source/Guidelines/en/CE-CertaintyResponsibility.xml
@@ -242,8 +242,8 @@ of the tagging throughout a document, rather than (as discussed so far) in one
 particular instance. The <att>match</att> attribute may be used to
 supply a pattern identifying the portion of a document
 concerning which certainty is being expressed. The value of the
-<att>match</att> attribute is an XPath expression <!--XSLT2 pattern,-->
-using the syntax defined in <ptr target="#XSLT2"/>. In the following
+<att>match</att> attribute is an XPath expression <!--XSLT3 pattern,-->
+using the syntax defined in <ptr target="#XSLT3"/>. In the following
 example, we wish to indicate a low degree of confidence
 that the <gi>persName</gi> elements used throughout the whole document
 have been correctly applied:
@@ -276,8 +276,8 @@ elements which have a <att>type</att> value of <code>checked</code>:
 
 If an element in a document is matched by more than one match
 expression, then the most specific pattern applies. <note place="bottom">Specificity of
-pattern matching is defined further in the XSLT2 reference cited above
-(see <ptr target="http://www.w3.org/TR/xslt20/#conflict"/>)</note> As a
+pattern matching is defined further in the XSLT3 reference cited above
+(see <ptr target="https://www.w3.org/TR/xslt-30/#default-priority"/>)</note> As a
 simple case, if both the preceding <gi>certainty</gi> elements were
 present in the same document, a <gi>persName</gi> occurring within a
 <tag>div type="checked"</tag> element would potentially match both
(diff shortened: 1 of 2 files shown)


ebeshero:
Seeking a review in time to merge this, to address #2078 in time for the Guidelines 4.2.0 release. Please make sure the references to the XSLT 3 specs are accurate and the tiny change to remarks is okay!

ebeshero:
(I assigned four reviewers mainly just because we're getting close to refrigeration: probably if only one or two review it's okay. I just want someone to double-check the language and make sure it's accurate.)  @sydb @peterstadler @martinascholger  @hcayless

P5/Source/Guidelines/en/CE-CertaintyResponsibility.xml
```diff
@@ -242,8 +242,8 @@ of the tagging throughout a document, rather than (as discussed so far) in one
 particular instance. The <att>match</att> attribute may be used to
 supply a pattern identifying the portion of a document
 concerning which certainty is being expressed. The value of the
-<att>match</att> attribute is an XPath expression <!--XSLT2 pattern,-->
-using the syntax defined in <ptr target="#XSLT2"/>. In the following
+<att>match</att> attribute is an XPath expression <!--XSLT3 pattern,-->
```
sydb: I do not think this is correct. I think the value of `@match` is a selection pattern. See [the XSLT spec](https://www.w3.org/TR/xslt-30/#dt-pattern). Thus:
~~~xml
The value of the <att>match</att> attribute is an XSLT selection pattern using the syntax defined in <ptr target="#XSLT3"/>.
~~~

P5/Source/Specs/att.scoping.xml
```diff
@@ -38,8 +38,8 @@ $Id$
         </egXML>
       </exemplum>
 
-      <remarks versionDate="2016-03-15" xml:lang="en">
-        <p>The expression of certainty applies to the nodeset identified by the value of the
+      <remarks versionDate="2021-01-30" xml:lang="en">
+        <p>The expression applies to the nodeset identified by the value of the
```
sydb: I am uncomfortable with this. “The expression” seems ambiguous. I would prefer something like “The semantics of the element upon which this attribute appears apply to the …”.

sydb (commented):
Nothing to do with the changes you have made so far, @ebeshero, but I also realize that the `<remarks>` in att.scoping are in the wrong place. They are remarks about both `@match` and `@target` (and thus, in some sense, about the entire class) but re inside the `<attDef ident="match">`; they should be a child of the `<classSpec>`, I think.

Furthermore, the description of `@target` leaves a bit to be desired:
> points at one or several elements or sets of elements by means of one or more data pointers, using the URI syntax.

I propose instead

> points at one or more sets of zero or more elements each.

Is it necessary to say anything about the syntax here?

P5/Source/Specs/att.scoping.xml
ebeshero: @sydb If I'm moving these remarks so they're a child of `classSpec`, it won't make sense to be discussing just the `@match` attribute, but rather the whole element. I'm trying just this (slightly simpler) phrasing: 
`The semantics of this element apply to the nodeset identified by the value of the
      <att>target</att> attribute, possibly modified by the value of the
      <att>match</att> attribute.`

ebeshero: @sydb I totally agree about simplifying the `desc` for `@target`...implementing that.

ebeshero:
@sydb I've updated my PR with some fresh edits thanks to your suggestions. Take another look?

ebeshero:
Realized I needed to position the `<remarks>` to follow the `attDef`s. Fixed now.

ebeshero:
@sydb @hcayless @peterstadler @martinascholger This PR should be ready to “press the button” and merge in...Just want to make sure it is not forgotten!

sydb:
Review now …

sydb:
(I had not forgotten, @ebeshero, just working on other PRs. :-)
Had to pull in changes from dev to get it to build, but seems to be working now.

PR #2102, opened 2021-02-04
Issue 2054 ruby

martindholmes:
This is now complete: pointers to the new prose from other locations in the Guidelines have been added, all suggested fixes have been made, and the examples have all been validated with a schema built from the source. A build from this branch is available on the UVic Jenkins for easy proofing. Let's try to get this merged before the freeze on the 9th.


P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1023,164 +1023,365 @@ overshadows the historical content</quote>.</egXML>
...
+        <figure><graphic width="10rem" url="Images/rbTEI.png" mimeType="image/png"/></figure>
+        
+        <p>In Japanese, furigana (振り仮名) ruby annotations are often used to provide pronunciation 
+          guidance for readers; characters from the largely phonetic hiragana or katakana syllabaries
+          accompany Chinese characters, like this:
+        </p>
+        
+        <figure><graphic width="22rem" url="Images/rbNhkEasy.png" mimeType="image/png"/>
+          <head>The first line of a news story from NHK News Web Easy intended for Japanese
+            learners, in which every Chinese character has a ruby gloss.</head>
```
sydb: If it is in fact the case (which to my monolinguistic eyes appears to be true), then I think “… in which each Chinese character has a ruby gloss directly above it.” would be better.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -2664,6 +2664,21 @@ referring to the zone marked in purple on the right
             annotation. </p>
       </div>
    </div>
+   
+   
+   <div  type="div2" xml:id="PHRB">
+      <head>Transcription and Ruby</head>
+      
+      <p>These Guidelines also provide special elements to support the encoding of
+         Ruby annotations, which are common in East Asian textual traditions. These
```
martinascholger: In section 3.4.2 "ruby" is written in lower case (in running text).

martinascholger (changes requested):

martinascholger:
@martindholmes: could you please change the reference in the introduction of chapter 3 from 3.4.1 Terms and Glosses to 3.4 Terms and Glosses, Ruby Annotations, etc.

P5/Source/Guidelines/en/PH-PrimarySources.xml
martindholmes: Done! Thanks Martina.

martindholmes:
Fixed the capitalization too.

martinascholger (approved):

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1023,164 +1023,365 @@ overshadows the historical content</quote>.</egXML>
...
+              <rb>大学</rb>
+              <rt place="above">だいがく</rt>
+            </ruby>
+          </egXML>          
+          In the example above, the word <mentioned>大学</mentioned>
+          (<mentioned>daigaku</mentioned> = university) is provided
+          with a phonation gloss in hiragana. The full gloss is
+          applied to the complete word. However, it might instead be
+          broken down by character:
+          <egXML xmlns="http://www.tei-c.org/ns/Examples">
```
sydb, line 1242: Do we have to worry about the fact that this example encodes a space between “大” and “学” (and the corresponding ruby text)? Same concern for next example.
Note that we cannot just remove the newline between `</ruby>` and `<ruby>`, the Stylesheets will still render them on separate lines anyway. So if we do care, we need to mention it in the prose, or combine them in a `<w>` element.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1023,164 +1023,365 @@ overshadows the historical content</quote>.</egXML>
...
+          is glossed with  two different pronunciations: <q>biriādo</q> (its English equivalent)
+          and <q>dakyū</q>, a phonation guide for the first two characters.
+          
+          <figure>
+            <head><mentioned>Billiard hall</mentioned> with two ruby glosses. 
+              <ref target="http://school.nijl.ac.jp/kindai/NIJL/NIJL-01116.html#37">国文学研究資料館所蔵：：英国／龍動新繁昌記</ref>.</head>
+            <graphic url="Images/billiardhall.jpg" width="149px" height="276px"/>
+          </figure>
+          
+          This example is intriguing in that the right-side ruby
```
sydb, line 1286: I don’t think “intriguing” is quite the right word; I think “interesting” would be better.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1023,164 +1023,365 @@ overshadows the historical content</quote>.</egXML>
...
+                    <rb>打</rb>
+                    <rt place="right">ダ</rt>
+                  </ruby> 
+                  <ruby>
+                    <rb>球</rb>
+                    <rt place="right">キウ</rt>
+                  </ruby>
+                  場
+                </rb>
+                <rt place="left">ビリヤード</rt>
```
sydb, line 1310: Do we need to mention that ‘ー’ (U+30FC, Katakana-Hiragana Prolonged Sound Mark) is is written horizontally in horizontal text and vertically in vertical text?

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1023,164 +1023,365 @@ overshadows the historical content</quote>.</egXML>
...
+                <rt place="right" target="#c2">キウ</rt>
+              </ruby> 
+              
+              [...]
+            </p>
+          </egXML>
+          
+        </p>
+        
+        <p>The current support for ruby is rudimentary, and in future releases of the
```
sydb, line 1356: I would prefer something like
~~~xml
<p>The current support for ruby annotations is
rudimentary, and future releases of these
<title>Guidelines</title> may provide further
recommendations.</p>
~~~
although I don’t think that’s quite right, either.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -2664,6 +2664,21 @@ referring to the zone marked in purple on the right
...
       </div>
    </div>
+   
+   
+   <div  type="div2" xml:id="PHRB">
+      <head>Transcription and Ruby</head>
+      
+      <p>These Guidelines also provide special elements to support the encoding of
+         ruby annotations, which are common in East Asian textual traditions. These
+         elements provide yet another method of capturing a specific type of annotation,
```
sydb: This sentence “These elements provide … text and annotations.” is a bit of a run-on. It should at least be punctuated with semicolons to help the reader see the major clauses, but better yet divvied into 2 or 3 sentences.

P5/Source/Specs/ruby.xml
```diff
@@ -0,0 +1,57 @@
...
+<gloss versionDate="2021-01-31" xml:lang="ja">ルビのためのコンテナ要素。</gloss>
+<desc versionDate="2020-02-28" xml:lang="en">contains a
+  passage of base text along with its associated ruby gloss(es).</desc>
+<desc versionDate="2021-01-31" xml:lang="ja">ルビ及びその対象となるテキストを含む。</desc>
+<classes>
+  <memberOf key="att.global"/>
+  <memberOf key="att.typed"/>
+  <memberOf key="model.phrase"/>
+</classes>
+  <content>
```
sydb, line 21: I am feeling a bit sheepish about this, as I think it is too late to do anything now. But this content model is clearly too restrictive. Elements like `<addSpan>`, `<certainty>`, and `<metamark>` (all members of model.global) surely should be allowed inside `<ruby>`.

I am also thinking maybe `<ruby>` should be a member of att.breaking.

I do not think either of these should be addressed now. Separate tickets for 4.x.0 (where x ≥ 3), I guess.

P5/Source/Guidelines/en/CO-CoreElements.xml
martindholmes: There are no spaces between characters in Japanese, so the space in the encoding is irrelevant, I think. The original proposal has lots of examples like this.

sydb: Check.

P5/Source/Guidelines/en/CO-CoreElements.xml
martindholmes: It's not actually true that each character has a gloss; you've found a different problem. In this particular case, it's each WORD consisting of Chinese characters that has a ruby gloss. The ruby is attached to the word rather than the characters in this case. I'll update accordingly.

P5/Source/Guidelines/en/CO-CoreElements.xml
martindholmes: intriguing: "arousing one's curiosity or interest; fascinating." I think it means exactly what I think it means, in this case.

P5/Source/Guidelines/en/CO-CoreElements.xml
martindholmes: There's surely no-one who needs to understand this to whom it's not obvious; and no-one to whom it's not obvious who needs to understand it. It's not something specific to ruby; it's just general knowledge about the Japanese writing system.

P5/Source/Guidelines/en/CO-CoreElements.xml
martindholmes: I don't see anything wrong with the sentence as it stands, and since you don't think your own suggestion is right either, I don't see that it's worth changing it. That sentence is only going to be there for a single release of the Guidelines, I hope.

P5/Source/Guidelines/en/CO-CoreElements.xml
sydb: I was not trying to suggest that “intriguing”  _means_ the wrong thing, only that it doesn’t sound right. Pretentious would be too strong a word to describe my complaint, but it’s on the right track. But if you like that word, and don’t think it harder for non-native readers, how about making the sentence a little more approachable by swapping “because” in for “in that”?

P5/Source/Guidelines/en/CO-CoreElements.xml
sydb: Check.

P5/Source/Guidelines/en/PH-PrimarySources.xml
martindholmes: This is a short paragraph already divided into three sentences, of which the longest is has only 39 words. AFAICS there's nothing syntactically wrong with the longest; it's not a run-on sentence (https://www.grammarly.com/blog/run-on-sentence-basics/). I think if you believe it should be changed, you need to provide your suggested version.

P5/Source/Specs/ruby.xml
martindholmes: This is a bit worrying. The complex stuff should really appear in rb rather than in ruby, no? Or am I misunderstanding?

P5/Source/Guidelines/en/CO-CoreElements.xml
martindholmes: The phrase "in that" appears 43 times in the Guidelines, so I think if it's likely to be confusing, then we have to change a lot of other locations. I generally favour simpler prose, but I don't think the current formulation constitutes the sort of impenetrable Britishism we used to have too much of. Perhaps what you're responding to is the fact that it makes the assertion that something is interesting at all -- maybe that's not for the Guidelines to say, and we should instead have "In this example, the right-side ruby...". Would that be better?

sydb: Hmmm … this is tweak that, if needed, should not delay merging.

P5/Source/Specs/ruby.xml
sydb: But what if it is not actually part of the base text, but rather a mark (read `<metamark>`) on the manuscript that indicates which ruby annotation should be associated with which ruby base character? Or you want to use the `<certainty>` element to express uncertainty about a ruby interpretation? Or the added span of text is (perhaps incorrectly or inappropriately) inserted in between the ruby base and ruby annotation?

All of these are quite reasonable, but also things that @hcayless would say are not (yet) needed as no one has asked for them (yet).

P5/Source/Guidelines/en/PH-PrimarySources.xml
sydb: First crack:
~~~xml
<p>These Guidelines also provide special elements to support the
encoding of ruby annotations, which are common in East Asian textual
traditions. These elements provide a method of capturing a specific
type of annotation, in addition to the generic methods like the
<gi>note</gi> or <gi>interp</gi> elements. Both the specific and
general methods should integrate well with the transcriptional
elements described above, allowing authorial and scribal features to
be captured in conjunction with ruby base text and annotations. See
<ptr target="#COHTGRB"/> for more information about these
elements.</p>
~~~

P5/Source/Specs/ruby.xml
martindholmes: I think some of these projected issues might be based on a misunderstanding, and without specific examples I don't know how to respond to them. We know that after this release there will be at least seven tickets arising out of @duncdrum's examples that will require either enhanced examples/recommendations and/or schema adjustments, and the certainly issue perhaps makes an eighth (although I think what you are expressing uncertainty about in most cases is either going to be something in the rb or the rt). There will also be more from the Japanese community as they develop their own guidelines based on this release. But without genuine examples it's difficult to imagine what needs to be done. The only question that matters, I think, is whether at this point we are building any decision into the Guidelines which we later have to roll back in such a way as to break backwards compatibility; an overly-restrictive content model is not such a problem, where the reverse would be. 
Regarding the question of uncertainty about which rb/rt relationship is correct, a simple choice element is an obvious solution, or the use of the `@exclude` attribute.

sydb: > an overly-restrictive content model is not such a problem, where the reverse would be

Agreed, completely. Which is why I am not suggesting any changes at this time.

P5/Source/Guidelines/en/PH-PrimarySources.xml
martindholmes: Done! Thanks.

P5/Source/Specs/ruby.xml
martindholmes: My first job as soon as the merge is done is going to be to go through (with your help, I hope) all @duncdrum's tickets on the rubyForTEI repo and turn them into appropriate issues on the TEI repo, tagged for a future milestone. We should also refine these concerns into specific tickets, hopefully with real examples from the scholars involved.

PR #2105, opened 2021-02-10
add new class "att.predicate"

### LINKED ISSUE #1925: <equiv> should get the @predicate attribute
The `<equiv>` element should get the same or similarly defined `@predicate` attribute that is available on `<model>`.  This would enable filtered equivalences for all sorts of uses, but for example being able to map uses of a single element to different conceptual models. i.e. `<name type="person">` could map to one concept in (say) CIDOC CRM or other ontologies but `<name type="place">` could map to another. These would be documented in the ODD not in the document instance.  This strikes me as a very flexible addition to equiv that is easy to implement.

peterstadler:
* move `@predicate` from `<model>` to the new class
* make `<equiv>` a member of this new class as requested in #1925

### DIFF
--- a/P5/Source/Specs/att.predicate.xml
+++ b/P5/Source/Specs/att.predicate.xml
@@ -36,9 +36,9 @@ See the file COPYING.txt for details
             concepts of "PERSON" and "PLACE" depending on the XPATH expression given in <att>predicate</att></p>
         <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#NONE">
             <elementSpec ident="name" mode="change">
-                <equiv name="PERSON" predicate="@type='person'"
+                <equiv name="PERSON" predicate="@type eq 'person'"
                     uri="http://www.example.com/entities/person"/>
-                <equiv name="PLACE" predicate="@type='place'"
+                <equiv name="PLACE" predicate="@type eq 'place'"
                     uri="http://www.example.com/entities/place"/>
             </elementSpec>
         </egXML>
--- a/P5/Source/Specs/att.predicate.xml
+++ b/P5/Source/Specs/att.predicate.xml
@@ -18,7 +18,7 @@ See the file COPYING.txt for details
         </attDef>
     </attList>
     <exemplum xml:lang="en">
-        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#UND">
+        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#NONE">
             <model predicate="parent::person"
                 behaviour="inline">
                 <desc versionDate="2015-08-21"
@@ -30,7 +30,7 @@ See the file COPYING.txt for details
     <exemplum xml:lang="en">
         <p>The following example declares that the <gi>name</gi> element can be mapped to (is equivalent to) the external 
             concepts of "PERSON" and "PLACE" depending on the XPATH expression given in <att>predicate</att></p>
-        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#UND">
+        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#NONE">
             <elementSpec ident="name" mode="change">
                 <equiv name="PERSON" predicate="@type='person'"
                     uri="http://www.example.com/entities/person"/>
--- a/P5/Source/Specs/att.predicate.xml
+++ b/P5/Source/Specs/att.predicate.xml
@@ -32,8 +32,8 @@ See the file COPYING.txt for details
         </egXML>
     </exemplum>
     <exemplum xml:lang="en">
-        <p>The following example declares that the <gi>name</gi> element can be mapped to (is equivalent to) the external 
-            concepts of "PERSON" and "PLACE" depending on the XPATH expression given in <att>predicate</att></p>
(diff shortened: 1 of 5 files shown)


P5/Source/Specs/att.predicate.xml
```diff
@@ -0,0 +1,42 @@
...
+            concepts of "PERSON" and "PLACE" depending on the XPATH expression given in <att>predicate</att></p>
+        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#NONE">
+            <elementSpec ident="name" mode="change">
+                <equiv name="PERSON" predicate="@type='person'"
+                    uri="http://www.example.com/entities/person"/>
+                <equiv name="PLACE" predicate="@type='place'"
+                    uri="http://www.example.com/entities/place"/>
+            </elementSpec>
+        </egXML>
+    </exemplum>
```
sydb, line 45: Probably a good idea to provide a `<listRef>` here. Off the cuff, I think it should be
~~~xml
     <listRef>
       <ptr target="#TDcrystalsCE"/>
       <ptr target="#TDPMPM"/>
     </listRef>
~~~

P5/Source/Specs/att.predicate.xml
```diff
@@ -0,0 +1,42 @@
...
+            <model predicate="parent::person"
+                behaviour="inline">
+                <desc versionDate="2015-08-21"
+                    xml:lang="en">If it is a child of a person element, treat as
+                    inline</desc>
+            </model>
+        </egXML>
+    </exemplum>
+    <exemplum xml:lang="en">
+        <p>The following example declares that the <gi>name</gi> element can be mapped to (is equivalent to) the external
```
sydb: I think I would mildly prefer 
~~~xml
<p>The following example declares that the <gi>name</gi> element can be mapped to, or is equivalent to, the external
~~~

P5/Source/Specs/att.predicate.xml
```diff
@@ -0,0 +1,42 @@
...
+                behaviour="inline">
+                <desc versionDate="2015-08-21"
+                    xml:lang="en">If it is a child of a person element, treat as
+                    inline</desc>
+            </model>
+        </egXML>
+    </exemplum>
+    <exemplum xml:lang="en">
+        <p>The following example declares that the <gi>name</gi> element can be mapped to (is equivalent to) the external 
+            concepts of "PERSON" and "PLACE" depending on the XPATH expression given in <att>predicate</att></p>
```
sydb: I’m not sure, but I think "PERSON" and "PLACE" should perhaps be `<mentioned>PERSON</mentioned>` and `<mentioned>PLACE</mentioned>`. (Or maybe `<val>` or `<soCalled>` or `<q>`.)
But I am sure that “XPATH” should be “XPath”.

P5/Source/Specs/att.predicate.xml
```diff
@@ -0,0 +1,42 @@
...
+        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#NONE">
+            <elementSpec ident="name" mode="change">
+                <equiv name="PERSON" predicate="@type='person'"
```
sydb: `@type eq 'person'` ?

P5/Source/Specs/att.predicate.xml
```diff
@@ -0,0 +1,42 @@
...
+                    uri="http://www.example.com/entities/person"/>
+                <equiv name="PLACE" predicate="@type='place'"
```
sydb: `@type eq 'place'` ?

sydb (changes requested):
I think somewhere it should be explicitly mentioned that the value of `@predicate` is an XPath expression _without the square brackets_.

P5/Source/Specs/att.predicate.xml
tuurma: doesn't really matter here, no?

sydb: In the sense that both `=` and `eq` will always yield the same results, no, it does not matter. In the sense that `@type` of `<name>` and the literal `'person`' are both, by definition, singular, it is suboptimal to be using a sequence operator.

P5/Source/Specs/att.predicate.xml
peterstadler: addressed in 116f2b4c6

P5/Source/Specs/att.predicate.xml
peterstadler: addressed in 0a5a314ed

P5/Source/Specs/att.predicate.xml
peterstadler: addressed in 0a5a314

P5/Source/Specs/att.predicate.xml
peterstadler: addressed in 4c0fca7b7

P5/Source/Specs/att.predicate.xml
peterstadler: addressed in 4c0fca7b7

peterstadler:
addressed @sydb 's concerns, please have a look (again)

cc'ing @jamescummings

tuurma (approved):

sydb (approved):
Do it, Chewie!

PR #2118, opened 2021-03-06
Oxford commas and other tiny tweaks in FT

sydb:
* mostly adding Oxford commas
 * also changed “(tm)” to “™”
 * minor whitespace changes, some inadvertent, some on purpose.

These are very minor changes. As far as I am concerned, once any reviewer who understands the rules of English punctuation better than I do says it is OK, this can be merged. (I.e., I think either @ebeshero or @JanelleJenstad could act as both reviewer and merger.)

### DIFF
--- a/P5/Source/Guidelines/en/FT-TablesFormulaeGraphics.xml
+++ b/P5/Source/Guidelines/en/FT-TablesFormulaeGraphics.xml
@@ -8,7 +8,7 @@ $Id$
 -->
 <?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
 <div xmlns="http://www.tei-c.org/ns/1.0" type="div1" xml:id="FT" n="22">
-  <head>Tables, Formulæ, Graphics and Notated Music</head>
+  <head>Tables, Formulæ, Graphics, and Notated Music</head>
   <p>Many documents, both historical and contemporary, include not only text,
     but also graphics, artwork, and other images. Although some types of images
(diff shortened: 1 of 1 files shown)


P5/Source/Guidelines/en/FT-TablesFormulaeGraphics.xml
```diff
@@ -559,14 +561,14 @@ discussion at .</note-->
...
+      enters the text flow in a way similar to figures, images, or graphs. On
       other occasions, elements of music notation are treated as inline
       characters in running text.</p>
     <p><gi>notatedMusic</gi> provides a way to signal the presence of music
       notation in text, but defer to other representations, which are not
       covered by the TEI guidelines, to describe the music notation itself. In
-      fact, several commercial, academic and standard bodies have developed
-      digital representations of music notation, and given the topic's
+      fact several commercial, academic, and standard bodies have developed
+      digital representations of music notation; and given the topic's
```
ebeshero: @sydb I dissent with this use of the semicolon, and I think you should put the comma back. When you use "and" to link two parts of a compound sentence, the conjunction is preceded by a comma. You don't need a semicolon to disambiguate this "and" from the one at the end of the list earlier. If you were to change anything here, it might just be to split this into two sentences.

ebeshero (changes requested):
@sydb I'm *mostly* okay with this, but request that you make one change here, where you introduced a semicolon. I think that should either be a comma or that you should split the compound sentence into two sentences. (I've highlighted it in the PR.)

P5/Source/Guidelines/en/FT-TablesFormulaeGraphics.xml
sydb: Split sentence in two (in commit c22b97f6d, I think).

JanelleJenstad (approved):
Checked each change. All good. Approve merging.

Issue #2128, opened 2021-03-26
Bug in overridden vallists


JanelleJenstad:
Have a look at the element spec for tag (https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-tag.html). There should be a custom vallist, but it's not displaying. We think there's an error in the spec file. Bug occurred between 3.6 (which has the values) and 4.0.0 (which doesn't have the values). vallist should have mode=add

JanelleJenstad:
There are 21 element specs with the same problem.

martindholmes:
Other possible victim elements from a quick XPath:
castItem, constitution, derivation, desc, dimensions, domain, factuality, fw, gram, graph, interaction, iType, move, node, oRef, preparedness, purpose, tag, tech, usg, xr

ebeshero:
@JanelleJenstad @martindholmes @sydb This seems weirdly relevant to our discussion of https://github.com/TEIC/Stylesheets/issues/272 yesterday, over behavior of mode attributes related to vallists.

martindholmes:
@ebeshero Yes, I think @sydb may have assumed that mode=add was not needed, but I think it is. Whether it should be or not is a different question. I think this might warrant an interim release, assuming it affects schemas as well as the web guidelines.

sydb:
Yes, schemas affected. Yes, I think an interim release would be appropriate. I will try to investigate further later this weekend.

martindholmes:
This does raise an interesting question about the `@mode` issue, though. We have two competing assumptions:

 - `@mode` is inherited
 - if there is no `@mode`, then "add" is assumed

In this case, it appears that the first is in effect, so `@mode="change"` is assumed on the `<valList>`, and since there is no original valList, nothing happens. That's the generous assumption, though; most likely there's also a bug in the Stylesheets whereby a `<valList>` with no `@mode` just gets ignored in this context.

I'm coming around to the idea that we should insist on `@mode` on all structures at any level that are expected to cause something to happen. That should be easy to enforce, and much easier to write processing for.

martindholmes:
Working on this in branch issue-2128-valLists.

martindholmes:
PR #2129 is live -- please review and merge.

sydb:
I disagree — having what I think it is _supposed_ to be, and what I vaguely recall Sebastian implying, is that it is inherited; if there is nothing to inherit, it is `"add"`. This should be trivially easy to do in XSLT 2+:
~~~xml
[...]

PR #2129, opened 2021-03-28
Add missing @mode='add' to overridden @types with valLists per issue #2128

martindholmes:
This PR fixes issue #2128: 21 cases of valLists which needed `@mode="add"` and checked the results; no impacts on the test suite.


sydb:
I still have reservations that these should all be `"change"`, not `"add"`, but
- Whereas there is no consensus on the matter; and
- Whereas @martindholmes is strongly in favor of `"add"`; and
- Whereas it should not be difficult to switch to `"change"` in the future should we change our minds; and
- Whereas we would like to get an interim release out as early as possible …

I am hereby merging this in the hopes that the 4.2.0 (and 4.2.1) release techs can get an interim release out sooner rather than later.

JanelleJenstad:
Thank you!

PR #2135, opened 2021-04-27
allow multiple ptrs on calendar= attribute per #2028

sydb:
This branch builds locally for me and in docker for @MegJBrown, but Mr. Travis says it failed.


ebeshero (commented):
I agree with the substance of these edits...we're just waiting to see if we can get this to pass build inspection...

PR #2151, opened 2021-05-20
allow <schemaSpec> in <front> and <back>

sydb:
This change is intended to catch us up to the new prose (implemented in #2041 for #1921) by putting `<schemaSpec>` into model.frontPart now, so it can be a child of `<front>` or `<back>` (which will happen in roughly late October, anyway, but should be corrected immediately).


ebeshero (approved):

ebeshero:
Just merging this since it seems uncontroversial.

PR #2156, opened 2021-06-11
improve prose, constrain value of global source= attr

### LINKED ISSUE #1869: ODD spec elements should have their own source attribute
@martindholmes and I are somewhat uncomfortable with the idea that the `@source` attribute has **entirely different** semantics when it is on “elements describing schema components” vs all other elements. (And I even think we should tighten up the definition of global `@source`, at that.) In addition, it is probably a bad idea to allow multiple values on the `@source` of, say, `<schemaSpec>` or `<elementRef>` (at least, until such time as we know what multiples would _mean_.)

The two obvious solutions are:
1. Separate out the two attributes, so that they have different properties and definitions,  even if the same name.
2. Invent a new attribute (e.g., `@oddSource` or `@schemaSource`) to be used on the ODD elements, so that they can still have the regular global `@source`, too.

sydb:
I think what I have in the new prose in the `<remarks>` is a mild improvement, but we can probably do better. (So if you, dear reviewer, have improvements feel free to just make them on this branch yourself.)


sydb:
How do we make Travis re-run the tests? (@peterstadler ?)

ebeshero (commented):
@sydb I've read your revised remarks, and I tried working out a simpler way to communicate this part: "it identifies the source from which declarations for the components of the object being defined may be obtained." I think it's okay, but would we lose anything if we trimmed it like this? 

"it identifies the source from which declarations for the components may be obtained."

Also, I take it that this last sentence of the previous remarks is no longer relevant, since the only case where we're now allowing multiple values is bibliography references. 
 "If more than one location is specified, the default assumption is that the required          source should be obtained by combining the resources indicated."
 
 So if I understand this correctly, I think the remarks are fine--maybe just need a light trim. Now, why is the Travis build unhappy?

sydb:
Alternate wording put in there, @ebeshero. And yes, since only 1 value now, no point in asserting what multiple does, especially since the Stylesheets don’t do it.

I don’t know what Mr Travis’ problem was … built fine on my local system. Let’s see if he’s happy this time.

If so, I am hoping we can merge this in right away. If not, I will ask @peterstadler or @martindholmes to have Mr. Jenkins try.

JanelleJenstad (approved):
New wording is good.

ebeshero (approved):
Looks good to me, too!

sydb:
Error in test is from `<sch:let>` that is direct child of `<constraint>`. Plan (per today’s Council meeting) is to wrap an `<sch:rule>` in there to avoid validity error. (And perhaps open another ticket, as several of us think it should not be an error.)

ebeshero:
@sydb We just realized the build error problem is due to the positioning of the sch:let in the constraintSpec.

sydb:
@peterstadler: As the prose has already been approved, the `<constraint>` has now been fixed, and all tests are now passing, I think this can be merged toot-sweet.

That said, we may want to open another ticket to make the Schematron more efficient. (Once we are including a `@context`, may as well put the test for *Spec parentage on the `@context`, not on the `@test`. The reasons I do not want to bother doing that now are a) to minimize changes immediately before a release, and b) we are considering a ticket that would allow `<sch:let>` inside `<constraint>`, in which case the `@context` could be, and probably would be, removed.)

PR #2157, opened 2021-06-17
TEI Basic

hcayless:
OK, hear me out. I've been mildly irritated for ages that the best starting point for a project that's still figuring out how it wants to encode stuff is TEI All. There's things in there that really shouldn't be used when encoding a regular document, and I've seen people pick stuff up from, say tagdocs—to be fair without properly reading the documentation, but nevertheless—that shouldn't be available at all. But starting from a more minimal schema than All is psychologically hard because users are driven by what's available when they're playing around, figuring out what to do. So what if we had an exemplar that had _almost_ everything?

I seem to remember Martin Mueller, I think when proposing what became SimplePrint, saying that what TEI needed was the equivalent of a Toyota Corolla—something relatively small but easy to understand and likely to be popular. This is not that. It's more like a midsize SUV. Which turns out to be what most people actually want in a car...Instead of 80/20, this is more like 90/95.

BUT, my thesis is that it's easier to develop a customization by refining something and that you can start with an ODD that only throws out the things you're very unlikely to use. It only removes tagdocs, iso-fs, dictionaries, nets, and numbered divs (my personal pet hate). 

Thoughts?

### DIFF
--- a/P5/Exemplars/Makefile
+++ b/P5/Exemplars/Makefile
@@ -5,6 +5,7 @@ JING=java -jar ../Utilities/lib/jing.jar
 BIN=${PREFIX}/bin
 NORMALODDS= \
 	tei_bare.dtd \
+	tei_basic.dtd \
 	tei_corpus.dtd \
 	tei_drama.dtd \
 	tei_minimal.dtd \
--- a/P5/Exemplars/tei_basic.odd
+++ b/P5/Exemplars/tei_basic.odd
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:lang="en">
+  <teiHeader>
+    <fileDesc>
+      <titleStmt>
+        <title>TEI Basic</title>
+        <author>Hugh Cayless</author>
+      </titleStmt>
+      <publicationStmt>
+        <publisher>TEI Consortium</publisher>
+        <availability status="free">
+          <p>TEI material can be licensed differently depending on the
+          use you intend to make of it. Hence it is made available
+          under both the CC+BY and BSD-2 licences. The CC+BY licence
+          is generally appropriate for usages which treat TEI content
+          as data or documentation. The BSD-2 licence is generally
+          appropriate for usage of TEI content in a software
+          environment. For further information or clarification,
+          please contact the TEI Consortium (info@tei-c.org).</p>
+        </availability>
+      </publicationStmt>
+      <sourceDesc>
+        <p>Derived from TEI All.</p>
+      </sourceDesc>
+    </fileDesc>
+  </teiHeader>
+  <text>
+    <body>
+      <head>TEI Basic</head>
+      <p>This TEI customization describes a schema that includes
+      <emph>most</emph> of the TEI P5 modules and elements. It eliminates
+      some specialized modules and numbered divisions. TEI Basic is 
+      intended to be a better starting point for most projects than TEI
+      All. It should work for nearly all texts. You are encouraged to
+      further customize it.</p>
+      <schemaSpec ident="tei_basic" start="TEI teiCorpus" prefix="tei_" targetLang="en" docLang="en">
+        <moduleRef n="01" key="tei"/>
+        <moduleRef n="02" key="header"/>
+        <moduleRef n="03" key="core"/>
+        <moduleRef n="04" key="textstructure" except="div1 div2 div3 div4 div5 div6 div7"/>
+        <moduleRef n="05" key="gaiji"/>
+        <moduleRef n="06" key="verse"/>
+        <moduleRef n="07" key="drama"/>
+        <moduleRef n="08" key="spoken"/>
+        <moduleRef n="17" key="analysis"/>
+        <moduleRef n="10" key="msdescription"/>
+        <moduleRef n="11" key="transcr"/>
+        <moduleRef n="12" key="textcrit"/>
+        <moduleRef n="13" key="namesdates"/>
+        <moduleRef n="14" key="figures"/>
+        <moduleRef n="15" key="corpus"/>
+        <moduleRef n="16" key="linking"/>        
+        <moduleRef n="21" key="certainty"/>
+      </schemaSpec>
+    </body>
+  </text>
+</TEI>
--- a/P5/Exemplars/tei_basic.tei
+++ b/P5/Exemplars/tei_basic.tei
@@ -1,0 +1,21 @@
+<TEI xmlns="http://www.tei-c.org/ns/1.0">
+  <teiHeader>
+    <fileDesc>
+      <titleStmt>
+        <title><!-- supply a title --></title>
+      </titleStmt>
+      <publicationStmt>
+        <p><!-- supply publication information--> </p>
+      </publicationStmt>
+      <sourceDesc>
+        <p><!-- supply information about the source --></p>
+      </sourceDesc>
+    </fileDesc>
+  </teiHeader>
+  <text>
+    <body>
+      <!-- the text. possibly a <p> or two? -->
+      <p/>
+    </body>
+  </text>
+</TEI>
--- a/P5/Exemplars/tei_basic.template
+++ b/P5/Exemplars/tei_basic.template
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<TEI n="Basic"
+  xmlns="http://www.tei-c.org/ns/1.0">
+  <teiHeader>
+    <fileDesc>
+      <titleStmt>
+        <title>Title</title>
+      </titleStmt>
+      <publicationStmt>
+        <p>Publication Information</p>
+      </publicationStmt>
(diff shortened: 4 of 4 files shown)


tuurma:
I remember @lb42 saying that TEI Lite was meant to be first in line of TEI customizations aiming at different communities/use cases. Request for a "reasonable starting point" are popular and seem to only differ in deciding what is "reasonable". Simple tried to be reasonable and now e-editiones as a result of one of the community meetup discussions is trying to create yet another one, aiming at the kind of documents many of our institutional members have. Definitely doesn't include any from @hcayless list so there's a clear overlap ;-)

Its working title is *TEI Vanilla*, work in progress, see https://github.com/eeditiones/tei-vanilla/blob/master/tei_vanilla.odd and will be shipped with the next TEI Publisher release with a sort of "guarantee" or at least a strong promise that if you stick to these schema and encoding guidelines, your documents will be quite nicely handled by Publisher and ready to be plugged into Archives Online portal for cross-searching. Our next step is to adjust all Publisher examples to this schema (and possibly refine it some more) and add human readable Guidelines and some sort of best practice recommendations.

While we may bicker endlessly about what should be exactly in or out of such an "entry-level" schema, in practice projects are quite happy to adapt to proven solutions. This has been the case in SSRQ ODD and TEI Publisher based application widely and voluntarily adopted by a number of other editions. Similarly, I have seen people base their individual projects on Publisher's samples which is why I think it's quite important they are good quality and can be vouched as indeed best practice by wider communities.

ebeshero:
@hcayless I like this TEI Basic very much, because my first and widest ranging TEI project started this way (combining a writer’s letters with drama, poetry, and prose works, making versioning possible too, using some manuscript tagging). I had to read a lot of TEI chapters to work out what we needed—so something like this would have been helpful to guide my way. Given the number of ODDs that promise simplicity or reduction, I think we might not want to call this “TEI basic” but something more like “TEI-Exploratory” or TEI-Ranging”.

One thing I particularly like about this ODD exemplar is that it expands horizons, helps share a view of what’s possible to encode and what range of text structures and metadata might be combined in a project. A thing I worry about with the way we model TEI in workshops or custom publishing frameworks is oversimplification and conformity in the name of making it easy for people. A little of that goes a long way, but may not encourage complex research questions or scholarly experimentation. This kind of ODD goes a long way to teach what’s possible and guide new TEI and practitioners to make sound decisions. I’d like to maybe add some guiding comments and sample customizations of elements to this, maybe a little Schematron constraintSpec, say to sanity check date ranges or some such. Anyway I like this a lot and would love to share it in workshops.

hcayless:
@ebeshero Names are hard. I thought about "Base" as well, as in "this is a base pretty much anyone can start from". "Flex" as in "flexible" is also a possibility. I've often wondered whether we should take steps to prevent people from using TEI All directly, but the problem is that there isn't a good, not quite maximal, alternative. This is meant to be something that could serve as a default, in the way the All does now, just with some of the bits you won't need unless you're doing something very specialized, like a dictionary or an ODD (and we already have a couple of exemplars for the latter).

This PR is something I would use (hence the PR). I may be an outlier, but the way I approach most TEI-based projects is not to prejudge what they'll need, but by working with the project team to figure out what the basic shape of their desired outcomes are and only then to design an ODD for them, so by pruning rather than building up. We begin by just marking up some example texts without excluding anything and refining from there.

@tuurma, I very much admire the work you're doing with Vanilla, and I think it's a great approach. I'd have liked to be involved in it, but I'm afraid I just can't do one more thing. My problem with this approach though is that it just won't ever work for me. I couldn't make a critical edition with what's in Vanilla, for example (no `@source`! No `<wit>`! I don't think it could deal with transpositions either). My argument is that I see a need for a starting point where people who don't understand what I'm doing haven't made a lot of decisions for me before I started.

tuurma:
Sure @hcayless. My point is we rather need more customizations, not less, each likely targeting a community of practice and sustained by such community. SSRQ project did create very much a model ODD for the kind of material they have (beautifully described, with proper i18n etc) so no wonder it has been adopted. Nevertheless it wouldn't be of much help to someone doing dictionaries. My hope is after Vanilla there will be Strawberry and Salted Caramel and ...

And also, I think numbered divs should be deprecated because they are abomination :-)

npcole:
Worth doing if only to get rid of numbered divs. But perhaps they should be deprecated anyway.

My only real question -- does this strip out enough to make it worth maintaining? I know this is a difficult blance.

hcayless:
What else we could remove is certainly worth discussing, but broadly I think this is something that would require very little maintenance and wouldn't _really_ have a community behind it. It is deliberately un-opinionated (ok, it does hate numbered divs). It's a starting point.

lb42:
I hate to disagree with Hugh. It never works. But I really must say that I find the arguments for doing this unpersuasive. The case being made is *not* "here's another TEI subset well adapted to a specific community's needs" (which I could get behind without any problem, I think), but rather  "here's  a better  starting point for adapting the TEI to a specific community's needs than TEI All". I don't see how that can be true, without making a lot of rather dubious assumptions about what EVERYONE is "very unlikely to use". Do you really think no-one should ever use numbered divs? If so, propose their removal. Do you really think no-one ever uses  iso-fs or dictionaries or nets ?  That's not my perception. There are plenty of people doing editions of dictionary-like texts who would question that; there are even people doing manuscript editions who use the nets module to model stemmata. I take the point that iso-fs and tagdocs are a bit "meta" for your average electronic edition, but you could say the same about a lot of the TEI Header too. Will your customization have a cut-down  header too? Is this more than a plea for dumbing down? 

I don't do as much teaching of TEI as I used to, and probably much less than others on this list, but fwiw my experience has been that the number of learners who feel encouraged and empowered when exposed to the full riches of the TEI far exceeds those who feel daunted by them.  Isn't it also rather important to learn that not everyone sees texts in exactly the same way as your own little in-group does ?

ebeshero:
@lb42 I find example ODDs very useful as “seed-starts” for my students, trying to wrap their heads around the TEI. I like the idea of assigning ODDs for them to read, and the expansive ones like this offer lots to consider and discuss, especially when accompanied by good glosses and descs. I do agree with your last point here about expanding frames of reference beyond the narrow niche. I guess that is kind of my job, really, in getting to teach this stuff to university students. And that is why I like what @hcayless has prepared here. I could easily pile on a few more things, or better yet, invite my students to try it.

hcayless:
@lb42 My prime driver for this was to have a starting point without tagdocs. You only need tagdocs if you're doing documentation, and if you're doing documentation, you should probably start with one of the ODDs for doing customizations. Leaving out dictionaries has a similar motivation: you're either doing a dictionary or you're not. I could be argued out of that one. Similarly, iso-fs: you might use it if you're doing certain types of knowledge base, but I suspect most people would use RDF for that kind of thing nowadays. Nets are another rarely-used one with plenty of competition. I wouldn't use it for a stemma, both because stemmata are usually just figures, and because there are lots of tools that do graph analysis that use other formats. And, frankly, because in most of the texts I work with it's foolhardy to be that confident about lines of transmission—therefore best to stick with drawings.

I'd like to see numbered divs deprecated. Every time the discussion comes up, the conclusion is "we'll do that in P6". I don't like them, never use them, would never recommend using them, but I could be persuaded they should be left in.

Tagdocs is the big one for me. I have seen people be confused by it and waste time on it often. 

Maybe I'm the only one who likes to work this way? I can take my weird customization of into a corner and host it myself if that's the case :-).

ebeshero:
@hcayless Obviously you are not the only person who likes to work this way, though perhaps our posts crossed. Before you take this away, perhaps I had better merge the PR!

ebeshero:
My only quibble with this is the name, but I think the ODD itself is important to keep as a new exemplar.

lb42:
@hcayless As I said, I'd be happy to get behind an ODD which is presented as "Hugh's preferred starting point for customisation". Phrases like "I wouldn't use it for..." and " in most of the texts I work with it's foolhardy to..." all make it sound more like that. (And as regards nets -- the stemmatological uses proposed in the Guidelines are very precisely intended to represent graphics, so I don't see what your point is, unless it's the usual one bemoaning the absence of TEI-savvy drawing software.

lb42:
p.s. if you delete the tagdocs module, you will also lose several useful phrase level elements such as code, ident, tag, val  and indeed eg. which you might regret when encoding quite a few texts aside from TEI specifications. That's why for example TEI Lite includes some of them.

hcayless:
@lb42 I think this needs to be a _TEI_ customization though. People start with TEI All because we host that customization, so the point of the PR was to see if we could agree on something that was less expansive than TEI All.

It's a good point about some of those, like `<code>`. I'm open to putting at least that one back. Funnily enough, `<val>` was one of the things that triggered me to do this: people see it and think "oh, this thing is a value, I can use this" and have to be corrected later. It just wastes time.

As for stemmata in TEI, I think it made perfect sense to propose this as a method, but I also think things have moved on since then. When I need to draw a stemma, I'll do it in Dot or SVG or something like that. I don't think doing it in TEI adds anything, and it means you have to worry about how to render it too.

Issue #1735, opened 2018-01-19
occurences of <content>


sydb:
In an `<elementSpec>`, the `<content>` element is optional. Seems to me, at least in tei_customization, and maybe in P5 in general, it should be
 * required if the `@mode` is "add" or "replace"
 * optional if the `@mode` is "change"
 * not allowed if the `@mode` is "delete" (along with any other children)

Does that sound right, or am I being too hasty, here?

Furthermore, `<content>` is (deliberately) allowed 0 or more times in the content of both `<dataSpec>` and `<macroSpec>`. At first blush, this seems like a mistake to me. What does a 2nd (or 3rd or 4th or 40th) `<content>` _mean_?

raffazizzi:
I'm all in favor of tightening these rules. I imagine you're suggesting to implement this with schematron?

Re your observations about `<dataSpec>` and `<macroSpec>`, I don't know, and P5 doesn't seem to have any with more than one `<content>`. But, I see that the content model for both has:

```
<alternate minOccurs="0" maxOccurs="unbounded">
   <elementRef key="content"/>
   <elementRef key="valList"/>
</alternate>
```

So, while in `<elemnetSpec>` both `<content>` and `<valList>` are allowed, they are mutually exclusive in this case. I wonder the one above is the only way of expressing this, at the expense of allowing unnecessary multiple `<content>`s?

But: why would one need multiple `<valList>`s anyway? Actually, what's the use of `<valList>` within anything but an `<attList>` or a `<content>`?

sydb:
Yes, @raffazizzi, Schematron does seem like the way to go. (While this kind of co-occurrence constraint is easy to do in RELAX NG and _can_ be done in ODD, it is probably not something to exemplify.) I have just added
~~~~~xml
<constraintSpec scheme="schematron" ident="content_when_adding">
  <constraint>
    <sch:rule context="tei:elementSpec[@mode = ('add','replace')]">
      <sch:assert test="tei:content">When adding a new element (whether replacing an old one or not), a content model must be specified; but this ＜elementSpec＞ does not have a ＜content＞ child.</sch:assert>
    </sch:rule>
  </constraint>
</constraintSpec>
<constraintSpec scheme="schematron" ident="empty_when_deleting">
  <constraint>
    <sch:rule context="tei:elementSpec[ @mode eq 'delete']">
      <sch:report test="*">When used to delete an element from your schema, the ＜elementSpec＞ should be empty</sch:report>
    </sch:rule>
[...]

PR #2163, opened 2021-07-13
Constrain elementSpec based on Mode #1735

MegJBrown:
Implemented changes #1735

### DIFF
--- a/P5/Source/Specs/dataSpec.xml
+++ b/P5/Source/Specs/dataSpec.xml
@@ -20,11 +20,10 @@ $Id$
       <alternate minOccurs="0" maxOccurs="unbounded">
         <classRef key="model.glossLike"/>
         <classRef key="model.descLike"/>
-      </alternate>
-      <!-- why ? -->
-      <alternate minOccurs="0" maxOccurs="unbounded">
-        <elementRef key="content"/> <!-- need to constrain this -->
-        <elementRef key="valList"/> <!-- also now available inside content -->
+      </alternate> 
+      <alternate minOccurs="0" maxOccurs="1">
+        <elementRef key="content"/> 
+        <elementRef key="valList"/> 
       </alternate>
       <elementRef key="constraintSpec" minOccurs="0" maxOccurs="unbounded"/>
       <elementRef key="exemplum" minOccurs="0" maxOccurs="unbounded"/>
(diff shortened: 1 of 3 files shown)


sydb:
If _anyone_ can review this within the next, say, 18 hours …
It should be an easy review. 2 of the 3 files changed have a very trivial change (which is changing a content model snippet from `( valList | content )*` to `( valList | content )?`). The third change is just a new `<constraintSpec>`, and while the XPaths make use of sequence comparison and the `except` operator, they are not particularly gnarly. Pretty straightforward stuff.

martindholmes (commented):
As far as I can see, this correctly implements the discussion on the ticket. The only thing that slightly bothers me is the word "incongruous" in the Schematron error message; I'm not sure how helpful that is, especially to a non-native speaker. Would it be better to explain why it's not workable?

sydb:
I think it is a valid complaint — that “incongruous” may not be the best wording; but I have not come up with anything better. I came up with something worse: “does not make sense”.

Ideas welcomed.

martindholmes:
@sydb Fair enough. I guess I was looking for "...incongruous because...", but it's difficult to write a single explanation that covers all violations of the rule.

JanelleJenstad:
I don't like "incongruous" either, mainly because it doesn't tell the user what to do to fix the problem. Suggested wording:

"This elementSpec element has both a
          mode= of "delete" and child elements, which is incongruous" --> "This elementSpec element has a mode= of "delete" even though it has child elements. Change the mode= to "change" or remove the child elements."

JanelleJenstad:
"This elementSpec element has a 
          mode= of "<sch:value-of select="@mode"/>", but does not have any child elements or schema-changing
          attributes, which is incongruous" --> "This elementSpec element has a 
          mode= of "<sch:value-of select="@mode"/>", but does not have any child elements or schema-changing
          attributes. Specify child elements and/or change the attributes, or change the ....."

JanelleJenstad (changes requested):
Rework wording as suggested on ticket to avoid "incongruous."

sydb:
Thank you, @JanelleJenstad. Changes made, checked in at 8c27b978c.
I was able to test Schematron on commandline, looks OK. For reasons I cannot explain, I cannot get tei_all.isosch to work in oXygen (23.1) at all, even after deleting all the children of `<schema>`. Will try to look at that a little more tomorrow.

ebeshero (approved):
This rewording looks good to me, as do the new Schematron constraints.

sydb:
Tried to reproduce the problem described as an aside [above](#issuecomment-906030233) on a different machine, also running OXygen 23.1; problem did not occur.

PR #2165, opened 2021-07-25
Ellipsis issue 1877

### LINKED ISSUE #1877: FR for new `<elision>` element
The TEI-L discussion here:

https://listserv.brown.edu/cgi-bin/wa?A2=ind1904&L=TEI-L&P=62275

culminates in a consensus that there is a need either for a new element tentatively called `<elision>` or for a clear recommendation for encoding the phenomenon whereby the primary source being encoded has quoted or included another source, but omitted part of it. Some people use `<gap>` for this despite the fact that the description of `<gap>` strongly suggests that it should only be for cases where the encoder creating the digital edition is omitting material from the source text for some reason. 

The three components of the phenomenon which would need to be encodable are:

1. The fact of the elision.
2. The signal (if any) of the elision in the original text.
3. The omitted text.

My suggested encoding would look something like this:

~~~~~
<elision resp="#editorFrom1880">
   <metamark function="elisionSignalMultiline"> *  *  *  * </metamark>
   <note resp="#editorFrom1966">Three lines omitted</note>
   <supplied resp="#teiPerson2019" source="#msFrom1860">
     <l>line one....</l>
     <l>line two....</l>
     <l>line three....</l>
   </supplied>

</elision>
~~~~~

Given the array of possibilities for the signal, a fairly rich content model might be required.

ebeshero:
This contains the specs for the new element `<ellipsis>` , together with explanation in the CO chapter and examples with new BIB entries.  Please review and merge to dev if you think it's ready, or let me know what it still needs.  @martindholmes @martinascholger @HelenaSabel @hcayless @sydb


martindholmes (commented):
Looks really good. The only thing I would change is to insert the word "signalling" in the element desc:

<desc versionDate="2021-07-07" xml:lang="en">indicates a purposeful marking in the source document signalling
    that content has been omitted and supplies the omitted content if available.</desc>

ebeshero:
Thanks @martindholmes for the revision, and @sydb, thanks for applying the missing `<sequence>` !

ebeshero:
By the way, there's a weird number of commits here (a whole series got repeated in the history): I was rebasing the branch when I updated it from dev so I could put all the commits for `<ellipsis>` at the tip, but I think my re-sequencing wound up complicating the history with the Travis checks on our repo. I think maybe I won't try that (rebasing) again *unless* I do a series of commits without pushing to the branch. (I remember we talked about squashing as another alternative.  But I tend to want to push each commit as I do my work, especially if there are a few days/weeks in between work sessions, just to back things up.)

P5/Source/Guidelines/en/BIB-Bibliography.xml
```diff
@@ -70,6 +76,10 @@ $Id$
       <bibl xml:id="NZETC01">
         <author>Baker, James K.</author>. <title>Night in Tarras</title>. In <title level="j">Hilltop: A Literary
       Paper</title>, vol 1 no 2. Wellington: Victoria University College Literary Society. (<date>1949</date>). </bibl>
+      <bibl xml:id="COEDADD-eg-93"><title>Base de datos paleográfica da lírica galego-portuguesa (PalMed)</title>.
+        Versión 1.2. <pubPlace>Santiago de Compostela</pubPlace>: <publisher>Centro Ramón Piñeiro para
+          a Investigación en Humanidades</publisher>.
```
HelenaSabel:
```suggestion
          a Investigación en Humanidades</publisher>. <ref target="http://www.cirp.gal/palmed">http://www.cirp.gal/palmed</ref>. 
```
It doesn’t seem possible to provide the URL of the `<citedRange>` (I guess that it is due to how the queries are processed) but at least we can add the URL to the database.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1729,12 +1730,47 @@ be marked with the <gi>supplied</gi> and <gi>surplus</gi> elements
...
+    <lg>
+      <l>E buscalhi cō uosco qᷓnto mal</l>
+      <l>Ela mays ᴘode aq̄sto sei eu</l>
+      <l>E ᴛodaq̄stela faz ᴘolo seu</l>
+      <l>E pºeste p<unclear>ẏ</unclear>te non por al</l>
+      <ellipsis>
+        <metamark function="signalChorus">꜒</metamark>
+        <l>Por q̄ nūca <metamark function="ellipsis">⁖—</metamark></l>
+        <supplied>
+          <l>Por q̄ nūca cō el pode poer</l>
```
HelenaSabel:
```suggestion
        <l part="I">Por q̄ nūca <metamark function="ellipsis">⁖—</metamark></l>
        <supplied>
          <l part="F">cō el pode poer</l>
```
At first I was going to suggest using two different `<supplied>` elements to solve the overlapping issue (the first line of the chorus is partially written so I think only the second part of that line should be supplied). However, I think the use of `@part` might be more elegant: What do you think @ebeshero ?

HelenaSabel (approved):
I suggested some minor changes related to the Medieval exemplum. I have a small concern about this example (and I should have thought about it before sending @ebeshero  the transcription). Although the example does not use any private use characters, I am worried about the font support of [U+1DD1](https://www.fileformat.info/info/unicode/char/1dd1/index.htm) and [U+1DD3](https://www.fileformat.info/info/unicode/char/1dd3/index.htm). If anybody thinks they could be problematic, I can quickly replace all the abbreviation marks with the corresponding `<ex>` element.

P5/Source/Guidelines/en/CO-CoreElements.xml
ebeshero: @HelenaSabel Yes, that makes sense to me! Thanks for the revision.

ebeshero:
@HelenaSabel I checked in your revisions. I did go looking for those Unicode characters for the example—and I think we’re okay but I only looked at the HTML build. Maybe we should take a look at the PDF build to make sure it turns out okay? I’ll try it later, but someone else may get to it first.

martinascholger:
Thanks @ebeshero! This looks good. 
I have two questions:
1. I wonder if the gloss that you provided for `ellipsis` is more a very short description rather than a gloss, and if this should be "ellipsis".
2. With the addition of `<classRef key="model.divPart"/>` to the content model, `schemaSpec`, `eTree`, `graph` etc. is allowed within `ellipsis` and I was wondering whether this is intended.

ebeshero:
@martinascholger Thanks. Responses:
1. I think we are concerned about how our community will understand “ellipsis” and we should not take that as a given. I think it’s important we are very precise in this particular gloss. This element is not for everything we might associate with the word “ellipsis”. 
2. No, that isn’t intended—or I can’t immediately think of an example that would include something like `<tree>` and certainly not `<schemaSpec>`, but we want to ensure this is broadly applicable to many kinds of documents so I tried to identify a model that would be maximally inclusive of anything that can be inside a `<div>` and rely on coders’ good judgment for the application.

ebeshero:
@martinascholger That said, what should we do? I keep thinking that `<ellipsis>` might conceivably want `<div>` elements to introduce higher levels of deliberately signaled omission. And, certainly, excluding things in `tagdocs` makes sense. I don’t want to complicate the content model too much, but maybe this is both _too_ flexible and _not flexible enough_. I don’t know if maybe we want a new model to include the kinds of text structures that seem plausible here? Or just tweak this one to exclude the stuff that really doesn’t belong?

ebeshero:
@martinascholger I added a commit to modify the content model of `<ellipsis>` basically to make sure it excluded things that don't make sense, and to include elements that seemed plausible. (This seems a bit like creating a new model class, but I wasn't sure I really wanted to go that far just for this new element.)

@HelenaSabel I fear you may be correct about the font support for those special characters! I just built the PDF version of the Guidelines and several characters (including the one for the ellipsis marking) came out empty or wrong. Mostly they're okay in the HTML, but there the [U+1DD3](https://www.fileformat.info/info/unicode/char/1dd3/index.htm) was showing correctly in Firefox, but not in Chrome.

HelenaSabel:
@ebeshero should I proceed with the replacement of the abbreviation marks with `<ex>` elements? Do you prefer to do something else?

ebeshero:
@HelenaSabel Yes, I think you should proceed since you understand these abbreviations (thank you). As a bonus, the Guidelines will have a new example of `<ex>`!

ebeshero:
In the PDF version, I also noticed that the picture of the example we’re discussing is pushed to a new page a little after the text that introduces it, so my prose doesn’t read as smoothly as I’d prefer at that point. (In the PDF we see two consecutive sentences starting with “Here is…”). I’m leaving myself a note here to smooth that out and take another look after @HelenaSabel ‘s edits to the egXML.

martinascholger:
@ebeshero I understand what you mean. The included elements seem plausible now. I would also not create a new class.

ebeshero:
I've just revised some of the language surrounding the example so it flows smoothly (I hope) in both PDF and web guidelines, since the image of our example doesn't appear in the same location in the PDF as on the web. Are we ready to pull this in to dev?

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1732,11 +1732,11 @@ distinguish editorial changes from those visible in the source text.
 </p>
 <p>The <gi>ellipsis</gi> element provides a means to encode the deliberate, marked omission of content in a source document. 
   Within <gi>ellipsis</gi>, <gi>metamark</gi> describes the denotation of omitted material in the source, and encoders may optionally provide 
-  a <gi>note</gi> describing the omission and <gi>supplied</gi> to indicate the content omitted at this point. Here is an example of such an indicated omission from a manuscript:
+  a <gi>note</gi> describing the omission and <gi>supplied</gi> to indicate the content omitted at this point. We provide an example of such an indicated omission from a manuscript.
   <figure><graphic url="Images/cod-10991.jpg" mimeType="image/jpeg"/>
```
martinascholger: I've just noticed that the source of the manuscript page is not given in the caption (resp. in the <head>).

martinascholger (changes requested):

P5/Source/Guidelines/en/CO-CoreElements.xml
ebeshero: Ah yes! I've just added the source to the `<head>`!

ebeshero: (Also revised the `<head>` so it's a little easier to read now, I think.)

martinascholger (approved):

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1729,12 +1730,47 @@ be marked with the <gi>supplied</gi> and <gi>surplus</gi> elements
...
+      <metamark function="signalChorus">꜒</metamark>
+      <l>Por q<ex>ue</ex> nu<ex>n</ex>ca co<ex>n</ex> el pode poer</l>
+      <l>Queo podesse por amigauer</l>
+    </lg>
+    <lg>
+      <l>E buscalhi co<ex>n</ex> uosco q<ex>ua</ex>nto mal</l>
+      <l>Ela mays pode aq<ex>ue</ex>sto sei eu</l>
+      <l>E todaq<ex>ue</ex>stela faz polo seu</l>
+      <l>E p<ex>or</ex>este p<unclear>ẏ</unclear>te non por al</l>
+      <ellipsis>
```
hcayless: @syd's suggestion
```xml
<metamark function="signalChorus">꜒</metamark>
<l rend="nobreak" part="I">Por q<ex>ue</ex> nu<ex>n</ex>ca</l>
<ellipsis>
  <metamark rend="nobreak" function="ellipsis">·:—</metamark>
    <supplied>
      <l style="nobreak" part="F">co<ex>n</ex> el pode poer</l>
      <l style="break">Queo podesse por amigauer</l>
    </supplied>
</ellipsis>
```
may be better.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1626,13 +1626,14 @@ editorial practices:
...
   <specDesc key="del"/>
+  <specDesc key="ellipsis"/>
 </specList></p>
 <p>Encoders may choose to omit parts of the copy text for reasons
 ranging from illegibility of the source or impossibility of transcribing
 it, to editorial policy, e.g. a systematic exclusion of poetry or prose
 from an encoding. The full details of the policy decisions concerned
 should be documented in the TEI header (see section <ptr target="#HD5"/>).
-Each place in the text at which omission has taken place should be
+When text is illegible or impossible to decipher, this should be
```
hcayless: I don't like this rewrite, which changes the meaning significantly. It would be good to discuss here or later the relationship between `<gap>` and `<ellipsis>`. My understanding is that gap is for unmarked omissions of text or for illegible text while ellipsis is for marked, omitted text.

hcayless: Put another way, `<gap>` *must* be a legal way to mark omitted text (as well as illegible text).

hcayless (changes requested):
The meeting over the content model needs to happen--I currently fall on the side of those who think it should start out more restrictive. I think the wording around gap needs correcting, and perhaps a sentence or so about the relationship between gap, supplied, and ellipsis would be very useful. I worry that people may get confused about when to use one or the other.

Does ellipsis need `@reason` like gap and supplied have? Each has their own definition of it...

P5/Source/Guidelines/en/CO-CoreElements.xml
ebeshero: From today's conversation, `<metamark>` is not semantically part of a `<l>`, so it doesn't need to be inside it. And it is preferable for `<metamark function="ellipsis">` to be the first child of `<ellipsis>` to remove ambiguity, especially in processing.

P5/Source/Guidelines/en/CO-CoreElements.xml
ebeshero: @hcayless To discuss `<gap>` and `<ellipsis>` together, we should probably rearrange the section so that the detailed explanation and example of `<ellipsis>` follows immediately after `<gap>`. 

I'm going to try rewriting the paragraph here. What do you think of this? New phrasing for `<gap>` is marked in bold: 

Encoders may choose to omit parts of the copy text for reasons ranging from illegibility of the source or impossibility of transcribing it, to editorial policy, e.g. a systematic exclusion of poetry or prose from an encoding. The full details of the policy decisions concerned should be documented in the TEI header (see section 2.3 The Encoding Description). **The `<gi>gap</gi>` element describes a passage that is unreadable, or an omission that is not marked in the source document. Attributes on the `<gi>gap</gi>` optionally provide** further information about the reason for the omission, its extent, and the person or agency responsible for it, as in the following examples:

[. . . examples of `<gap>` follow . . .]

**Quite different from `<gi>gap</gi>`, the `<gi>ellipsis</gi>` element is used when a source document deliberately signals with symbols or markings that content is omitted.** Within `<gi>ellipsis</gi>`, `<gi>metamark</gi>` describes the denotation of omitted material in the source, and encoders may optionally provide a `<gi>note</gi>` describing the omission and `<gi>supplied</gi>` to indicate the content omitted at this point. 

[. . . example(s) of `<ellipsis>` follow . . .]

ebeshero: I've been tinkering with this rewriting, so please read on GitHub.

ebeshero:
- By modifying the example as we discussed in today's Council meeting, it's easy to make the content model more restrictive (as we began). 
- I've tried rewriting the paragraph here in the PR, to clarify exactly what makes `<ellipsis>` different from `<gap>`. In the process, I found myself wanting to provide the `@reason` attribute to `<ellipsis>` just as @hcayless suggests.  If we provide some suggested values, this may also help to clarify how very different `<ellipsis>` is from `<gap>`. 
- I'm going to try adding all this to the PR before we meet to reflect what we figured out today.

hcayless:
I'm wondering if ellipsis ought to be able to contain gap as well as supplied. How would you do an ellipsis with a partially supplied text?

ebeshero:
So far our examples are either _all_ or _nothing_: we either know exactly what is omitted and can supply it (as a verse chorus), or we have no notion of what to supply (as in the example in the spec where we provide a `<note>` to describe what the printer seems to be doing). Is it appropriate to use`<gap>` if we don't know what to supply in an `<ellipsis>`?
Two thoughts:

-  Perhaps`<gap>` inside `<ellipsis>` would be redundant, since `<ellipsis>` already indicates that something is missing.  A `<note>` could be sufficient. But...
-   If we can't supply what belongs in an `<ellipsis>`, does `<gap>` help us? Well, maybe, because it would provide a way to indicate a `@reason` for why we don't know what's supposed to be here, as well as a `@unit` and `@quantity` of something missing.  

I suppose we could re-encode @martindholmes 's example (in the spec) to use `<gap>` to indicate how much is supposed to be printed at the point of the deliberate omission markings.  But I wonder if that is *really* a `<gap>`.

What if, instead, we simply put the `@reason`, `@quantity`, `@unit` (and the other attributes we provide to describe `<gap>`) directly on `<ellipsis>`?  When we cannot supply the missing content, we could at least use the same attributes we would on `<gap>` to describe what's missing.

ebeshero:
If we only have *partially supplied* text I could see a reason to permit `<gap>`, but it would surely go inside `<supplied>`. And `<supplied>` already contains `<gap>`.

hcayless:
@ebeshero I'd argue that `<gap>` inside `<supplied>` (absent a declaration otherwise in the Guidelines) means something quite specific: a conjectured omission, i.e. "I think the original text was missing something." So you couldn't signal "I know some of what this ellipsis leaves out" that way. Anyway, it's not a major objection, but something to consider because `<gap>` and `<supplied>` often work together (side by side) in this way.

ebeshero:
@hcayless About `<gap>` inside `<supplied>`: That's fair enough for describing a typical usage of these two together. But `<gap>` does describe something omitted because the editor does not know what belongs here, and it takes a `@reason` attribute, which could have special values for the case of `//ellipsis/supplied/gap`. I think we might well expect `<gap>` to be used in this way, if we're not going to indicate that a `<supplied>` element is only partial with just a `<note>`.

ebeshero:
Looking at this the other way around: We wouldn't want to *remove* `<gap>` from `<supplied>` within the context of `<ellipsis>`, would we?

ebeshero:
We could explain something more in the Guidelines about what `//ellipsis/supplied/gap` would indicate, and that might be aided by an example...

hcayless:
I would never want to remove it, but I think what it would be saying is "This ellipsis mark stands in place of a piece of text that itself contains an omission." I would rather you ignore my suggestion entirely than try to come up with new contextual rules about this stuff. Absent an actual example, I think you'd be safe leaving `<gap>` out of the content model. I'm just pointing out that `<gap>` and `<supplied>` often work in tandem, so we might have to revise it later.

ebeshero:
Well, the only way we have `<gap>` now is within the content model of `<supplied>`. It's not listed in the content model for `<ellipsis>` (as of now). So, perhaps the less we specify on this gappy subject, the better (for now)...? (I suspect most of the time people would handle a partially understood ellipsis with a `<note>` anyway.)

hcayless:
My suggestion was precisely as fodder for the content model discussion :-).

lb42:
I haven't looked at this in detail, so apologies if this is a stupid question, but I wonder whether usages like "D----n your  eyes, Mistress P----" or "the Countess of  B____ shire" would be appropriately tagged with this new ellipsis element. If so, I'd suggest some examples doing so, since this is a very frequent phenomenon in  early print. If not, some explanation why not would be useful, since the definition so far does not seem to preclude it.

PFSchaffner:
Coming to this as an encoder, I would be looking to see it it was the right choice for any of the various categories of textual situation in which the reader is expected to supply information in response to a cue in the original. Among these would be (1) the polite ellipsis (such as Lou mentions); (2) the 'ditto' ellipsis (either literal ditto marks or something similar, like the underscore typically used in a bibliography to indicate a repeated author -- any time one is expected to supply boilerplate from context); (3) the 'fillable form' ellipsis, in which blanks are left to be filled differently by different readers ("I ____ have examined the good ship ____ on this ___ day of ____, 20__, and find it fit to sail."); and the (4) 'deferred information' ellipsis, such as is found frequently in name lists, where the author can't remember the full name, so leaves a blank to be filled in later; or an initial to be illuminated later, etc. A marginal case would be rebuses and cryptograms and 'concrete poems,' etc., which are meant to be read as text but do not appear as such.

martindholmes:
@lb42 There's a similar thing in Japanese where instead of the name of a person or place, a romaji character representing its supposed initial is used:

W県刑務所

The W_____ Prefecture penitentiary

In that case, the use of "W" is a sort of metamark for omission or ellipsis; there are no underscore or dash characters representing the absence.

ebeshero:
Hi @hcayless @sydb @tuurma and all reviewing this: I've constrained the content model of `<ellipsis>` following our discussion in the TEI Council meeting, and also (following discussion here) gave it some characteristics of `<gap>` (added optional `model.descLike` and `model.certLike` and gave it membership in `att.dimensions`). I have also added some super simple examples to balance our interesting complicated one (and used @sydb 's encoding for that one).

sydb:
Oh, thank you @lb42. I read that and was doubting my understanding of the word “denote”. I am sure, however, that I do not understand what @ebeshero means by “Saussure”. I presume a reference to the linguist, but still don’t get the meaning.

Anyway, I like @lb42’s suggested wording better, too. However, I do not like the word “describes”, as the `<metamark>` does not really describe a mark, but rather encodes and thus represents it; also I would prefer either articles in front of both element names, or neither. I.e., something like either
~~~xml
Within <gi>ellipsis</gi>, <gi>metamark</gi> is used
to encode how the presence of omitted material is
indicated in the source,
~~~
or
~~~xml
Within an <gi>ellipsis</gi>, a required <gi>metamark</gi> is used
to represent the characters or markings that were used in the
source to indicate the presence of omitted material,
~~~
Also might be reasonable to divide this sentence into 2.

ebeshero:
@sydb Ha! I was revising just as you were suggesting revisions. I like your second one better (and yeah, I'd already divided into two sentences...)

As for Saussure, well its really a whole field of semiotics he introduced. Here's a fun crash course: https://literariness.org/2016/03/18/connotation-and-denotation/ (I used to get to lecture on this stuff in my old life as an English prof.)

sydb:
I have a bone to pick with the definition of `<ellipsis>`:
> indicates a purposeful marking in the source document signalling that content has been omitted, and supplies the omitted content if available

This implies that, if the omitted content _is_ available, an encoder is compelled to supply it. I think encoders should feel free to supply the omitted text or not. Thus: “indicates a purposeful marking in the source document signalling that content has been omitted, and may also supply the omitted content” or some such.

ebeshero:
@sydb Agreed! I'm tweaking that language now to make clearer what's required (the metamark) and what's optional. 

As for the Saussure crash course, I foolishly thought it contained pictures helping to describe the semiotic concepts. I might have been accidentally correct: what I was seeing were ads, which are chock full of denotation/connotation examples. Now and again semiotics creeps into my writing about metamarks--gotta be careful. But I *do* know a particularly spicy example of Lord Byron's use of a particular metamark to convey something obscene that he did not intend to put into words. . .

lb42:
Sometimes of course the whole point of an authorial ellipsis is that there is no way of supplying what has been ostensibly omitted.

ebeshero:
@sydb How about this definition in the Spec for `<ellipsis>`?
"indicates a purposeful marking in the source document signalling that content has been omitted, and may also supply or describe the omitted content."  (I just added the capacity to describe there, since it's built into the content model and the example on the spec).

sydb:
Warning: Am pushing change to ellipsis content model now …

ebeshero:
@sydb Thanks! Pulling that in on my end. I think I've done that before--not tucked all the content into `<sequence>`. Wonder if we should be testing for that...

sydb:
I was going to say “We _do_ test for that, how do you think I found it?”, but in fact the part of the build that fails is `make exemplars`, not `make test`.

So there is an argument to be made for adding a test that would catch this a bit earlier. But there is also, of course, the perennial argument to stop supporting DTDs.

tuurma (approved):

tuurma:
@ebeshero @sydb How about this example for ellipsis with supplied using copyOf attribute to point to relevant content?

```xml
 <lg>
            <l>You think you've lost your love </l>
            <l>Well, I saw her yesterday </l>
            <l>It's you she's thinking of </l>
            <l>And she told me what to say</l>
         </lg>
         <lg xml:id="chorus">
            <label>[Refrain]</label>
            <l>She says she loves you </l>
            <l>And you know that can't be bad </l>
            <l>Yes, she loves you </l>
            <l>And you know you should be glad</l>
         </lg>
         <lg>
            <l>She said you hurt her so </l>
            <l>She almost lost her mind </l>
            <l>But now she said she knows </l>
            <l>You're not the hurting kind</l>
         </lg>
         <ellipsis>
            <metamark>[Refrain]</metamark>
            <supplied copyOf="#chorus"/>
         </ellipsis>
```

ebeshero:
@tuurma Thank you for the good and speedy example! I'll work it in.

ebeshero:
@sydb @hcayless @MegJBrown Okay, I've pushed updates to simplify the content model as we discussed, and included @tuurma 's new example. We didn't need to change much since just one example was affected by the change, and a small bit of prose in CO needed to be updated. Hope it's ready now.

tuurma (approved):

sydb (approved):

P5/Source/Specs/ellipsis.xml
```diff
@@ -0,0 +1,95 @@
...
+        <l>But a short time ago, to a quack you would go,</l>
+        <l>To steam a fat man to a thinner;</l>
+        <l>Now changed from all that, if you wish to get <emph>fat</emph>,</l>
+        <l>Come to Barton’s and eat a <emph>steam dinner!</emph></l>
+        <l>Oh dear! think of a scheme, odd though it seem—</l>
+        <l>I’m sure ’twill succeed if you make it by steam.</l>
+      </lg>
+      <lg>
+        <l>You may sleep, you may dream, you may travel by steam,</l>
+        <l>For the outrcry is still to go faster;</l>
```
hcayless: Typo? "outcry"?

ebeshero: Yes—good catch.

hcayless (approved):

PR #2168, opened 2021-07-30
Attlocatable issue 1910

### LINKED ISSUE #1910: Add location information to the new `<conversion>` element
I am thrilled by the new `<unitDecl>` and its children introduced with version 3.6.0 of the guidelines (the discussion around this element can be found [here](https://github.com/TEIC/TEI/issues/1707)). I am working extensively with mark-up of food prices in sources from the late Ottoman mediterranean in various languages (Arabic, Ottoman, French, English, and German). What strikes me as missing from this wonderful new system of declaring non-SI units, is the possibility to add location information to `<conversion>`s. Unit names for volumes, weights, etc. were largely the same across the geographic space of interest to me, but the actual measures differed substantially not just over time (accounted for by `<conversion>` being a member of `att.datable.w3c`) but also between cities and regions. Thus, a *shunbul* (a volume used to measure grain) equalled 2.25 *madd* in Acre, 3 *madd* in Aleppo and 72 (sic) *madd* in Damascus.

Currently, I can only specify the geographic region for a unit at the level of the `<unitDecl>`, which would mean, I would have to declare three different units for my small example above, whereas with, for instance, the option to allow `<placeName>` as child of `<conversion>` one could just add additional `<conversion>` elements to an existing  `<unitDecl>`. In a sense, I ultimately image a system similar to the advantages of `att.datable.w3c`, which is already present.

This would remove a lot of overhead to the mark-up and the mark-up process. I have encoded every one of my 20k sources (mostly letters and periodical articles) as an individual TEI file that allows for easy deduction of the geographic region/place a source. With the proposed `<placeName>` child, in addition to `@from` and `@to` attributes, one could then easily look up the relevant `conversion/@formula` for normalising and comparing measures.

ebeshero:
With this PR I've created an attribute class, `att.locatable` for uses of `@where` that point to canonical `<place>` entries in a placeography. It resolves #1910 (asking for `@where` on `<conversion>`), and applies the new `att.locatable` to `@where` on `<event>` as well.

### DIFF
--- a/P5/Source/Guidelines/en/BIB-Bibliography.xml
+++ b/P5/Source/Guidelines/en/BIB-Bibliography.xml
@@ -424,7 +424,7 @@ $Id$
                 <author>Grallert, Till</author>, <title>Tools for the Computational Normalisation and Analysis of Food Prices 
                   and Food Riots in the Eastern Mediterranean (Bilād al-Shām) in the 19th and 20th Centuries</title>.
                 <publisher>Zenodo</publisher>, <date>2021</date>. 
-                <ref target="https://doi.org/10.5281/zenodo.5159020">https://doi.org/10.5281/zenodo.5159020</ref>.
+                <ptr target="https://doi.org/10.5281/zenodo.5159020"/>.
               </bibl>
               <bibl xml:id="PH-eg-07">
                 <author>Graves, Robert</author>. <title>Rough draft of letter to Desmond Flower</title>. <date>17 Dec
@@ -666,7 +666,7 @@ Papers of the Presidents: 1789-1897</title>, Washington,
                     McCarty</editor>. <pubPlace>London</pubPlace>: <publisher>Ashgate</publisher> (<date>2012</date>).</bibl>
                     <bibl xml:id="NDGEOGste-Falls"><author>McCarthy, Niall</author>, <title level="a">No Parachute: The Highest Falls People Survived</title>,
                     <title level="m">Statista</title>, <date when="2020-11-18">November 18, 2020</date>.
-                    <ref target="https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/">https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/</ref>
+                    <ptr target="https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/"/>.
                     </bibl>
                     <bibl xml:id="DSHD-eg-30"><author>Melville, Herman</author>, <title level="m">Moby Dick</title>.
                     (<date>1851</date>).</bibl>
--- a/P5/Source/Guidelines/en/BIB-Bibliography.xml
+++ b/P5/Source/Guidelines/en/BIB-Bibliography.xml
@@ -420,6 +420,12 @@ $Id$
               <bibl xml:id="G2S">
                 <author>[Goldsmith, Oliver]</author>, <title>The History of Little GOODY TWO-SHOES; Otherwise called, Mrs.
               MARGERY TWO-SHOES...</title> (Printed for J. Newbery, <date>1766</date>). </bibl>
+              <bibl xml:id="HDUDECL-TG">
+                <author>Grallert, Till</author>, <title>Tools for the Computational Normalisation and Analysis of Food Prices 
+                  and Food Riots in the Eastern Mediterranean (Bilād al-Shām) in the 19th and 20th Centuries</title>.
+                <publisher>Zenodo</publisher>, <date>2021</date>. 
+                <ref target="https://doi.org/10.5281/zenodo.5159020">https://doi.org/10.5281/zenodo.5159020</ref>.
+              </bibl>
               <bibl xml:id="PH-eg-07">
                 <author>Graves, Robert</author>. <title>Rough draft of letter to Desmond Flower</title>. <date>17 Dec
                 1938</date>. (from <title level="m">Diary of Robert Graves 1935-39 and ancillary materials</title>, compiled
--- a/P5/Source/Guidelines/en/BIB-Bibliography.xml
+++ b/P5/Source/Guidelines/en/BIB-Bibliography.xml
@@ -664,6 +664,10 @@ Papers of the Presidents: 1789-1897</title>, Washington,
                     level="m">Collaborative Research in the Digital Humanities</title>: A volume in honour of Harold Short on the
                     occasion of his 65th birthday, September 2010, ed. <editor>Marilyn Deegan</editor> and <editor>Willard
                     McCarty</editor>. <pubPlace>London</pubPlace>: <publisher>Ashgate</publisher> (<date>2012</date>).</bibl>
+                    <bibl xml:id="NDGEOGste-Falls"><author>McCarthy, Niall</author>, <title level="a">No Parachute: The Highest Falls People Survived</title>,
+                    <title level="m">Statista</title>, <date when="2020-11-18">November 18, 2020</date>.
+                    <ref target="https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/">https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/</ref>
+                    </bibl>
                     <bibl xml:id="DSHD-eg-30"><author>Melville, Herman</author>, <title level="m">Moby Dick</title>.
                     (<date>1851</date>).</bibl>
                     <!-- ?check Melville, Moby Dick (but where?) -->
@@ -1839,7 +1843,7 @@ Papers of the Presidents: 1789-1897</title>, Washington,
                                             <surname>DeRose</surname>
                                           </editor>
                                           <title level="m">XML Path Language (XPath) Version 1.0</title>
-                                          <ptr target="http://www.w3.org/TR/xpath/"/>
+                                          <ptr target="https://www.w3.org/TR/xpath/"/>
                                           <imprint>
                                             <publisher>W3C</publisher>
                                             <date when="1999-11-16">16 November 1999</date>
@@ -1873,10 +1877,58 @@ Papers of the Presidents: 1789-1897</title>, Washington,
                                             <surname> Siméon</surname>
                                           </editor>
                                           <title level="m">XML Path Language (XPath) 2.0</title>
-                                          <ptr target="http://www.w3.org/TR/xpath20/"/>
+                                          <ptr target="https://www.w3.org/TR/xpath20/"/>
                                           <imprint>
                                             <publisher>W3C</publisher>
                                             <date when="2007-01-23">23 January 2007</date>
+                                          </imprint>
+                                        </monogr>
+                                      </biblStruct>
+                                      <biblStruct xml:id="XPATH30">
+                                        <monogr>
+                                          <editor>
+                                            <forename>Jonathan</forename>
+                                            <surname>Robie</surname>
+                                          </editor>
+                                          <editor>
+                                            <forename>Don</forename>
+                                            <surname>Chamberlin</surname>
+                                          </editor>
+                                          <editor>
+                                            <forename>Michael</forename>
+                                            <surname>Dyck</surname>
+                                          </editor>
+                                          <editor>
+                                            <forename>Jon</forename>
+                                            <surname>Snelson</surname>
+                                          </editor>
+                                          <title level="m">XML Path Language (XPath) 3.0</title>
+                                          <ptr target="https://www.w3.org/TR/xpath-30/"/>
+                                          <imprint>
+                                            <publisher>W3C</publisher>
+                                            <date when="2014-04-08">8 April 2014</date>
+                                          </imprint>
+                                        </monogr>
+                                      </biblStruct>
+                                      <biblStruct xml:id="XPATH31">
+                                        <monogr>
+                                          <editor>
+                                            <forename>Jonathan</forename>
+                                            <surname>Robie</surname>
+                                          </editor>
+                                          <editor>
+                                            <forename>Michael</forename>
+                                            <surname>Dyck</surname>
+                                          </editor>
+                                          <editor>
+                                            <forename>Josh</forename>
+                                            <surname>Spiegel</surname>
+                                          </editor>
+                                          <title level="m">XML Path Language (XPath) 3.1</title>
+                                          <ptr target="https://www.w3.org/TR/xpath-31/"/>
+                                          <imprint>
+                                            <publisher>W3C</publisher>
+                                            <date when="2017-03-21">21 March 2017</date>
                                           </imprint>
                                         </monogr>
                                       </biblStruct>
--- a/P5/Source/Guidelines/en/HD-Header.xml
+++ b/P5/Source/Guidelines/en/HD-Header.xml
@@ -1799,10 +1799,11 @@ more than one language, as in the following example:
       <specList>
         <specDesc key="unitDef"/>
         <specDesc key="unit"/>
-        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when"/> 
+        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when where"/> 
         <specDesc key="att.formula" atts="formula"/>
+        <specDesc key="att.locatable" atts="where"/>
       </specList>
-Within the <gi>unitDef</gi>, a <gi>conversion</gi> element may be used to store information relating to conversion between units. The <gi>conversion</gi> element holds a special pair of attributes, <att>fromUnit</att> and <att>toUnit</att>, which serve to indicate the direction of a calculation from one unit of measure (stored in <att>fromUnit</att>) to another (stored in <att>toUnit</att>). A mathematical calculation to define the relation between these units may be stored in <att>formula</att>, as shown in the following examples. The <att>formula</att> attribute takes a value expressed as an XPath expression, which means that division must be expressed with <q>div</q> so as not to be confused with the forward slash used in path navigation. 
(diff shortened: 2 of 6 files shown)


martinascholger (approved):
Thanks @ebeshero! This looks good to me.

peterstadler (approved):
Thanks @ebeshero

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -1799,10 +1799,11 @@ more than one language, as in the following example:
       <specList>
         <specDesc key="unitDef"/>
         <specDesc key="unit"/>
-        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when"/> 
+        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when where"/> 
         <specDesc key="att.formula" atts="formula"/>
```
sydb, line 1803: _Either_ `@formula` and `@where` should be listed on `specDesc/@key`, OR there should be `<specDesc>`s for att.formula and att.locatable, not both.
(I realize this is not specific to this PR — `@formula` was listed twice before this PR came along, but still …)

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -1799,10 +1799,11 @@ more than one language, as in the following example:
       <specList>
         <specDesc key="unitDef"/>
         <specDesc key="unit"/>
-        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when"/> 
+        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when where"/>
```
sydb: See line 1803 comment.

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -1799,10 +1799,11 @@ more than one language, as in the following example:
       <specList>
         <specDesc key="unitDef"/>
         <specDesc key="unit"/>
-        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when"/> 
+        <specDesc key="conversion" atts="formula fromUnit toUnit from to notBefore notAfter when where"/> 
         <specDesc key="att.formula" atts="formula"/>
+        <specDesc key="att.locatable" atts="where"/>
```
sydb, line 1804: See line 1803 comment.

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -1799,10 +1799,11 @@ more than one language, as in the following example:
...
+Within the <gi>unitDef</gi>, a <gi>conversion</gi> element may be used to store information relating to conversion between units. The <gi>conversion</gi> element holds a special pair of attributes, <att>fromUnit</att> and <att>toUnit</att>, which serve to indicate the direction of a calculation from one unit of measure (stored in <att>fromUnit</att>) to another (stored in <att>toUnit</att>). A mathematical calculation to define the relation between these units may be stored in <att>formula</att>, as shown in the following examples. The <att>formula</att> attribute takes a value expressed as an XPath expression, which means that division must be expressed with <q>div</q> so as not to be confused with the forward slash used in path navigation. The <gi>conversion</gi> element may also hold attributes to store dates and date ranges as well as location(s) that specify when and where a conversion formula was used.
```
sydb: 1. [Not directly related to this FR:] Why does the penultimate sentence on this line (why are there multiple sentences on one line, anyway?) advise that “div” (as opposed to ‘/’) must be used for division, but does not mention other features of XPath expressions, or at least others that are likely to be used in a formula? (E.g., that modulo is expressed “mod”, or that “fn:abs()” is used for absolutely value?) I am not convinced this is wrong, but it should at least be worded in a way that makes clear there are other restrictions, and should give the user a pointer to more info. Perhaps “`The <att>formula</att> attribute takes a value expressed as an XPath expression, for which see <ptr target="#XPATH2"/>. Among other things, this means that division must be expressed with <q>div</q> so as not to be confused with the forward slash used in path navigation.`”
2. In the ultimate sentence [thus directly related to this PR], I am a bit uncomfortable with the verb “to hold”. There is no where else in the _Guidelines_ (except this element, <conversion>) where we say an element “holds” an attribute.

P5/Source/Guidelines/en/ND-NamesDates.xml
```diff
@@ -816,6 +816,7 @@ target="#CONA"/>).  The tag -->The
           <specDesc key="state"/>
           <specDesc key="trait"/>
           <specDesc key="event" atts="where"/>
+          <specDesc key="att.locatable" atts="where"/>
```
sydb, line 819: See comment for HD line 1803.

P5/Source/Specs/att.locatable.xml
```diff
@@ -0,0 +1,25 @@
...
+    <attDef ident="where" usage="opt">
+      <desc versionDate="2021-07-29" xml:lang="en">indicates one or more locations by pointing to a <gi>place</gi> element</desc>
+      <desc versionDate="2021-07-29" xml:lang="fr">indique une ou plusieurs localisations en pointant vers un élément <gi>place</gi></desc>
+      <desc versionDate="2021-07-29" xml:lang="es">indica una o más localizaciónes señalando un elemento <gi>place</gi></desc>
+      <desc versionDate="2021-07-29" xml:lang="it">indica una o più posizioni facendo riferimento a un elemento <gi>place</gi></desc>
+      <datatype maxOccurs="unbounded"><dataRef key="teidata.pointer"/></datatype>
+    </attDef>
+  </attList>
+  <listRef>
+    <ptr target="#HDUDECL" type="div3"/>
```
sydb, line 23: I am wondering (aloud) if "#NDPERSbp" should be listed here, too. I had thought so, because that is the other place where att.locatable is used, after all. But in fact the section there only mentions it using `<specDesc>`, not in any actual prose. So maybe not.

sydb (changes requested):

P5/Source/Guidelines/en/HD-Header.xml
ebeshero: @sydb Here are responses with questions:
1. This is an opportunity to improve on the initial documentation on these comparatively new elements. Would a reference to Michael Kay (2007) (in our BIBL) satisfy the need to provide more information on XPath expressions? It is what we provide in [teidata.xpath](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-teidata.xpath.html).

 I remember that we expected lots of arithmetic operations and especially division in conversion formulas, so that's why we highlighted the special use of "div" in XPath. That said, I like this rewriting you've suggested. Maybe we work Michael Kay into it thus (with link to Kay in the BIBL): 
"The formula attribute takes a value expressed as an XPath expression, for which see Kay (ed.) (2007) . Among other things, this means that division must be expressed with div so as not to be confused with the forward slash used in path navigation."

2. Surveying the Guidelines and Specs for how we describe elements in relation to attributes, apparently it's more typical for elements to "use" attributes to say something specific. Or they "supply" information. So, how about this revision to that last sentence?
 
"Attributes on the <gi>conversion</gi> element may also supply dates and date ranges as well as location(s) that specify when and where a conversion formula was used."

ebeshero: (updated this a little for clarity): read on GitHub.

P5/Source/Specs/att.locatable.xml
ebeshero: @sydb I also was wondering if we should add an example of `<event where="#someplace">` in the ND chapter, even just applying it to one of the many examples. I think an easy place to exemplify `@where` on `<event>` is in #NDGEOGste, with  discussion of `<listEvent>`. (And then, of course, provide a reference in the element spec for event.) Shall I just do that?

P5/Source/Guidelines/en/HD-Header.xml
sydb: 1. Yes, could certainly use the XSLT spec, which is what “Kay” refers to from teidata.xpath. As you can now see (since I have edited my post) I had suggested the XPath spec itself rather than the XSLT spec, which I think makes more sense here (and maybe in teidata.xpath, too). But I think a further tweak to the wording is needed, thus: “`The <att>formula</att> attribute takes a value expressed as an XPath expression, for which see <ptr target="#XPATH2"/>. Among other things, this means that division must be expressed with <q>div</q> as opposed to the more common forward slash (<q>/</q>, U+002F), so as not to be confused with path navigation.`”.

P5/Source/Specs/att.locatable.xml
sydb: There is only one example in there with `<listEvent>`. If you know or can find out where the events “Treaty of Utrecht”, “Quebec Act”, and “Treaty of Fort Pitt” took place, I think it is an excellent idea.

If not, perhaps we should make up a new combined `<listEvent>`, `<listPerson>`, and `<listPlace>` example based on Harry Potter or Nelson Mandella or The Holocaust or the discoveries of blood (circulation by Harvey in 1628, groups by Landsteiner in 1900, types by Janský in 1907, Rh factor (although that is not a simple discovery with a date & place, so maybe not)).

sydb: OK, see the attached file. (You have to change the extension, gitHub does not like .tei nor .xml.)
[survivors_from_fall_from_airplane.txt](https://github.com/TEIC/TEI/files/6988779/survivors_from_fall_from_airplane.txt)

P5/Source/Guidelines/en/HD-Header.xml
ebeshero: check. (checking these off as I go)

P5/Source/Guidelines/en/HD-Header.xml
ebeshero: check

P5/Source/Guidelines/en/HD-Header.xml
ebeshero: Check.

P5/Source/Guidelines/en/ND-NamesDates.xml
ebeshero: check

P5/Source/Specs/att.locatable.xml
ebeshero: Added the example (after condensing it a wee bit). Thank you, @sydb !

ebeshero:
@sydb Take a look through my revisions, and rendering of your example! Is this ready to pull in for the release?

P5/Source/Specs/att.locatable.xml
```diff
@@ -0,0 +1,26 @@
...
+$Date$
+$Id$
+-->
+<?xml-model href="http://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei"
+           type="atts" ident="att.locatable" predeclare="true">
+  <desc versionDate="2021-07-29" xml:lang="en">provides attributes for referencing locations by pointing to entries in a canonical list of places.</desc>
+  <attList>
+    <attDef ident="where" usage="opt">
+      <desc versionDate="2021-07-29" xml:lang="en">indicates one or more locations by pointing to a <gi>place</gi> element</desc>
```
sydb: One does not _have_ to point to a `<place>` element, eh? Perfectly legal to point instead to a Wikipedia page or GeoNames.org or the Getty TGN or some such, no?
I also don’t like the disagreement in number, but am having a hard time word-smithing around it. 
Maybe
~~~xml
<desc versionDate="2021-08-22" xml:lang="en">indicates a
location by pointing to a <gi>place</gi> element or other
canonical description</desc>
<!-- ... -->
<remarks>
  <p>This attribute may be used to point to multiple canonical
  descriptions simultaneously. These Guidelines make no
  recommendation as to whether multiple pointers should be
  used for different canonical references to the same place, or
  references to different places.</p>
</remarks>
~~~
or some such?

P5/Source/Guidelines/en/BIB-Bibliography.xml
```diff
@@ -420,6 +420,12 @@ $Id$
               <bibl xml:id="G2S">
                 <author>[Goldsmith, Oliver]</author>, <title>The History of Little GOODY TWO-SHOES; Otherwise called, Mrs.
               MARGERY TWO-SHOES...</title> (Printed for J. Newbery, <date>1766</date>). </bibl>
+              <bibl xml:id="HDUDECL-TG">
+                <author>Grallert, Till</author>, <title>Tools for the Computational Normalisation and Analysis of Food Prices 
+                  and Food Riots in the Eastern Mediterranean (Bilād al-Shām) in the 19th and 20th Centuries</title>.
+                <publisher>Zenodo</publisher>, <date>2021</date>. 
+                <ref target="https://doi.org/10.5281/zenodo.5159020">https://doi.org/10.5281/zenodo.5159020</ref>.
```
sydb: Probably should be `<ptr>`, not `<ref>`. 
There are only 7 other `<ref>` in BIB.xml whose content matches their `@target`, but 18 that do not. (And 137 `<ptr>`s.)

P5/Source/Guidelines/en/BIB-Bibliography.xml
```diff
@@ -658,6 +664,10 @@ Papers of the Presidents: 1789-1897</title>, Washington,
                     level="m">Collaborative Research in the Digital Humanities</title>: A volume in honour of Harold Short on the
                     occasion of his 65th birthday, September 2010, ed. <editor>Marilyn Deegan</editor> and <editor>Willard
                     McCarty</editor>. <pubPlace>London</pubPlace>: <publisher>Ashgate</publisher> (<date>2012</date>).</bibl>
+                    <bibl xml:id="NDGEOGste-Falls"><author>McCarthy, Niall</author>, <title level="a">No Parachute: The Highest Falls People Survived</title>,
+                    <title level="m">Statista</title>, <date when="2020-11-18">November 18, 2020</date>.
+                    <ref target="https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/">https://www.statista.com/chart/19708/known-occasions-where-people-survived-falls/</ref>
```
sydb: Probably should be `<ptr>`, not `<ref>`.

sydb (changes requested):

lb42:
"these guidelines make no recomm endation as to...." is one of the more annoying tropes in the gl prose. Since we're making this up anyway would it do no harm to actually take a position as to whether in general @foo = "#a #b" means #a AND #B or #a OR #b (my preference being the former)

Reluctantly using Outlook for Android<https://aka.ms/ghei36>

P5/Source/Specs/att.locatable.xml
sydb: If we are developing enough spine to actually say what multiple values on `@where` means, I would suggest something like
~~~xml
<p>This attribute may be used to point to multiple canonical
  descriptions simultaneously, in which case the implication is that
  the reference is to multiple places. For example, the November 2015
  Paris Attacks took place at 6 different locations in Paris; this
  might be encoded as an <gi>event</gi> with a <att>where</att>
  attribute with 6 space-separated values.</p>
~~~
But I bet you (@ebeshero or @lb42) could do better than that.
In the prose in ND we should discuss how to encode “here are several places you can go for more information about the place this event took place or where this conversion was used” and “here are several locations, one of which is where this event took place or this conversion was used”. I do not actually know off the top of my head how to encode “here are several locations, each of which may or may not have been a place this event happened or this conversion was used”.

ebeshero: @sydb I believe the example provided by Till Grallert on the ticket provides exactly the use we need, and the idea is quite simply that ALL of the locations are where this conversion is used. I don't believe any other modification is necessary.

If we aren't sure and are providing guesses, well, that's a different problem that would seem to require a different encoding. Perhaps we develop a spine enough to say, this is simply for **indicating** multiple locations for an event or where a conversion was used, not for creating a list of possibilities. I don't think it should be used in that way at all.

ebeshero: @sydb I am frankly exhausted and it is close to release time. 

 I can add prose that explains exactly what multiple values of `@where` should mean.  But I think it's out of scope of this PR and the release of att.locatable to explain how to encode where you can go for more information (and honestly I don't think that's even necessary, since it sounds like you're talking about pointing to a bibliography anyway). I don't think it's necessary for me right now before this release to write new examples to that represent *one of a list of possible* locations for a conversion or an event or conversion. I suppose I could vaguely suggest an encoder might want to use `<cert>` or `@cert`, and perhaps in those cases the encoder would be better off itemizing each location singly to specify the certainty level (whatever that means). Better yet, *only* point to locations in `@where` when you're confident of your assertions. When you want to express a lack of certainty, you could always do so the old-fashioned way with a nice `<note>` in which you discuss and point to those locations and encode your low level of certainty or confidence. 

Can we agree, in the interest of time, that the only _necessary_ thing to do right now before this release is to clarify that multiple values of `@where` simply mean multiple places at which a specific conversion applies, and multiple locations where an event occurred. 

If you want to take time to provide some encoding that shows people how to point to locations they have little confidence are actually accurate, please do so. I can't do that this week. I'm sorry.

ebeshero: @sydb I should probably add, I think you could well open a new ticket requesting examples that invite you to point with low confidence to one or some of a set, or to point to a full list of locations where more information is available. If we can agree that this is simply not what att.locatable was designed to do, can we please pause this for now and pick it up later in a new ticket?

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -1799,10 +1799,16 @@ more than one language, as in the following example:
...
+Within the <gi>unitDef</gi>, a <gi>conversion</gi> element may be used to store information relating to conversion between units. 
+      The <gi>conversion</gi> element holds a special pair of attributes, <att>fromUnit</att> and <att>toUnit</att>, which serve to indicate the direction of a calculation from one unit of measure (stored in <att>fromUnit</att>) to another (stored in <att>toUnit</att>). 
+      A mathematical calculation to define the relation between these units may be stored in <att>formula</att>, as shown in the following examples.  
+      The <att>formula</att> attribute takes a value expressed as an XPath expression, for which see <ptr target="#XPATH31"/>. 
+      Among other things, this means that division must be expressed with <q>div</q> as opposed to the more common forward slash (<q>/</q>, U+002F), so as not to be confused with path navigation.
+      The <gi>conversion</gi> element may also use attributes to store dates and date ranges as well as location(s) that specify when and where a conversion formula was applied.
```
ebeshero, line 1811: @sydb , really, this statement (line 1811) covers *exactly* what we mean. I don't really think it requires anything more to clarify.

Issue #2134, opened 2021-04-27
calendar example mismatch?


sydb:
We (@MegJBrown and I) notice some possible inconsistencies in section [13.4.4 Using Non-Gregorian Calendars](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ND.html#NDDATECUSTOM) while working on #2028. To wit:

The initial example sets up a calendar definition for `Julian_England`. We think the definition (“The Julian calendar, as used in late 16th-century England.”), is either lame or does not match the following examples. (It is true, of course, because the Julian calendar was in use in England until 1752.)

The examples that follow point to a _slightly_ different calendar `julianEngland`, which is defined as “Julian Calendar (including proleptic)” in the `<calendar>` tagdoc. If these two examples are supposed to point to the calendar just defined (i.e., this is just a typo), then the entire example setup seems weird, because these two examples are not from late 16th-century England. One is had a date from the 13th century, but certainly could have been written in the late 16th. But the other is from the early 17th century!

If it is not just a typo, but supposed to be a different calendar, we should probably show people where to find the definition.

martindholmes:
There were multiple variants of the Julian calendar in use over the centuries in England; IIRC, our intention was to specify that these dates were in the form one would expect to find most frequently in the last 16th century. Any of these dates could be expressed in any variant of that calendar. I'm sure the id/ref mismatch is just a typo.

lb42:
It's quite clearly a typo . However, while you're fixing that, it might be worth doing something about this sentence:

"However, just as we use, for instance, when, notBefore, notAfter etc. to provide more precise expressions of dates and times in a constrained and computable form, it is often necessary to express a date or a date-range from a non-Gregorian calendar in a more precise manner"

I take exception to "it is often necessary..." because it's only necessary if you are dealing with non-Gregorian calendars, which is not the general case, important though it is to those who do so deal! And the sentence is rather convoluted too. How about this:

"For reasons of consistency and computability, it is often useful to complement the content of a <date> element with attribute values such as when, notBefore, notAfter, etc. For dates expressed in a non-Gregorian calendar, such normalizations may also be  expressed in terms of the non-Gregorian calendar specified by the @calendar attribute. The attributes @when-custom etc. are provided for this purpose."

martindholmes:
@lb42 Not `@calendar`, `@datingMethod`. That's more important than the convoluted style, for which I expect I'm responsible. :-)

lb42:
Yes, quite.

[...]

PR #2172, opened 2021-08-17
Fixing language in #2134

MegJBrown:
Made changes inspired by @lb42's wording, fixed IDs in examples, and fixed definition of Julian Calendar for England.


martindholmes (commented):
Looks much better. The only change I would make is that I think the word "attribute" should be in here:

`the <att>datingMethod</att> is used to
          identify the calendar`
          
becomes:

the <att>datingMethod</att> attribute is used to
          identify the calendar

PR #2174, opened 2021-08-22
inconsistent wording of attr classes 2142

martinascholger:
This PR adresses #2142. I've changed the descriptions to plurals. In addition I've changed descriptions where the attributes were listed in the `<desc>` element, and I replaced 'supplies' with 'provides'.

PR #2176, opened 2021-08-26
revising restrictive pronouns in specs: which to that

ebeshero:
This is reopening the [just-closed PR](https://github.com/TEIC/TEI/pull/2174) to finish what we just discussed: changing the restrictive pronouns in the spec `<desc>` elements from "which" to "that". These are pretty small. Just reopening for a moment to register completion of the task.


ebeshero:
Merging to dev since this is pretty minor and finishes the task we started.

PR #2184, opened 2021-09-17
Sydb stylesheets issue 417

martindholmes:
This fix addresses Stylesheets issue 417 (https://github.com/TEIC/Stylesheets/issues/417), which turned out to be a problem in the TEI repo rather than the Stylesheets repo. Schematron validation will no longer trigger an error if an expired date is used for `@validUntil` inside an `<egXML>`, but other checks on `@validUntil` will still be effective.

### DIFF
--- a/P5/Utilities/extractegXML.xsl
+++ b/P5/Utilities/extractegXML.xsl
@@ -10,10 +10,11 @@
     ago by the amazing Sebastian Rahtz.
     
     Read in TEI P5, write out
-    a) a directory (./valid/) full of one-file-per-example each of which has a
+    a) a directory (./valid/) full of files, one for each example in
+       P5, each of which contains nothing but a copy of that single
        single example
-    b) a driver file (normal output of this stylesheet) that refers to each of
-       those files.
+    b) a driver file (normal output of this stylesheet) that refers to
+       each of those files.
 
     (Above is part of original program; feature described below added 2021-09-17
     by Syd Bauman and Martin Holmes in attempt to fix Stylesheets issue 417.)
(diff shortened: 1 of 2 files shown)


HelenaSabel (approved):
I verified the code and ran some tests and everything worked as expected.

P5/Utilities/extractegXML.xsl
```diff
@@ -19,65 +48,88 @@
...
\ No newline at end of file
+  </xsl:template>
+  
+  <xsl:template match="@validUntil" mode="copy_egXML_checking_validUntil">
+    <!--
+      First, knock off leading & trailing spaces, which are allowed in the
+      attribute value but may not be in the functions we're calling.
+    -->
+    <xsl:variable name="validUntil_date" select="normalize-space(.)"/>
+    <!--
```
peterstadler, line 140: It took me some time to understand the magic (aka hack) here and I think it looks good. I would only recommend to add some warning about this particular check `$validUntil_date castable as xs:date` that this refers to the current datatype of the validUntil attribute. I.e., if we are to change it, we need to adjust the test. Hence it would be good to add this notice to att.deprecated.xml as well.

peterstadler (commented):
see details below

P5/Utilities/extractegXML.xsl
sydb: Well, seems like overkill to me. But since the error of over-documenting is far less problematic than the error of under-documenting, I will add comments to that effect both here and to att.deprecated.xml.
Will push results into this branch.

sydb: Done at 17a189797.

PR #2188, opened 2021-10-01
Fix #2187, I hope

### LINKED ISSUE #2187: subset does not contain `<constraintSpec>`s from other places
I just discovered that p5subset.xml only retains those `<constraintSpec>`s that happen to have been inside an `<elementSpec>`, `<macroSpec>`, `<classSpec>`, `<moduleSpec>`, or `<dataSpec>`. “But that’s all of them.” you say. Well, it was, which is why no one ever noticed this before.

But `<constraintSpec>` is allowed in all sorts of other places, most of which are ridiculous. But certainly inside a `<specGrp>` is not crazy, and a `<constraintSpec>` from anyplace, even if weird, crazy, or ridiculous should be copied into p5subset, no?

I have not tested it yet, but I believe the fix will be trivial. (Just adding `constraintSpec` to the list of elements matched by the `keep_important_stuff_with_contents` template of P5/Utilities/subset.xsl.)

sydb:
Just adding `<constraintSpec>` to elements that get copied to subset.


martindholmes (approved):
Looks good to me.

HelenaSabel:
Test fails with error: `/tei/P5/Test/detest.odd:457:86: error: element "constraintSpec" incomplete; missing required element "constraint"`

European VF2F thinks that the PR should also include the fix of the test.

sydb:
Ummm… the error is catching a real problem, isn’t it? That the content model of `<constraintSpec>` _requires_ a child `<constraint>` even if `@mode` is "delete". So it seems to me the right solution is to add an empty `<constraint>` to the detest.odd file, and open another ticket that the _Guidelines_ should properly explain how to delete or change a `<constraintSpec>`. (Which might include changing content models and adding constraints.)

sydb:
Council meeting of today votes to change Stylesheets so that 
~~~xml
<constraintSpec mode="delete" scheme="schematron"/>
~~~
is what is used to delete a constraint specification, not requiring a child `<constraint>`.

sydb:
OK. Schema, not test, fixed. I think this is quite ready to be merged, now.

peterstadler (approved):
I'm unable to find the change that caused the error reported at https://github.com/TEIC/TEI/pull/2188#issuecomment-1086611851 to vanish, but hey, seems ok to me :)

PR #2193, opened 2021-10-11
deprecate <altIdent> in bizarre places

sydb:
This PR is for moving forward on issue #2049. It tackles items (3), (4), and (12) in the list of things to do in the **Plan** section of the original post on that ticket.

### Dependency
This PR is dependent on the recent TEIC/Stylesheets#509 issue and TEIC/Stylesheets#510 PR that allow us to use `<idno>` instead of `<altIdent>` to specify the FPI of a module. If, for some reason, that change was backed out this one would have to be, too.

### Main change
The main changes are to the definition of `<altIdent>` itself (i.e., to Source/Specs/altIdent.xml) in commit 7a7f614a3. Although there are several whitespace changes to that file, the only significant changes are the addition of two `<constraintSpec>`s. In case you prefer to view it as formatted output instead of the raw ODD XML, I have made a copy of the HTML [tagdoc](https://bauman.zapto.org/~syd/temp/4TEICouncil/ref-altIdent_2049.html) available.

### Secondary changes
The rest of the changes in this PR deal with the fallout from the new constraints on `<altIdent>` (both the one added ~4 months ago and those added yesterday).

#### Using `<idno>` in source files
First, in order to avoid the stream of errors about existing `<altIdent>` used for FPIs of modules, we either have to change the expected results of validation or change those `<altIdent>`s to `<idno>` (step 12). I chose the latter, and thus the changes to the 21 chapter files in commit 792a126c4 in this PR are pretty much just that. (Also changed the 1 example in TD; as far as I know there is no prose discussion of this usage to be changed.)

#### Making above valid
Second, now that the FPIs are encoded with `<idno>` not `<altIdent>` (so that the Schematron validation does not scream),[1] the regular RELAX NG validation was upset because `<idno>` is not permitted as a child of `<moduleSpec>`. Thus in commit da77ca32e I changed the content model of `<moduleSpec>` to allow `<idno>` (and also changed the 1 example in the tagdoc), and changed the definition of `<idno>` in p5odds.odd to permit `"FPI"` as a value of `@type`.

### More to come
NB: When this is merged the branch should **not** be deleted. There is still quite a bit more work to do on the ticket.

## Notes
[1] In retrospect, this was probably not necessary. Since the Schematron rules have `@role` of `"nonfatal"`, they probably would not break the build.


martindholmes:
HI @sydb. Generally this looks OK to me, but I'm a little uneasy about the wording of this Schematron error message:

~~~~~
The <gi>altIdent</gi> element should contain a single <ident
      type="datatype">teidata.xmlName</ident> except when it is
      used to indicate the FPI (of a module), in which case the
      <gi>idno</gi> element should be used instead.
~~~~~

It seems to suggest initially that it's perfectly OK to use `<altIdent>` to specify the FPI of a module, then it reverses course and says you shouldn't do that. It might be clearer to say something like "If you are using this to indicate the Formal Public Identifier of a module, you should use the `<idno>` element instead." Secondly, the meaning of FPI is not specified, and of course `@type="FPI"` is not a suggested value for `@type` on `<altIdent>`. I know it's now defined on `<idno>`, but it's not obvious that's where you should look. So I would recommend just expanding it in the Schematron rule. I also think the definition on `<idno>` should perhaps do a little more than simply expanding FPI to Formal Public Identifier; it might point to Wikipedia and also explain what the TEI particularly means in using this to identify (say) a module.

sydb:
I was mildly unhappy with the wording of that description when I wrote it, but at the time could not come up with anything better. But keep in mind it is a description of the constraint, not the message the user gets. (Which is just “The content of this 'altIdent' element (_string_) should conform to teidata.xmlName.”, and that message does not occur if the `<altIdent>` in question is either a child of `<moduleSpec>` or has a `@type` of `"FPI"`.)
As for the wording, how about
> Other than when used to indicate the formal public identifier of a module (which is a deprecated usage), the content of `<altIdent>` should be a single `teidata.xmlName`.

martindholmes:
@sydb I like that wording. Everything else looks good to me.

sydb:
@npcole, @HelenaSabel, and @martindholmes:
I have _finally_ pushed the wording change @martindholmes and I agreed to back in Oct 21.

My suggestion for next step is either @npcole or @HelenaSabel should review, and if review is OK the other should merge (since @MegJBrown is not likely to do so).

HelenaSabel (approved):
Everything looks great, no missing `<altIdent>` elements found, and all the tests passed!

Issue #2179, opened 2021-08-31
TD section on schematron constraints needs work


lb42:
Here are some things thatoccurred to me  when re-reading section TDTAGCONS (22.5.2)

[For dispositions as of 2023-11-10 see #issuecomment-1711937223; short notes appended to each point here.]

1. The first four paras repeat each other. — DONE, per PR #2500.
2. The content model for constraintSpec shows that `<constraint>` is optional within it. What does it mean to have a constraintSpec with no constraint? — DONE
3. It would be helpful to explain why namespace prefixes are necessary within `<constraint>` (an analogy with xslt might help); where is the prefix `s:` defined? — deferred to #2330
4. The  language used by constraint messages varies wildly between formal passives ("x should be,,,")  and hortatory ("you should...") : either be consistent or thematize that this is a project choice — discussion needed
5.   The phrase " note here the use of a report and an assertion within one pattern:" is not very informative. This section doesn't have to provide a full schematron tutorial (though it might be useful to reference one) but gnomic remarks like this don't help. — discussion needed
6. That last example is the only place where I see a `<s:rule>` with a `@context` : this really needs to be explained more fully (e.g. to explain that most schematron constraints in TEI assume a default context and what that is) — will probably become irrelevant as we will likely require an `<sch:rule>` (with a `@context`, it is a required attribute) in _every_ Schematron `<constraint>`.

sydb:
Thanks, @lb42. I think this is mostly spot-on, and at least 2, 3, 4, & 6 should be addressed toot sweet. (I don’t really see much problematic repetitiveness in 1st 4 paras, and think 5 requires some thought or discussion. And as for namespaces, I think it is a bit worse than you describe, as a TEI customizer does _not_ have to declare the TEI namespace. (It is declared in P5 in [the definition of `<TEI>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-TEI.html) — which brings up another question, for which if no one has a good answer here, I will create a new ticket: why are `tei:` and `xs:` defined separately from `rng:`?))

martindholmes:
@lb42 @sydb and I have already been talking about the question of Schematron with and without `@context`. Since there are contexts in a `<schemaSpec>` in which  it's not possible for the processing to usefully infer what `@context` should be, I personally think that `@context` should be mandatory, enforced through Schematron itself, but at the very least we should explain clearly in what contexts it's necessary and not necessary, and why.

sydb:
[...]

PR #2196, opened 2021-10-13
#2179, item 2:

sydb:
Insist that `<constraintSpec>` elements in mode "add" or "replace" do have a child `<constraint>`, and those in "delete" do not.


martindholmes (commented):
Missing word "have" in this:

"...but does not a 'constraint' child element...." 

Also, I think this:

...has a mode= of "delete"...

should either be:

...has @mode="delete"...

or

...has a mode attribute whose value is "delete"...

Other than that, this looks great.

sydb:
Merging this unreviewed PR myself (in direct opposition to our normal practice) because I am reasonably convinced it will help fix our broken build.

PR #2201, opened 2021-10-21
Translation from raffazizzi/TEI

raffazizzi:


P5/Source/Specs/abstract.xml
```diff
@@ -10,6 +10,7 @@ $Id$
 <elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="header" ident="abstract">
   <desc versionDate="2012-12-27" xml:lang="en">contains a summary or formal
     abstract prefixed to an existing source document by the encoder.</desc>
+  <desc versiondate="2021-10-21" xml:lang="it">contiene un riepilogo o <foreign>abstract</foreign> formale aggiunto, in testa a un dato documento di origine, da chi codifica.   </desc>
```
sabineseifert: It also should be `@versionDate` with an upper case 'D'.

sabineseifert (changes requested):
See individual comment.

peterstadler:
Hi @raffazizzi, thanks for quick response! (sitting here with @sabineseifert on Zoom :-) 
It seems your PR features an outdated GH action that fails. Could you please update/rebase your PR on current dev to fix this?

sabineseifert (approved):
Thanks for the quick fix.

raffazizzi:
> It seems your PR features an outdated GH action that fails. Could you please update/rebase your PR on current dev to fix this?

Ok I think I've done it, but this could be a bit of an issue for the translation tool since the branch was created automatically on my fork and other users who contribute translations may not be able to make these fixes. cc @hcayless

raffazizzi:
Or maybe this isn't related to the tool, or it's an early test -- sorry can't quite remember

martinascholger:
Yes, I think this is an early test with the translation tool.

peterstadler:
My bad: the failure wasn't due to an outdated GH action but due to a missing secret "SLACK_WEBHOOK" in @raffazizzi 's fork. I will remove the Slack notification for PRs in general (or make that a non-critical failure) because that will be the case for all forks.

peterstadler:
… and I think this PR is safe to merge now :)

Issue #2185, opened 2021-09-23
xs:anyURI permits spaces, so we need teidata.anyURI to replace it


martindholmes:
@sydb's work on TCW32 and subsequent testing has revealed a weird anomaly that affects all the TEI attributes that permit multiple space-separated xs:anyURI values: Although [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt) specifies that spaces are excluded from URIs, the W3C XML Schema Part 2 spec [says that](https://www.w3.org/TR/xmlschema-2/#anyURI) they're allowed, although discouraged. That means that when we try to constrain an attribute value to a single xs:anyURI, the constraint fails because multiple values are valid as a single URI.

The Stylesheets Working Group suggests that we substitute our own datatype which does not permit unencoded spaces.

sydb:
Problem, as @martindholmes says above, is that a processor cannot differentiate between an xs:anyURI that happens to contain a space and two xs:anyURIs separated by a space.

But the solution, it turns out, is probably simpler than creating a new datatype (which I had suggested, so not ragging on Martin, here). Probably just changing the definition of teidata.pointer to
~~~xml
          <content>
            <dataRef restriction="\S+" name="anyURI"/>
          </content>
~~~
would do the trick for most things. (The `@property` of `<citeData>` uses xs:anyURI directly — incorrectly, IMHO — so it would have to be fixed. And we would have to decide what to do with teidata.namespace — probably just do the same thing. As for teidata.namespaceOrName I am not sure what we would do, as it already uses `<dataFacet>`; maybe they can be merged.)

sydb:
Note-to-self: PR #2204 should fix this. (Isn’t GitHub supposed to automatically link these two?)

sydb:
Please see [comment about problem](https://github.com/TEIC/TEI/pull/2204#issuecomment-981709846) (with spaces in the fragment identifier portion, in this case in arguments passed to an XPointer scheme; same could occur in query portion of URI, though) on the PR.

sydb:
Note also that although RFC 2396 forbids spaces, that spec is obsoleted by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A).

PR #2204, opened 2021-11-03
Address #2185 by restricting anyURI to not have spaces

sydb:
Changes:

- Change `citeData/@property` to use teidata.pointer rather than xs:anyURI directly
- Add `restriction="\S+"` to the `<dataRef>` that defines **teidata.pointer**
- Add `restriction="\S*"` to the `<dataRef>` that defines **teidata.namespace** (not ‘+’ because in some cases `ns=""` is actually used)
- Change pattern restricting xs:anyURI for **teidata.namespaceOrName** from `[^/]+:.*` to `[^/\s]+:\S*`.

Passes build locally. I have not tested the namespaceOrName bit thoroughly, though, but at least on a quick test it seems to work.


sydb:
Interesting. In chapter TC there is an example that includes 
~~~xml
        <app from="example.xml#string-range(WP.1a, 0, 10)">
~~~
Which is (with this PR) invalid. Possible solutions include:

1. Whitespace should not be allowed in a URI (lest it is hard to impossible to differentiate one URI from multiple URIs):
   1. encode this example (and all similar such examples) without the spaces: `<app from="example.xml#string-range(WP.1a,0,10)">`.
   1. encode this example (and all similar such examples) by percent-escaping the spaces: `<app from="example.xml#string-range(WP.1a,%200,%2010)">`.
   1. demonstrate both of the above solutions in the _Guidelines_.
1. Whitespace inside the argument of an XPointer scheme like this is obviously fine, and can be differentiated from multiple URIs by parsing on the parentheses:
   1. update the constraint so it is smart enough to differentiate; do the same for the query portion of URI, while we’re at it. (This may be hard to impossible to do correctly.)
   1. too hard for TEI to provide generic constraints, users need to provide them in their ODDs. Throw away this PR, and flag the ticket as “won’t fix”.

martindholmes:
I agree with @sydb that 1.iii is the best solution here.

sydb:
And maybe add a health warning reminding folks to check that their XPointer processor knows how to handle percent-encoded stuff.

sydb:
In commit 5ee708b0f @martindholmes and I have implemented solution 1.i: we have gone about removing all spaces from pointers in examples and prose of the _Guidelines_. Note that we only found 1 case with _internal_ whitespace (the one mentioned above); all the other cases were other _leading_ or _trailing_ whitespace. Much less of a problem, but we fixed ’em anyway. (Also fixed an error in an attr that used "Le Bars" when it meant "#Le_Bars".)

sydb:
Per Council meeting yesterday, @hcayless is tasked with adding prose to SA and the teidata.pointer tagdoc discussing when percent escaping should be used. Thus removing him as a reviewer here, adding him as an assignee to the ticket, #2185.

sydb:
Pinging @martindholmes and @npcole to review.

martindholmes (approved):
Looks good to me. What's the conflict in the French header file about, though?

P5/Source/Specs/teidata.namespace.xml
```diff
@@ -35,7 +35,7 @@ See the file COPYING.txt for details
     indicare i nomi degli spazi in XML come stabilito dalle raccomandazioni tecniche del W3C per gli
     <ref target="http://www.w3.org/TR/1999/REC-xml-names-19990114/">spazi dei nomi in XML</ref>.</desc>
   <content>
-      <dataRef name="anyURI"/>
+      <dataRef restriction="\S*" name="anyURI"/>
```
tuurma, line 38: why `\S*` and not `\S+` here?

sydb: Good question. Because the `@ns` of some things (in particular, `<attDef>`) sometimes needs to be empty to specify “in no namespace”.
I would not mind changing to `\S+` here and solving that problem (the need to be able to specify “no namespace”) by having the defintion of things like `attDef/@ns` be `"teidata.namespace | ''"` or `"teidata.namespace | '%NONE'"`.

sydb: Also, the current datatype (xs:anyURI without any restriction) allows nil. Thus I am going to assert that this ticket should not require characters to occur in the value of teidata.namespace. If we think one or more should be required (i.e., that nil should not be allowed, and the need to express “no namespace” should be moved to the definition of the attribute), that’s another ticket.

tuurma (approved):

tuurma:
Ready to be merged, I think @hcayless

PR #2207, opened 2021-11-08
Update link in example

rettinghaus:
This PR not only updates the link in the example to the latest version of the file in the current repo, it actually points now to the raw version (which indeed matches the given mimeType) instead of the html representation of the file within the repo.


sabineseifert (approved):

sabineseifert:
@JanelleJenstad Can you have a look at this issue that everything is ok?! Thanks!

PR #2224, opened 2022-01-28
make identifiers of constraint specifications unique

sydb:
Sorry, @peterstadler, but there are some white space changes in here, too.

This is a pretty simple PR, so I think the reviewer can merge it if she finds no problems. That said, I would give this a few days to see if there is any discussion on the issue on TEI-L, first.


HelenaSabel (approved):
Everything looks great. The tests passed and my own queries didn’t catch any missing repetitions of `@ident`.

PR #2229, opened 2022-02-11
Translation from martinascholger/TEI

martinascholger:


peterstadler:
@martinascholger I think this is by accident, again? It still has the old lowercase `@versiondate` attribute … Can we reject this PR or do you want to update it manually?

martinascholger:
Yes, we can close the issue. But we should investigate what happened. I suspect that I accidentially clicked on "submit" while Hugh presented the tool to us, but I didn't do any translations. And now the old changes (from 4 month ago) are back. @hcayless do you have an idea?

hcayless:
If your fork contained work from before I fixed the bug, then submitting it would just reopen the PR. It might be safest to go in and delete the branch you were using.

PR #2230, opened 2022-02-15
Issue 2217: better targets for PDF in P5/Makefile

### LINKED ISSUE #2217: Makefile target names for PDF are confusing
The Makefile has 3 targets for PDF: `Guidelines-pdf:`, `pdf-complete:`, and `pdf:`. When I inadvertently use the wrong one, you can chalk it up to my being a PDF-generation novice. But @martindholmes just made the same mistake, too, and he is well versed in this stuff. So I claim the names are misleading, and plan to change them:

- **pdf:** That which a user should use to build the PDF version of the _Guidelines_ — stays the same
- **Guielines-pdf:**  That which runs the first part of building the PDF version, running the `pdfonce` target in the antbuilder.xml file (after running `check.stamp`,  `p5.xml`, and checking `Utilities/guidelines-latex.xsl`) — rename to **pdf-init**
- **pdf-complete:** That which runs the final part of building the PDF version, running the `pdfrest` target in the antbuilder.xml file (why doesn’t it run `Guidelines-pdf:` first, anyway?) — rename to **pdf-finish**

sydb:
Try only 2 targets for PDF per @peterstadler suggestion of 2022-01-12.

### DIFF
--- a/P5/Makefile
+++ b/P5/Makefile
@@ -110,8 +110,6 @@ teiwebsiteguidelines:
 	make html-web ALLLANGUAGES="en es de ja ko fr it zh-TW" GOOGLEANALYTICS=UA-4372657-1
 	(cd Guidelines-web; zip -r -q ../teiwebsiteguidelines.zip . )
 
-pdf: Guidelines.pdf pdf-complete
-
 epub: Guidelines.epub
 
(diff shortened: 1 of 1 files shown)


martindholmes (approved):
Tested, and 'make pdf' works perfectly for me.

PR #2233, opened 2022-02-20
Translation from susannalles/TEI

susannalles:


P5/Source/Specs/TEI.xml
```diff
@@ -83,6 +81,7 @@ $Id$
           of a more precise association provided by the <att>source</att> attribute on the associated 
           <gi>schemaSpec</gi>. </p>
       </remarks>
+      <remarks versionDate="2022-02-20" xml:lang="es">Las ediciones más importantes de la Guías Directrices se han referido informalmente durante mucho tiempo a través de un nombre creado a partir de la letra P (por Propuesta) seguido de un dígito. La versión actual es una de tantas versiones de la quina edición de la Guías, conocida como P5. Este atributo puede ser utilizado para asociar un documento TEI con una versión específica de las Guías P5, en su ausencia de una asociación más precisa proporcionada por el atributo <att>source</att> en el <gi>schemaSpec</gi> asociado. </remarks>
```
HelenaSabel: typo: quina > quinta

P5/Source/Specs/TEI.xml
```diff
@@ -199,6 +198,7 @@ $Id$
     TEI namespace <code>http://www.tei-c.org/ns/1.0</code> on it, using
     the <att>xmlns</att> attribute.</p>
   </remarks>
+  <remarks versionDate="2022-02-20" xml:lang="es"><p>El elemento es obligatorio. Normalmente se especifica el espacio de nombre TEI en su interior <code>http://www.tei-c.org/ns/1.0</code>, utilizando el atributo <att>xmlns</att></p></remarks>
```
HelenaSabel: I think the most common term in Spanish (hopefully not just in Peninsular Spanish) is “espacio de nombres” (my suggestion here would be: “el espacio de nombres de TEI”)

HelenaSabel (changes requested):

P5/Source/Specs/TEI.xml
hcayless: I was just looking at a couple of cases where "namespace" is translated into Spanish, and the translation on the Mozilla Developer Network at least is "namespace". I assume that's because it's a technical term, not really English (even though it's obviously composed of English words). The concept afaik is exclusive to programming languages (and XML).  It also strikes me that maybe I should add notifications to the translation tool so users will see review comments like these there...

HelenaSabel: I think the MDN can be consider an authority source. However, to add a a reference in support of “espacio de nombres” instead, there is the [Spanish Wikipedia article](https://es.wikipedia.org/wiki/Espacio_de_nombres) which provides a couple of other references in Spanish that use it.

hcayless: That's fair. In that case, I'd recommend glossing it the other way, like Wikipedia does "espacio de nombres (namespace)".

sydb: See #2236.

susannalles: > 
Having a feature to see comments would be great, but with some of my students were talking that having a controlled vocabulary is quite important... (don't know in which state remained Raff spreadsheet) (like, is it "espacio de nombres" or "espacio de nombre", header is "encabezado" o "cabecera", etc.).

susannalles: Most probably you are right @HelenaSabel ;) In my case I always used it in singular since we normally use one name space (but we can have multiples).

P5/Source/Specs/TEI.xml
sabineseifert: Done.

P5/Source/Specs/TEI.xml
sabineseifert: Update as suggested ("espacio de nombres (namespace) de TEI"). Please have a look again.

sabineseifert:
Merged current dev branch into pull request and resolved conflicts by choosing the more recent wordings from the dev branch. Please feel free to take this as a starting point for further wordsmithing and corrections/additions.

HelenaSabel (approved):

Issue #2106, opened 2021-02-11
broken links to non-English bibliography entries from examples-* pages


peterstadler:
While refactoring our Linkchecker Jenkins job I came across various internal references to missing entries in the bibliography from some examples-* pages. Turns out the references are to non-english entries in the bibliography, e.g. the first example on  https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/doc/tei-p5-doc/en/html/examples-condition.html which links to the "bibliography"  like 
```html
<a href="javascript:void(0)" onclick="showPopupBibl('fr-ex-BnF-Reliures')">bibliography</a>
```

Yet, the target "fr-ex-BnF-Reliures" is only declared in the french section of the bibliography and unavailable to the english HTML version.
https://github.com/TEIC/TEI/blob/efeb3baae510141c70fc497e699a4d4c13df2029/P5/Source/Guidelines/en/BIB-Bibliography.xml#L1031

Quick searching in the TEI and the Stylesheets repo showed some related tickets (#1476, https://github.com/TEIC/Stylesheets/issues/173), yet I didn't find a description of this particular issue.

martindholmes:
The JavaScript there is misleading; in fact in the source it's a regular link:

`<a href="BIB.html#fr-ex-BnF-Reliures">bibliography</a>`

which is converted into a popup call on load, if JS is enabled. The problem is either:

 - The examples pages themselves should not include examples from other languages than the current one; or
 - The bibliography should include all language items, whatever the current GL build language is; or
 - Examples from other languages should not have bibl links attached to them; or
 - Examples in other languages should link to the bibliography of the other language (potentially quite confusing, I think).

I think I favour the second.

hcayless:
I think the ideal would be a single bibliography that was able to handle works in multiple languages (so if there are both English and French versions of a book, e.g.), they would both be listed under the same heading. I haven't looked at all at what might be involved in making that work...

ebeshero:
On Council discussion today (2021-12-09), this is green for @HelenaSabel to check to see if we have any examples of a `<bibl>` resource expressed in multiple languages, and if we do, to try to merge them into one entry. Move the `@xml:lang` into the freshly merged `<bibl>` then.

ebeshero:
[...]

PR #2235, opened 2022-02-20
Issue 2106: broken links to non-English bibliography entries

HelenaSabel:
PR to tackle issue #2106.

The decided solution is to merge the two language-specific `<listBibl>` (French and Taiwanese Mandarin) into the main `<listBibl>` of the examples. 

One of the assigned tasks was to verify whether there were any cases of the same bibliographic work in two different languages. This was not the case, but I did encounter a reference to the same work (in the same language) that entails two different `<biblScope>` elements and thus is encoded as two independent `<bibl>` elements; see [entries 166 and 167 in the current release of the Guidelines](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/BIB.html). If this is the decided criterion, it would make sense that for different languages, two different entries are created which means that merging these `<listBibl>` was even less problematic than originally thought.

Thus, the main difficulty of the merge is the ordering of the new `<listBibl>`. The encoding of the bibliographic references is not consistent, which complicates matters. However, my main doubt is how to deal with alphabetic ordering of non-Latin script languages. I am only familiar with Latin and Greek scripts and the convention I know is to order first the Latin script and then the Greek one. However, by looking at the current release, I see that entries 208 and 295 have specific positions which makes me think that there might be a different criterion at play other than just ordering the scripts individually. Maybe one of the Japanese-speaker colleagues (@knagasaki ?) could confirm that the order of Japanese entries in the [current release](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/BIB.html) is the expected one? 

The Taiwanese mandarin entries were ordered in their own block, but with the merging of the lists, I also wonder what their optimal location is.


knagasaki:
> 

Thank you for confirming it.
In the alphabetic ordering based on pronounciation,
No. 208 (Author: Osumi, Akiko) is put on an appropriate place.
However, the author name of the No. 295 is "Sakiyama, Osamu" (not Zakiyama). So, move it to No. 232 or so.

HelenaSabel:
Thank you very much, @knagasaki!

sydb:
I strongly suspect that the duplication of the bibliographic entries for examples taken form Terry’s book (166 & 167) is attributable to two different authors writing those parts of the _Guidelines_. In any case, seems pretty clear to me that either

1. The entire work should be cited (like 167, but not scoped to just page 1), and both examples should point to the one entry that cites it; OR
2. The specific spot in the work should be cited in both cases (i.e., put more info into 167 to make it like 166 — probably requires someone finding a physical copy of the book or getting PDF, e.g. from [researchgate](https://www.researchgate.net/publication/230876242_The_Vastness_of_Natural_Languages); I believe the 1984 edition has ISBN 0631134611.)

HelenaSabel:
I added the “Needs discussion” label because I am not familiar with the right convention to order the Taiwanese mandarin entries. 1) Should they be their “own block” or 2) should they be ordered based on pronunciation like for Japanese?

sydb:
I do not know, @HelenaSabel, but @mbingenheimer (Marcus Bingenheimer <bingenheimer@temple.edu>), @cwittern (Christian Wittern <cwittern@gmail.com>), and @acmuller (Charles Muller <acmuller@l.u-tokyo.ac.jp>) would be the first people I would ask.

747:
@HelenaSabel 

> ... 1) Should they be their “own block” or 2) should they be ordered based on pronunciation like for Japanese?

Since Taiwan does not have an agreed phonetic representation of Chinese words, a Han script author-year bibliography is sorted on author's name (surname first) by stroke count and stroke type (much the same way as the Beijing Olympics ordering, but need to follow the Taiwanese stroke order). As a result, Japanese (and maybe Korean) literature is mergeable, but characters should be transcribed into the Traditional variant. The Han and alphabetical bibliographies thus must be kept in separate blocks, and the former comes first.

HelenaSabel:
@JanelleJenstad @bleekere @ebeshero I’ve merged the bibliographies following the discussion we had in the TEIC meeting.  I am sure there are errors in the sorting, but at least this fixes the broken links problem (which was the main issue).

I’ve tried to correct some inconsistencies of the bibliographic references, but this would be a completely new ticket and would need a thorough review with a predefined template for bibliographic citations. In any case, you will notice that there are changes in the `Specs` because there were some entries that were repeated (but with different IDs) thus I included a single entry and then corrected the URIs accordingly.

JanelleJenstad (approved):
Thanks for doing all this work! We'll need to make a note to check sorting after the relesease.

ebeshero (approved):
Any and all repairs to the confounding BIB are welcome, and there will always be more to correct. Thank you for your systematic care with this, @HelenaSabel!

ebeshero:
I'm taking the liberty of accepting this PR in the interests of time, as the clock is ticking for the upcoming release and it's a holiday weekend.

PR #2232, opened 2022-02-19
Remove those constructs deprecated for Unihan

sydb:
I _think_ I have this right. See #1805.
I wanted to ask @duncdrum to review, but he was not in the drop-down list of possibilities.
I made this high priority since the build is at least messed up, if not broken, until these deprecations are removed.

PR #2238, opened 2022-02-23
Update W3C links

rettinghaus:
This PR just changes `http` to `https` for links to the _W3C_ and the `Unicode Technical Site` in the Guidelines and the Specs.


sydb:
Note that the checks that have failed are failing due to #2232, not this PR.
(Testing locally now — if local checks pass and I hear no objections, will just merge.)

sydb:
Passes tests locally (in a docker container, including the “Check validity with special-purpose XSL code, looking for bad links etc” bit), except for the break caused by #2232.
So I plan to merge this shortly …

Issue #2249, opened 2022-03-23
Update data.temporal.iso


rettinghaus:
TEI should adopt the updates to ISO 8601 published in 2019. 

E.g. it eliminates the possibility to encode "24:00", which we specifically address in the remarks in `data.temporal.iso`. 

However, it needs to be discussed if this will be just a minor update, or if we should remodel it completely, as ISO 8601 now comes in two parts. 

The second part (ISO 8601-2:2019) allows e.g. ranges. So if we adopt the complete standard (as somewhat suggested by the name `data.temporal.iso`) we could boil it down to `@when-iso`, because `@notAfter-iso`, `@notBefore-iso`, `@from-iso`, and `@to-iso` can now all be written in one expression.

bansp:
With regard to the last statement, I would say, instinctively, that the changes in the ISO standard are perhaps not enough to deprecate attributes other than `@when-iso`, for two interconnected reasons:

- the Guidelines as a whole are not a reference implementation of ISO standards (although I would definitely argue that they should make it possible to easily tailor the model/schema to become such an implementation -- (i) as a matter of common sense in today's world and given ISO's significance, and also (ii) as a matter of not alienating the user base that needs a reference implementation)
- the Guidelines are by (implied?) definition a toolkit to craft your models with.

So I'd suggest making it possible for users to only keep `@when-iso` and enjoy the full benefit of ISO 8601:2019 in their customization, but also to keep the extended repertoire of attributes in the generic Guidelines, even if the attributes may seem redundant from the point of view of an ISO expert.

sydb:
Either way, yes, updating to ISO 8601:2019 is in the cards, and the preliminary step is on my to-do list. I had plans to do some significant work in this area first over winter break 2020–2021, but COVID problems hit others around me forcing me to take on other responsibilities; and then again over winter break 2021–2022, but I got quite sick myself (not COVID). I do not expect to make any progress on this for another month or so, but am very hopeful after that …

And, BTW, my thoughts are to do both — support the entirety of 8601:2019 (both parts), and still permit the old `@from-iso`, et al. Yes, that means there would be 2 ways of expressing a variety of cases. Nothing new to the TEI.

martinascholger:
Council F2F: we removed the remark; 
updating to ISO 8601:2019 is blocked for now until Coucil has access to the specification.

ebeshero:
[...]

PR #2250, opened 2022-03-23
Update ISO 8601 to specific implementation

rettinghaus:
This specifies the ISO 8601 standard named in `att.datable.iso.xml` and `teidata.temporal.iso.xml` to be the one from the standard published in 2004 (i.e. _not_ the latest standard). 

connected to #2249


bansp:
This seems to be a sensible modification, simply because it makes things more precise, by pinpointing the version of the ISO standard that the Guidelines currently use. Given what Klaus says in @2249, perhaps I should even say that "it makes things right", because if you provide a non-dated reference to an ISO standard, it is interpretable as, roughly, "all versions up till the most recent one", and as far as I understand, "24:00" was OK in the 2004 version but is not OK now.

sydb:
@bleekere and I looked at this (for the purpose of assigning to someone), and we both think it looks just fine. Thus we did not assign, we just merged.

Issue #2110, opened 2021-02-11
Ruby: Multiple ruby streams in the same orientation


martindholmes:
It is not uncommon for two distinct ruby streams to appear in the same place relative to the main text--two annotations on the right, for example, as @duncdrum suggests for this:

https://github.com/martindholmes/rubyForTEI/blob/main/Images/Bopomofo.gif

We should specify a recommended encoding strategy for this, either through the use of extra values for `@place` (right-right, right-right-right), or through the understanding that a sequence of two `<rt>` elements, both with the same value for `@place`, are assumed to appear in their document sequence, so the first of two `<rb place="right">` elements would appear immediately to the right of the ruby base, and the second would appear to the right of the first.

knagasaki:
Thanks for the good example, @duncdrum . and for the good suggestion, @martindholmes 

In the case of the nested ruby, it can be marked up like below:

```xml
      <ruby>
         <rb>
            <ruby>
               <rb>ㄅ</rb>
               <rt place="right">B</rt>
            </ruby>
         </rb>
         <rt place="right">博</rt>
      </ruby>
```

It seems to be easily understood the relationship of place of both `rt`.

In the case of non-nested, 'right-right' for `@place` is so helpful like below:

```xml
      <ruby>
         <rb><anchor xml:id="B"/>ㄅ<anchor  xml:id="owari"/></rb>
         <rt place="right" from="#B" to="#owari">B</rt>
         <rt place="right-right" from="#B" to="#owari">博</rt>
      </ruby>
```

```xml
      <ruby>
         <rb><c xml:id="B">ㄅ</c></rb>
         <rt place="right" target="#B">B</rt>
         <rt place="right-right" target="#B">博</rt>
      </ruby>
```

martindholmes:
@MegJBrown, @martinascholger, @hcayless and I have some questions on this: 

1. With nested rubys, the structure appears to us to suggest that the second rt is glossing both the rb and the rt before it. Is this true? 
2. With non-nested rubys, would it be correct to say that the rts might be columnar? In other words, might there be a "right-right" rt which lines up with other right-rights before it, but doesn't have a preceding "right" rt? 
3. Can we have a clearer explanation of what exactly is happening in the example image? What are the two rts doing? 

@knagasaki and @duncrum, what do you think?

knagasaki:
@martindholmes I'm sorry that the doubled `<rt>` in the non-nexted examples above would be my mistake. I revised both examples.

[...]

PR #2255, opened 2022-04-07
multiple ruby annotations with same alignment

martinascholger:
example added demonstrating the parallel approach, for #2110. An example for the nested approach is still needed.

### DIFF
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -1364,7 +1364,8 @@ overshadows the historical content</quote>.</egXML>
             </p>
           </egXML>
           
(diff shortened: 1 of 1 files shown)


martindholmes (commented):
There's a comma splice here:

The first of the two <gi>rt</gi> elements would appear to the right of the ruby base (<gi>rb</gi>), the second would appear...

I would suggest '..., while the second would appear'.

The xml:id="B" collides with an id in the TC chapter, so perhaps it could be made unique by adding a prefix "ancB".

martinascholger:
@raffazizzi , @martindholmes can you please have a look at this PR

martindholmes (approved):
Looks good to me.

HelenaSabel:
It looks good to me to so I proceed to merge it (without waiting for a second review)

Issue #1999, opened 2020-05-18
why is <graphic> not a member of att.typed?


sydb:
The example we have of `<graphic>` in the tagdoc is a bit, well, lame:
```xml
 <surfaceGrp n="leaf1">
  <surface>
   <graphic url="page1.png"/>
  </surface>
  <surface>
   <graphic url="page2-highRes.png"/>
   <graphic url="page2-lowRes.png"/>
  </surface>
 </surfaceGrp>
```
Seems to me one might want to differentiate high from low resolution images **without** examining the URL, as it may not contain that information, or not in an easily parsable way. (Heck, that’s what **I** want to do, which is what led me to discover there is no `@type` on `<graphic>`.) The `<graphic>` element certainly meets @jamescummings’s criteria: it is repeatable and categorizable.

If it were a member of att.typed, a contrived example might be:
```xml
<facsimile>
  <surfaceGrp n="leaf 1" xml:id="spi001">
    <surface xml:id="spi001r">
      <graphic type="normal" subtype="thumbnail" url="./rscrs/kfo49740720.jpg"/>
      <graphic type="normal" subtype="low-res" url="./rscrs/kfo49740352.jpg"/>
      <graphic type="normal" subtype="high-res" url="./rscrs/kfo49739952.jpg"/>
      <graphic type="high-contrast" subtype="low-res" url="./rscrs/kfo49739616.jpg"/>
      <graphic type="high-contrast" subtype="high-res" url="./rscrs/kfo49739488.jpg"/>
    </surface>
    <surface xml:id="spi001v">
      <graphic type="normal" subtype="thumbnail" url="./rscrs/kfo49739360.jpg"/>
      <graphic type="normal" subtype="low-res" url="./rscrs/kfo49739232.jpg"/>
      <graphic type="normal" subtype="high-res" url="./rscrs/kfo49739056.jpg"/>
      <graphic type="high-contrast" subtype="low-res" url="./rscrs/kfo49738800.jpg"/>
      <graphic type="high-contrast" subtype="high-res" url="./rscrs/kfo49738640.jpg"/>
      <zone xml:id="spi001v_detail01">
	<graphic type="normal" subtype="thumbnail" url="./rscrs/kfo46397008.jpg"/>
	<graphic type="normal" subtype="low-res" url="./rscrs/kfo46396832.jpg"/>
	<graphic type="normal" subtype="high-res" url="./rscrs/kfo46396304.jpg"/>
[...]

PR #2256, opened 2022-04-08
graphic should be a member of att.typed, #1999

martinascholger:
graphic is now a member of att.typed and an example has been added to the specs, for #1999

### DIFF
--- a/P5/Source/Specs/graphic.xml
+++ b/P5/Source/Specs/graphic.xml
@@ -30,6 +30,7 @@
     <memberOf key="att.media"/>
     <memberOf key="att.resourced"/>
     <memberOf key="att.declaring"/>
+    <memberOf key="att.typed"/>
   </classes>
   <content>
        <classRef key="model.descLike" minOccurs="0" maxOccurs="unbounded"/>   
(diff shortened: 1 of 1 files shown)


martinascholger:
@JanelleJenstad, @sydb can you please have a look at this PR

ebeshero (commented):
The example looks good. I wonder if there is a better term than “normal” for the first `@type`? Maybe “unfiltered”? I’d be happy to accept this either way. I’ll wait for @JanelleJenstad or @sydb to review before pulling this in.

sydb:
Things I have checked:

- Counted all `@type`, looking for misfits (9 normal, 6 high-contrast; total 15)
- Counted all `@subtype`, looking for misfits (6 high-res, 6 low-res, 3 thumbnail; total 15)
- Extracted all `@url` looking for misfits by eye.
- Counted each of the following strings in the values of `@url`:
  - spi: 15
  - contrast: 6
  - normal: 6
  - thumb: 3
  - [above 3 add up to 15]
  - highRes: 6
  - lowRes: 6
  - [above 2 add up to 12, the # of non-thumbs]
  - 001v: 10
  - 001r: 5
  - detail: 5
  - [above 3 add up to 15]
- Ran most of the build (make clean validate html-web test exemplars) on this branch: no sigficant errors or failures.

As for the word “normal”, it is what I have heard photographers use to refer to the images that have not had contrast boosted.

The only thing I question is the `@source` of "#UND" — is this not a made-up example? (In which case "#NONE".) I do not remember, maybe this was based on something real.

sydb (approved):
Can be merged with or without considering `@source` of `<egXML>`. (Too minor a _possible_ issue to prevent putting in this release.)

PR #2264, opened 2022-04-10
removed unnecessary constraint and amended content model of objectIdentifier for #2260

### LINKED ISSUE #2260: objectIdentifier_minimal test is both complex and unnecessary
The report test of the "objectIdentifier_minimal" constraint reads
~~~XPath
not(count(*) gt 0)
~~~
Well, since the result of a count() _cannot_ be negative (or fractional), saying “is not greater than zero” is the same as saying “is zero”, which would seem to be a lot simpler. So either
~~~xml
<sch:report test="count(*) eq 0">     OR     <sch:assert test="count(*) ne 0">
~~~
would be a better way to express this, I think. But more importantly, the content model of `<objectIdentifier>` is 
~~~
( model.placeNamePart | institution | repository | collection | idno | msName | objectName | altIdentifier | address )*
~~~
If we just change that `*` (zero or more) at the end to `+` (one or more), the entire constraint specification could be dropped, no?

ebeshero:
This is a super simple adjustment to resolve #2260, recently opened by @sydb.

### DIFF
--- a/P5/Source/Specs/objectIdentifier.xml
+++ b/P5/Source/Specs/objectIdentifier.xml
@@ -15,7 +15,7 @@ $Id$
     <memberOf key="model.biblPart"/>
   </classes>
   <content>
-    <alternate minOccurs="0" maxOccurs="unbounded">
+    <alternate minOccurs="1" maxOccurs="unbounded">
       <classRef key="model.placeNamePart"/>
       <elementRef key="institution"/>
       <elementRef key="repository"/>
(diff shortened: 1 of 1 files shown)


ebeshero:
So, @sydb opened #2260  yesterday and it seemed super easy to implement (removing an unnecessary Schematron constraint and amending the content model of `<objectIdentifier>`. I just did the work and it's passing tests. If this looks good, @raffazizzi , want to pull it in for the release?

ebeshero:
@raffazizzi Okay, maybe hold off on merging this a moment: @sydb is calling for discussion on the ticket: https://github.com/TEIC/TEI/issues/2260#issuecomment-1094492579

sydb (approved):
Having not heard any objection from @jamescummings, I think this should just be merged.

PR #2265, opened 2022-04-12
Fix some editorial errors regarding moduleRef

dmj:


sydb:
I word-smithed the message a bit, and now consider this PR ready-to-go (as it passed auto-tests and my local build, which I have to admit is problematic at the moment).
But since this PR changes not just the prose but also adds code, I think only one of the release techs (JJ, HC, RV) should actually merge.

sydb (approved):
Post word-smithing this change seems fine to me — fixing corrigible error and correcting egregious oversight.

PR #2253, opened 2022-04-04
List ref issue 2209

ebeshero:
I believe I've resolved our issues with remodeling `<listRef>` so it's now passing tests! (I needed to update a line in expected-results in the Tests on TEI repo to reflect its newly acceptable range of locations.)

I believe this is ready to merge in to resolve #2209. Can someone review this and merge it in to dev if you agree?

PR #2266, opened 2022-04-12
detest fix

sydb:
Fix the detest process per bug I introduced recently that @peterstadler [noticed](https://github.com/TEIC/TEI/pull/2253#commitcomment-71106644). This involved creating a section of detest.xml that contains ODD code so it could be tested there.


sydb:
Apologies to @peterstadler — you’re going to want to ignore whitespace when comparing.

martindholmes (approved):
Looks good to me.

Issue #2267, opened 2022-04-12
detest target needs to be more comprehensible


sydb:
The **detest:** target of Test/Makefile currently does (roughly) the following:

1. Compiles detest.odd into detest.rng, detest.isosch, detest.rng, and detest.xsl
2. Under message “Validate ODD file using Schematron”, validates detest.odd (putting output into detest1.log):
  a) against ../p5odds.rng using jing
  b) against ../p5odds.message.isosch.xsl using Saxon
3. Under message “Validate ODD file using RELAX NG schema”, validates detest.xml against detest.rnc putting output into detest2.log.
4. Under message “Validate XML file using Schematron”, validates detest.xml against detest.xsl putting output into detest3.log
5. Combines the 3 log files into one
6. Compares that one log file to the expected_results/ output.
7. Deletes the detst1, 2, & 3 log files

I propose to make this more explicit, and thus more understandable by mere mortals. E.g.:

- Step 2 should occur first. Preferably it would emit separate messages saying what it was doing for steps (a) and (b).
- The message for 3 is just plain wrong, it is the XML instance that is being validated against the compiled outputs of the ODD file, not the ODD file being validated
- Outputs should not go to temp files detest1, 2, & 3, but rather to retained files
  1. detest_odd_schematron.log
  2. detest_xml_relaxng.log
  3. detest_xml_schematron.log
- Each of these log files should be diffed separately with -C0 -bBw switches, with a message preceding

I believe that the only controversial issue in all of the above is using 3 separate log files instead of only 1. Thus I am flagging this as “Needs discussion” for that detail, and “Go” for the rest; and will do those 2 things in separate commits. (Of course, anyone can object to the rest and discuss here, or say you think the separation into 3 outputs should be Go, too. :-)

duncdrum:
Yes please to all of the above

sydb:
This can be closed now, yes?

martinascholger:
PR is merged, so we can close the issue.

PR #2268, opened 2022-04-12
tidy up detestable detest

sydb:
Address issue #2267 in 2 separate commits as 2 separate steps in case folks like most of the ideas therein but do not want to see the detest.log file divided into 3 output log files.

### Note
I do not think this PR requires two separate reviewers.

If one of the 2 requested reviewers okays it, then one of the 3 assignees needs to EITHER a) review it with an eye towards the upcoming release — will this make upcoming release that much easier (because this part of the build is less confusing) or harder (because something is different), and merge if the former; OR b) deliberately punt it to next release as too difficult to review now.


P5/Test/Makefile
```diff
@@ -107,19 +107,29 @@ detest: checkrnv
 	@echo " "
 	@echo "--------- work on target $@ ---------"
 	@echo BUILD: Test file with deliberate mistakes
+	@echo "--- run ant to validate ODD file (./detest.odd):"
+#	Note that output below is called 'detest_odd_schematron' because if there
+#	is a RelaxNG error the build dies and no output is generated
+	$(ANT) -f antruntest.xml -Doutputname=detest -DoddFile=detest.odd validateodd > detest_odd_schematron.log 2>&1
+	@echo "--- run ant to compile ODD file inoto RelaxNG and Schematron"
+	@echo "(detest.odd → detest.rng & detst.isosch):"
```
martindholmes: Shouldn't that be "detest.isosch"?

sydb: Yup.

sydb: Still working on other bugs, will fix that typo, too, in a few mins.

sydb:
OK. Took another crack at this. My apologies, but it turns out to be a pain to test, because when tested on my local system of course the paths match, so if you fail to delete them properly the result is the same!

JanelleJenstad:
Bump @HelenaSabel or @martindholmes. Martin pointed out a typo in his review. We need one of you to okay the merge.

sydb:
1. Typo @martindholmes spotted fixed in 48c7996.
2. It was just in an informative message, anyway. I.e., only those who build the GLs would ever have seen it.

martindholmes:
Tests pass and in my checking they run faster than before. Nice work!

sydb:
BTW, assignees, the thing you want to look at is the Makefile. No point in really paying attention to the log files, I don’t think. And you should not feel compelled (although you should feel free) to scrutinize every line of code. The main test is does it work when you use it to test the GLs.
The basic change is instead of validating in 3 different ways[1] generating one big log file, and then comparing it to expected, we now validate the same 3 different ways, but generate 3 separate log files (with semi-intelligent names) and compare each of them to expected individually. (This way, when there is a problem, you know which validation process it was in.) Also improved the messages you get as this happens so it is (a little) less confusing.

[1] 
1. validate detest.odd against p5odds.rng and p5odds.isosch (actually p5odds.message.isosch.xsl)
2. validate detest.xml against detest.rnc
3. validate detest.xml against detest.isosch (actually detest.xsl)

HelenaSabel (approved):
Thank you for the explanations, @sydb

I run the tests and indeed you notice the time difference!

PR #2281, opened 2022-04-18
Translation from knagasaki/TEI

knagasaki:


martindholmes (approved):
This looks fine except for the punctuation question in https://github.com/TEIC/TEI/issues/2337, which we can deal with separately. I'll merge this.

PR #2287, opened 2022-04-25
rewrite constraint "only_1_ODD_source"

peterstadler:
This is a simple rewrite of the original constraint. It avoids to match the `@source` attribute and rather match the parent element; this is to avoid those pesky schematron warnings "SXWN9000: The child axis starting at an attribute node will never select anything" when Schematron contexts are matching attributes.


sydb:
Seems fine (see review).
But why is it necessary? The occurrence of that message clearly indicates a bug in your Schematron processor, not in the schema. And I had thought we no longer had that bug in our build process. Do we?

P5/Source/Specs/att.global.source.xml
```diff
@@ -28,20 +28,20 @@ $Id$
...
-                              | parent::tei:elementRef
-                              | parent::tei:macroRef
-                              | parent::tei:moduleRef
-                              | parent::tei:schemaSpec )
-                              and
-                              $srcs[2]">
+          <sch:rule context="tei:*[@source]">
+            <sch:let name="srcs" value="tokenize( normalize-space(@source),'&#x20;')"/>
+            <sch:report test="( self::tei:classRef
+              | self::tei:dataRef
```
sydb, line 28: What? An unnecessary change in whitespace (that makes code less readable!)?

P5/Source/Specs/att.global.source.xml
```diff
@@ -28,20 +28,20 @@ $Id$
...
+              | self::tei:moduleRef
+              | self::tei:schemaSpec )
+              and
+              $srcs[2]">
               When used on a schema description element (like
-              <sch:value-of select="name(..)"/>), the @source attribute
+              <sch:value-of select="name(.)"/>), the @source attribute
               should have only 1 value. (This one has <sch:value-of
-              select="count($srcs)"/>.)
+                select="count($srcs)"/>.)
```
sydb, line 38: Ditto. (But does not really affect readability, and, depending on your aesthetics, might even look better.)

sydb (approved):
Go for it, Mr. Asignee.

sydb (approved):
Not sure why Mr. Travis can’t find docker image, but I do not think it has anything to do with this PR, and would not worry about it.

P5/Source/Specs/att.global.source.xml
peterstadler: Sorry, I should have flagged this PR with a warning about whitespace changes ;)

peterstadler:
> Not sure why Mr. Travis can’t find docker image, but I do not think it has anything to do with this PR, and would not worry about it.

Yes, I think GitHub had some issues yesterday. Will try to rerun those tests …

peterstadler:
@raffazizzi what do you think? Ready to merge?

PR #2293, opened 2022-05-04
Add unabbreviated name

rettinghaus:
Most abbreviated elements and attributes show their unabbreviated form in parentheses, `@decls` didn't. I suspect it stands for "declarations", so I added this one.


peterstadler (changes requested):
Thanks for spotting! 
Technically this expanded form should go into a dedicated `<gloss>` element (without parentheses, but with corresponding `@xml:lang` and `@versionDate` attributes), not into the `<desc>`.

rettinghaus:
Thanks for the hint! Changed it accordingly.

peterstadler:
Help needed here from @TEIC/tei-technical-council : Is "declarations" the proper gloss for "decls"?

lb42:
yes

peterstadler (approved):
No objections anymore :) 
Thanks @lb42 for confirming and @rettinghaus for the patience!

PR #2302, opened 2022-06-09
add missing glossary to grouping elements

rettinghaus:
This PR adds some small missing gloss entries.


sabineseifert:
Thank you for the PR.

PR #2305, opened 2022-06-14
spanish translations

susannalles:
There are some terms where I am not sure at all (branch, fork, ...) We could discuss it in the Editaton.

### DIFF
--- a/I18N/spec_translator.xml
+++ b/I18N/spec_translator.xml
@@ -193,7 +193,7 @@
              </item>
              <item>
                <label>saving</label>
-               <gloss xml:lang="en" versionDate="2022-06-08">Guardando...</gloss>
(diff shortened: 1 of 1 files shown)


hcayless (changes requested):
On line 196, change 'en' to 'es'. Otherwise, it looks great! I too am unsure about 'branch'. Since it's the name of a Git concept and command, maybe it would be better left as-is, like 'fork'.

susannalles:
Ok @hcayless I think I changed it. I agree, perhaps 'branch' should remain as-is. I used 'rama' cause I saw some websites about github using it.

hcayless:
My rule of thumb about this stuff is "Could I Google and find sensible results using this word or phrase?" If 'rama' is actually used in Spanish Git documentation, then it's fine. 'Fork' is a more specialized, GitHub-only thing, so they don't necessarily need to be treated the same way.

PR #2308, opened 2022-06-16
add missing glossary to description elements

rettinghaus:
This PR adds some small missing gloss entries. Similar to #2302.


sabineseifert:
Thank you for the PR.

PR #2310, opened 2022-06-16
Translation from KharolusIII/TEI

KharolusIII:


hcayless:
This is from a workshop, and there was a bug that caused a weird save. I'll handle fixing it.

hcayless:
Merged via command line.

PR #2312, opened 2022-06-16
Translation from dh-miami/TEI

dh-miami:


hcayless:
This is from a workshop, and there was a bug that caused a weird save. I'll handle fixing it.

hcayless:
Merged manually.

PR #2325, opened 2022-06-30
fixed Schematron check for unused <rendition> definitions

rvdb:


peterstadler (approved):
If I read it correctly, the issue is that `key('idrefs', @xml:id)` may return a sequence of attribute nodes. Now it checks for _some_ `@rendition` attribute (NB not _every_) with the appropriate value to point at the current `rendition` element.
This looks like a valid fix to me.

rvdb:
@peterstadler Indeed, thanks for elaborating. The test is sufficient: a `<rendition>` is in use once it's being referred to from at least one `@rendition` attribute. I agree the expression might be a bit terse; this version would probably be more transparent:

    some $i in key('idrefs', @xml:id) satisfies $i instance of attribute(rendition)
    
I could change it, but then again there will be lots of such implicit expressions in the file (whose implicitness I'm probably not even aware of) that might be candidates for more transparency.

peterstadler:
Hi @rvdb , thanks for confirming! From my side, no change is needed—I was just spelling it out for documentary reasons :)
Will merge now!

PR #2333, opened 2022-09-09
Little fixes for #2326:

### LINKED ISSUE #2326: constraints on `<dataFacet>` and `@restriction` should be explicit
Should have Schematron to enforce `<dataFacet>` or `@restriction` can occur iff `@name`, as the remarks in the [dataRef tagdoc](https://tei-c.org/Vault/P5/4.4.0/doc/tei-p5-doc/en/html/ref-dataRef.html) make clear.

(Note that #1744 may have some bearing on this, but it will take a long time, and thus should not stop us from an explicit constraint on the current system.)

sydb:
* Make all constraint spec tests nonfatal
 * Combine the two constraint specificaions with the same context into one
 * Re-word so that messages are parallel
 * Also re-word remarks to be clearer

### DIFF
--- a/P5/Source/Specs/dataRef.xml
+++ b/P5/Source/Specs/dataRef.xml
@@ -22,21 +22,16 @@
     <constraint>
       <sch:rule context="tei:dataRef[tei:dataFacet]">
         <sch:assert test="@name" role="nonfatal">Data facets can only be specified for references to datatypes specified by
-          XML Schemas: Part 2: Datatypes</sch:assert>
-      </sch:rule>
-    </constraint>
-  </constraintSpec>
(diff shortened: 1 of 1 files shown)


peterstadler:
looks like a no-brainer to me. @sydb: Is there more to #2326 or will this PR address all issues from this ticket?

sydb:
I think we could probably come up with better wording, and others might handle the actual testing better (or at least differently). But this is such an improvement over what we have in dev now, we could merge and close and feel we had done a good job.

peterstadler:
> But this is such an improvement over what we have in dev now, we could merge and close and feel we had done a good job.

Sounds good to me, will do :)

Issue #1079, opened 2012-09-06
reconsider membership or use of model.glossLike


TEITechnicalCouncil:
In implementing bug 2812295 ( http://purl.org/tei/bug/2812295 ) in spring 2010, we made &lt;graphic&gt; a non-empty element that can contain model.glossLike*.  But this had what I suspect was an unintended consequence: permitting two members of the tagdocs module -- &lt;altIdent&gt; and &lt;equiv&gt; -- as children of &lt;graphic&gt; (and other members of model.glossLike).

Sebastian said he thinks we put them in model.glossLike when that model class was used mainly by ODD elements and then forgot about them when glossLike was more widely adopted.

We should revisit either the way this was implemented or take a look at the membership of model.glossLike.

Sebastian suggested moving &lt;altIdent&gt; and &lt;equiv&gt; to a new model class (perhaps called "model.equivLike"), which would be used in classSpec, constraintSpec, elementSpec, macroSpec, and valItem.

Original comment by: @kshawkin

PR #2334, opened 2022-09-09
sydb 2049 — on <altIdent> reduex

sydb:
Note: I had an off-by-one error when transcribing some of the [step numbers](https://github.com/TEIC/TEI/issues/2049#issuecomment-1092296175) into the git commit comments. Sorry.

Also note that the content of `<altIdent>` is defined as `name=NCName` rather than `key=teidata.xmlName` due to TEIC/Stylesheets/issues/569.


hcayless:
So...having both a model.indentInfo and a model.identInformation makes me want to run away screaming. Why?

sydb:
I have no idea — I do not even know if you are unhappy with the _names_ (in which case feel free to suggest different ones), or the idea that we need 2 classes.

sydb:
But in any case it is not properly passing tests, yet. Still working …

sydb:
Fixed. (All tests pass in docker container on my system.)

ebeshero:
@sydb I think I'm with @hcayless : I don't like there being two models with such similar names: model.identInfo and model.identInformation, but I'm sure there's a good reason why you introduced them. The naming seems a headache for maintaining later, isn't it? The only distinction I see is quite subtle:

* model.identInformation: `<desc versionDate="2022-09-09" xml:lang="en">groups elements which provide **an alternative name, explanation, or an equivalency** for a markup construct.</desc>`. Those include:
    * `<altIdent>`
    * `<equiv>`
    * `<gloss>`
    
* model.identInfo is just a member of model.identInformation: `<desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an explanation or an equivalency for a markup construct.</desc>`. Those include
    * `<equiv>`
    * `<gloss>`

What's the need for model.identInfo to be defined this way? 

At the very least, on the new model specs, we'd better remove the dated `<desc>` elements from other languages that go back to Spec for model.glossLike, moved to defunct here.

P5/Source/Specs/model.identInfo.xml
```diff
@@ -0,0 +1,24 @@
...
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" xml:id="IDENTINFO" type="model" ident="model.identInfo">
+  <desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an explanation or an equivalency for a markup construct.</desc>
+  <desc versionDate="2007-12-20" xml:lang="ko">마크업 구성에 대한 대체명, 설명 또는 기술을 제시하는 요소를 모아 놓는다.</desc>
```
ebeshero: @sydb  Delete lines 12-20 (translations from the old model.glossLike).

P5/Source/Specs/model.identInfo.xml
```diff
@@ -0,0 +1,24 @@
...
+<!--
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" xml:id="IDENTINFO" type="model" ident="model.identInfo">
+  <desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an explanation or an equivalency for a markup construct.</desc>
```
ebeshero: @sydb I'm not sure I understand why we need this model.identInfo. The name is very easy to confuse with the larger model of which this is a subset, model.identInformation. What about model.identEquiv instead?

P5/Source/Specs/model.identInformation.xml
```diff
@@ -0,0 +1,21 @@
...
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" xml:id="IDENTINFORMATION" type="model" ident="model.identInformation">
+  <desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an alternative name, explanation, or an equivalency for a markup construct.</desc>
+  <desc versionDate="2007-12-20" xml:lang="ko">마크업 구성에 대한 대체명, 설명 또는 기술을 제시하는 요소를 모아 놓는다.</desc>
```
ebeshero: @sydb Delete lines 12-20 (translations from the old model.glossLike).

ebeshero (changes requested):
@sydb I think I'm with @hcayless : I don't like there being two models with such similar names: model.identInfo and model.identInformation, but I'm sure there's a good reason why you introduced them. The naming seems a headache for maintaining and remembering the difference between the two. The only distinction I see is quite subtle:

* model.identInformation: `<desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an alternative name, explanation, or an equivalency for a markup construct.</desc>`. Those include:
    * `<altIdent>`
    * `<equiv>`
    * `<gloss>`
    
* model.identInfo is just a member of model.identInformation: `<desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an explanation or an equivalency for a markup construct.</desc>`. Those include
    * `<equiv>`
    * `<gloss>`

What's the need for model.identInfo to be defined this way? 
Could it be renamed to model.identEquiv or model.identGloss or something that's not to do with alternate naming?  

At the very least, on the new model specs, we'd better remove the dated `<desc>` elements from other languages that go back to Spec for model.glossLike, moved to defunct here.

P5/Source/Specs/model.identInformation.xml
```diff
@@ -0,0 +1,21 @@
...
+<!--
+Copyright TEI Consortium. 
+Dual-licensed under CC-by and BSD2 licences 
+See the file COPYING.txt for details
+$Date$
+$Id$
+-->
+<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tei" xml:id="IDENTINFORMATION" type="model" ident="model.identInformation">
+  <desc versionDate="2022-09-09" xml:lang="en">groups elements which provide an alternative name, explanation, or an equivalency for a markup construct.</desc>
```
ebeshero: @sydb Here, it seems as if the qualifier "alternative" might apply to the words "explanation" and "equivalency" just due to its placement at the head of a list. How about rephrasing this to change the order for clarity? 

"groups elements which provide an explanation, equivalency, or an alternative name for a markup construct."

sydb: Could also just reword to “… an alternative name, an explanation, or an equivalency …”, which is what I had intended. Thoughts?

P5/Source/Specs/model.identInfo.xml
sydb: Done.

P5/Source/Specs/model.identInformation.xml
ebeshero: @sydb Which version is clearer on translation? I think maybe my version, since it tucks "alternative" away at the end of the sequence and doesn't depend on English articles?

P5/Source/Specs/model.identInfo.xml
sydb: Well we certainly need 2 model classes, here. One is used by 6 elements, the other by 7 elements:
~~~
model.identInfo ↢ model modelGrp modelSequence moduleSpec paramSpec schemaSpec
model.identInformation ↢ attDef classSpec constraintSpec dataSpec elementSpec macroSpec valItem
~~~
But I could easily be convinced those are not particularly good names for these classes. I am convinced they are better than model.glossLike. But I am not remotely convinced we can’t do better. They are really just what jumped to mind.
I am not particularly fond of model.identEquiv, but I guess if you think of `<equiv>` as a formal XSLT-processable equivalent, and `<gloss>` as the natural language equivalent it is workable. Would you keep the larger class (`<gloss>`, `<equiv>`, `<altIdent>`) as model.identInformation, or change it to model.identEquivOrAlternate or something else?

P5/Source/Specs/model.identInformation.xml
sydb: Why? I can’t really read any of them, but from my spattering of Spanish it seems to be pretty close to spot on.

P5/Source/Specs/model.identInformation.xml
sydb: Good point. Will re-word in a few mins …

P5/Source/Specs/model.identInfo.xml
ebeshero: I think the larger class could be simplified as model.identInfo, if we make the smaller class model.identEquiv.

P5/Source/Specs/model.identInformation.xml
ebeshero: @sydb I had not checked the old desc for model.glossLike when I reviewed this, but I see you aren't making substantial changes to that desc. If you are, it's just reorganizing the list for clarity. Maybe these should be preserved for the time being. 

I see you've removed the  translations for the *smaller* model--all good!

ebeshero (dismissed):

lb42:
Howsabout calling it model.identSynonyms for clarity?

sydb:
@lb42 — Which one, the one that has `<gloss>`, `<equiv>`, _and_ `<altIdent>`, or the one that has only `<gloss>` and `<equiv>`?

Either way, much better than my model.identInfo[rmation], and perhaps even better than the model.identEquiv, thank you.

lb42:
I meant the one that has altIdent but lookibg closer at this thread i dont see much justification for havibg two classes anyway

sydb:
Would you use one class and add `<altIdent>` manually to the 7 elements that need it in their content model?
(Remember, the _point_ of the PR is to get `<altIdent>` **out** of the content models of the other 6 elements.)
In any case, I like model.identSynonyms much better for the one with `<altIdent>` (currently, and kinda uselessly, called model.identInfo in this branch). Unless someone objects or comes up with something better, I plan to change it to that name later today.

lb42:
So the issue is whether otslikely there might ever be a sevond such element which shd appear in the same places as altIdent. I think probly not.

sydb:
OK, then, it looks like we have 2 suggestions on the table; I will also add the no-class approach, the reverse approach, and the no-subclass approach for completeness.

### proposal one (classes all the way down)
~~~
model.identEquiv = gloss | equiv   [used by 6 elements]
model.identSynonyms = altIdent | model.identEquiv   [used by 7 elements]
~~~

### proposal two (one class +altIdent)
~~~
model.identEquiv = gloss | equiv   [used by 6 elements]
( model.identEquiv | altIdent )   [used by 7 elements]
~~~

### proposal three (no classes approach)
~~~
( gloss | equiv )   [used by 6 elements]
( gloss | equiv | altIdent )   [used by 7 elements]
~~~

### proposal four (no sublcass approach)
~~~
model.identEquiv = gloss | equiv   [used by 6 elements]
model.identSynonyms = gloss | equiv | altIdent   [used by 7 elements]
~~~

### proposal five (reverse of proposal two, sort of)
~~~
( gloss | equiv )   [used by 6 elements]
model.identSynonyms = gloss | equiv | altIdent   [used by 7 elements]
~~~

I personally like proposals one and four (i.e., defining 2 classes), but honestly would be happy with any of the above. I think @lb42’s observation that one of the motivating factors for using classes — to allow users to easily add a new element — is not really apropos (because not likely to be used by anyone) is accurate.

ebeshero:
@sydb I like your proposal one: it seems easiest to work with what you've already prepared, considering we're basically dithering over the model names.

lb42:
The trouble with the TEI class system is that some classes are "structural" (i.e. they provide a convenient means of grouping elements likely to appear together in a content model) and others are "semantic" (i.e. they group elements which have some semantic property in common. indicated by the name of the class). Usually, one implies the other, but in this case it doesn't seem to me that a "gloss" is a synonym to the extent that  an altIdent is: even less so for equiv. This maybe is lily-gilding, but  I'd rather see a  class named model.identLike containing  gloss and equiv, used in content models. I also think altIdent is enough of a special case that it should be explicitly referenced in content models where it's allowed, but if you want to make a special class for it model.identSynonym seems right. Though I will give five shillings (old money) to the first person who actually finds  a use for such a class in the wild.

ebeshero:
I admit was a little bothered by identSynonyms being inclusive of `<gloss>`, too, and was sort of stretching out the idea of “synonymous” to make it work conceptually as the larger model. 
 
We seem to need 
* a grouping of gloss and equiv, exclusive of altIdent.
* a grouping of all three together


We should have a name that works for whole group. We probably don’t need a class just for altIdent that says, “I’m a synonym.”

How about:

* identInfo or identLike (perhaps works for the more inclusive class)
* identEquiv for just gloss and equiv

?

bansp:
Let me just register my amazement that `equiv` is in a single class with `gloss` -- I've just looked at the [examples of `gloss`](https://tei-c.org/release/doc/tei-p5-doc/en/html/examples-gloss.html) and find it hard to imagine that `equiv` would have anything to do with that, even potentially.
I've gone through the discussion out of curiosity (Lou's remarks often make me do that), and I kept thinking that, at least from the semantic point of view, I could put `equiv` and `altIdent` together, but `gloss`? And then I looked at examples of `gloss` and started wondering about the structural paring of `gloss` and `equiv`, and that seems strange as well. (Of course, it's good to have them under `category` but I never thought that they came from a single element class). 
This is not meant to derail the current process, but I thought I'd just register my impressions in case an outsider view could help in any way.

sydb:
So @lb42 is in favor of **proposal two** with a different name for the class. @ebeshero is in favor of **proposal one** or **four** (can’t tell which) with different names. Seems like @bansp would prefer **proposal three** or a new arrangement that grouped only `<equiv>` and `<altIdent>` in a class.

As I said, I do not care much, and consider this decision pretty low impact on the TEI world. At the moment I am leaning towards **proposal two** on the theory that we only have to dream up 1 class name for it. As for that name, I shy away from @lb42 ’s proposed “model.identLike” due to the implication that `<ident>` would be a member. It is not a given that all *Like classes include the element that matches their name, but the vast majority (43 of 53) do.

## “*Like” class memebership
Using current version of this branch.

### does NOT contain its own so-named element

* model.certLike
  * `<certainty>`, `<precision>`, `<respons>`
* model.dimLike
  * `<depth>`, `<height>`, `<width>`
* model.morphLike
  * `<case>`, `<gen>`, `<gram>`, `<iType>`, `<mood>`, `<number>`, `<per>`, `<tns>`
* model.msQuoteLike
  * `<colophon>`, `<explicit>`, `<finalRubric>`, `<incipit>`, `<rubric>`, `<title>`
* model.nameLike.agent
  * `<name>`, `<orgName>`, `<persName>`
* model.persStateLike
  * `<affiliation>`, `<age>`, `<education>`, `<faith>`, `<floruit>`, `<gender>`, `<langKnowledge>`, `<nationality>`, `<occupation>`, `<persName>`, `<persona>`, `<persPronouns>`, `<residence>`, `<sex>`, `<socecStatus>`, `<state>`, `<trait>`
* model.placeStateLike
  * `<climate>`, `<location>`, `<model.placeNamePart>`, `<population>`, `<state>`, `<terrain>`, `<trait>`
* model.pLike.front
  * `<argument>`, `<byline>`, `<dateline>`, `<docAuthor>`, `<docDate>`, `<docEdition>`, `<docImprint>`, `<docTitle>`, `<epigraph>`, `<head>`, `<titlePart>`
* model.ptrLike.form
  * `<oRef>`, `<pRef>`
* model.respLike
  * `<author>`, `<editor>`, `<funder>`, `<meeting>`, `<principal>`, `<respStmt>`, `<sponsor>`

### only 1 member, named the same

* `<application>` is a member of model.applicationLike
* `<desc>` is a member of model.descLike
* `<div1>` is a member of model.div1Like
* `<div2>` is a member of model.div2Like
* `<div3>` is a member of model.div3Like
* `<div4>` is a member of model.div4Like
* `<div5>` is a member of model.div5Like
* `<div6>` is a member of model.div6Like
* `<div7>` is a member of model.div7Like
* `<divGen>` is a member of model.divGenLike
* `<div>` is a member of model.divLike
* `<g>` is a member of model.gLike
* `<head>` is a member of model.headLike
* `<l>` is a member of model.lLike
* `<state>` is a member of model.orgStateLike
* `<place>` is a member of model.placeLike
* `<rdg>` is a member of model.rdgLike

### contains its own so-named element

* model.egLike
  * `<eg>`, `<egXML>`
* model.divLike
  * `<div1–7>`, `<div>`, `<lg>`
* model.addressLike
  * `<address>`, `<affiliation>`, `<email>`
* model.annotationLike
  * `<annotationBlock>`, `<annotation>`, `<note>`
* model.biblLike
  * `<biblFull>`, `<bibl>`, `<biblStruct>`, `<listBibl>`, `<msDesc>`
* model.dateLike
  * `<date>`, `<time>`
* model.emphLike
  * `<code>`, `<distinct>`, `<emph>`, `<foreign>`, `<gloss>`, `<ident>`, `<mentioned>`, `<soCalled>`, `<term>`, `<title>`
* model.entryLike
  * `<entryFree>`, `<entry>`, `<superEntry>`
* model.eventLike
  * `<event>`, `<listEvent>`
* model.graphicLike
  * `<binaryObject>`, `<formula>`, `<graphic>`, `<media>`
* model.hiLike
  * `<hi>`, `<q>`
* model.labelLike
  * `<desc>`, `<label>`
* model.listLike
  * `<listApp>`, `<listEvent>`, `<list>`, `<listNym>`, `<listObject>`, `<listOrg>`, `<listPerson>`, `<listPlace>`, `<listRelation>`, `<listWit>`, `<table>`
* model.measureLike
  * `<depth>`, `<dim>`, `<geo>`, `<height>`, `<measureGrp>`, `<measure>`, `<num>`, `<unit>`, `<width>`
* model.milestoneLike
  * `<anchor>`, `<cb>`, `<fw>`, `<gb>`, `<lb>`, `<milestone>`, `<pb>`
* model.noteLike
  * `<noteGrp>`, `<note>`
* model.nameLike
  * `<idno>`, `<lang>`, `<model.nameLike.agent>`, `<model.offsetLike>`, `<model.persNamePart>`, `<model.placeStateLike>`, `<objectName>`, `<rs>`
* model.objectLike
  * `<listObject>`, `<object>`
* model.offsetLike
  * `<geogFeat>`, `<offset>`
* model.personLike
  * `<org>`, `<personGrp>`, `<person>`
* model.pLike
  * `<ab>`, `<p>`
* model.ptrLike
  * `<listRef>`, `<ptr>`, `<ref>`
* model.quoteLike
  * `<cit>`, `<quote>`
* model.segLike
  * `<c>`, `<cl>`, `<m>`, `<pc>`, `<phr>`, `<seg>`, `<s>`, `<w>`
* model.specDescLike
  * `<specDesc>`, `<specList>`
* model.stageLike
  * `<camera>`, `<caption>`, `<move>`, `<sound>`, `<stage>`, `<tech>`, `<view>`

bansp:
I'm not sure if label myself as "option 3". Let me explain where I come from to this.
Part of the reason I spoke up is `equiv`, which had recently made me go back all the way to 2012 for archived Council discussions on the interplay between `equiv` and `att.datcat`, also in view of what I would like to propose in the 'datcat' PR. In short, they overlap, or they are in what structuralism would call complementary distribution, with the same function (from my perspective) which is basically to provide an arc to a node, and to label that arc by implication from the element/attribute name. From that perspective, I would group `equiv` and `altIdent` because the latter performs roughly a very similar ontological (or ontology-related) function, except while `equiv` points at an object, `altIdent` points at a literal, a label.
But `gloss` is just a piece of annotation (in the sense of `a:` in RNG), so that's why I felt uneasy seeing it grouped with one of the "arc-like" elements, but not another. I could more or less clearly imagine the pragmatism of grouping `gloss` and `equiv` in tagdocs and in taxonomies (they are just handy to have around, together). But when I looked at those examples of the use of `gloss` I referenced above (and it's sometimes used as an empty element! oops :-)), I started doubting the pragmatism of grouping `gloss` and `equiv` -- the fact that they co-occur in `elementSpec` (etc.) started looking like a plain coincidence rather than any structural regularity (= regularity which one would expect to have a structural explanation).
So since this seems to be in a bit of a flux, I thought I would share my uneasiness, in case it could be useful :-)

martindholmes (approved):
I've checked out this PR and built the Guidelines without problems, and I see no issues with the changes at all.

hcayless:
I agree with @bansp.

ebeshero:
I like how @bansp points out the strangeness of  the grouping of equiv and gloss. It seems surely a matter of convenience for processing that we classify this way. I wish I knew what to recommend. If the only goal is to successfully build the Guidelines, the classes applied shouldn’t matter. They’re just a convenient cog in a machine that we keep tinkering with and repairing. If they are more than that, they’re for humans to read—but maybe these classes don’t need to be any more meaningful than altIdentClass and altIdentSubclass. Or maybe we don’t need them at all. I wonder if processing requirements might somehow be made their own meaningful classes so we aren’t straining words like synonym or equivalent or annotation when we really mean, process exclusively or inclusively.

jamescummings:
Hi all,

Coming to this very late and reading through the issue when I got to @sydb's proposals, I instantly thought Proposal 1 was the right TEI'ish way to do things. It seemed a good balance between leveraging the class system and clarity about what was going on. But then I too wondered at what `<equiv>` and `<gloss>` were doing in the same class together. So I went and read some of our history in this area, particularly https://github.com/TEIC/TEI/issues/1079 were one @lb42 notes the 'gloss-like' qualities of `<equiv>` and `<altIdent>` but not `<gloss>`. On reflection I think separating `<gloss>` from a class which has `<equiv>` and `<altIdent>` (if and only if useful... I'm sceptical that there will be much need to class-based modifications or new membership there) would make the most sense.

ebeshero:
I think the changes you made answer my questions, so I'm happy with this. 
Can someone test this locally?

HelenaSabel:
I tested it locally and the tests passed, so I think @sabineseifert  can click the green button safely :)

Issue #2340, opened 2022-09-13
Inconsistent ISO referencing: Two different entries for ISO 690 in BIB


ebeshero:
Found by @martindholmes  while investigating https://github.com/TEIC/TEI/issues/2331.
We need to decide which one of these to remove.

rettinghaus:
Does one of them have to go? One references an example in the document, the other one the whole document. 
I'm wondering more if the references to ISO documents shouldn't be updated in general, because those in there have been revised and withdrawn.

ISO 690:1987 -> [ISO 690:2021](https://www.iso.org/standard/72642.html)
ISO 12620:2009 -> [ISO 24611:2012](https://www.iso.org/standard/51934.html)
ISO 19757-3:2006 -> [ISO/IEC 19757-3:2020](https://www.iso.org/standard/74515.html)

Furthermore there are some ISO specs cited in the Guidelines but no reference in BIB.

ebeshero:
@rettinghaus Thanks for your advice and especially for [your PR](https://github.com/TEIC/TEI/pull/2344), which we've now merged and updated. I'm taking a slow "walk" through the ISO entries in the BIB to scan for inconsistencies and make some repairs as called for here. I notice that sometimes we link to ISO specs and sometimes we don't bother. I wonder which is the better strategy, but lean toward linking since eventually we *should* check to see if something was withdrawn and outdated.

ebeshero:
I'm also grepping for ISO mentions in Guidelines that lack BIB citations...will update as I locate them.

ebeshero:
After some grepping around the Specs and Guidelines, I organized a table to document which ISO specs we reference in the TEI Specs and Guidelines. Most currently don't have BIB entries. Sometimes we reference a very particular dated spec, and I'm not sure whether it's appropriate to update it. (And sometimes it's pretty clear we're talking about a very early spec and should cite it in particular.) 


| ISO Spec      | Where mentioned?               | BIB citation?      | Reasonably current ref? | 
| ---------     | --------------                 | -------------      | -------------      |
| ISO 690       | HD                             | Y                 | https://www.iso.org/standard/72642.html  |
| ISO 690:1987  | CO                             | Y (:1987)           | keep 1987 because an example is drawn from it |     
| ISO 24611     | AI, standOff              |  N        | (:2012) https://www.iso.org/standard/51934.html |
| ISO 24611:2012  | att.linguistic   |    N      |  (:2012) https://www.iso.org/standard/51934.html |
[...]

PR #2344, opened 2022-09-14
Add missing ISO references in BIB

rettinghaus:
This PR adds missing entries in the bibliography for cited ISO standards in the Guidelines. 
Also it removes an outdated link in `personGrp.xml`.
Relates to [#2340](https://github.com/TEIC/TEI/issues/2340#issuecomment-1246540706)

### DIFF
--- a/P5/Source/Guidelines/en/BIB-Bibliography.xml
+++ b/P5/Source/Guidelines/en/BIB-Bibliography.xml
@@ -3341,6 +3341,25 @@ $Id$
           </imprint>
         </monogr>
       </biblStruct>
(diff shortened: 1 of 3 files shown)


sydb (approved):
Just spoke to @ebeshero, who agrees with me: These changes are fine to pull in right now, but slightly sad because the work on 5218 will be wasted, as we are about to remove all remaining references to it!
Bad timing … if we had gotten to removing the last vestiges of 5218 a few days earlier we would have saved you some time, @rettinghaus! Nonetheless, thank you **very** much for this, the other tweaks are excellent.

PR #2348, opened 2022-09-20
Add Japanese translation for gender and sex elements

knagasaki:


peterstadler (approved):
Can only approve formally, but this looks ok to me

Issue #1929, opened 2019-09-20
What is a paragraph, really?


dariok:
Discussing between ourselves #1856 – more precisely, we wanted to see what the “semantic baggage” of `p` actually is –, @ebeshero and I stumbled across the prose definition of “paragraph” in the GL because we wanted to see what the actual difference between `ab` and `p` really is (being the radical *advocatus diaboli* I usually am in these discussions, I proposed we should get rid of `ab`).

The definition of `p` starts:
> The paragraph is the fundamental organizational unit for all prose texts, being the smallest regular unit into which prose can be divided.

While you could argue that this smallest unit may actually be the sentence, this definition already poses problems. Either
- headings, lists, etc. are not prose, or
- they are actually paragraphs (so we ditch `head` for `p type="heading"`...), or
- they are to be considered sth. that is below the chunk level and hence should be a child of a paragraph (which then makes `p` the only possible child of `div` – or rather `p` is the smallest `div` and the only difference is that i may contain text).

We felt rather uneasy about that.

So the problem is: How to spot and delimit a paragraph? Where does it start and end?
The GL state that
> Since paragraphs are usually explicitly marked in Western texts, typically by indentation, the application of the p tag usually presents few problems.

Well...
We admit that for non-Western texts this will be a problem. Actually, pre-modern Western texts have the same set of problems, be that in classical or medieval manuscripts or 16th century prints. Often, a lot of the text is one blob spanning many different pages and the idea of a subdivision is merely hinted at by, e.g., a ⸿ – but then again what is delimited by these would as often as not rather be a sub-chunk level division.
But even in modern texts the idea of delimiting paragraphs by purely typographical features such as indentation will break down: consider a novel with a lot of (unprefixed) direct speech. Is every utterance a paragraph in its own right? It is, after all, typographically distinct...

---

Far from being perfect, we could describe a paragraph as

- a collection of one ore more phrase-like bits of text (e.g., general purpose “sentences” (whatever that might be) or more specific descriptions),
- that may be considered as sharing a common (more or less complex) statement,
- that can (but need not) be typographically separated from each other (esp., they may actually comprise e.g. `q`s that all start on their own line),
- is a building block for larger, abstract dissections of a larger text (thus being one possible content for `div`, which, while possibly being explicitly marked by several levels of numbering in a heading, itself is basically an abstract segmentation of a larger whole)
[...]

PR #2352, opened 2022-09-26
Ab in ab issue #1856

### LINKED ISSUE #1856: `<ab>` should be able to nest
The `<ab>` element "(anonymous block) contains any arbitrary component-level unit of text, acting as an anonymous container for phrase or inter level elements analogous to, but without the semantic baggage of, a paragraph." In other words, it's a block-level element with no semantics. 

Many block-level elements (such as `<div>`) can nest, and this makes perfect sense; chapters contain sections contain subsections. One that cannot nest is `<p>`, and this is because of its semantics; we claim that the thing called a paragraph is a specific kind of thing that cannot directly contain other instances of itself.

`<ab>` is not a paragraph; if you're tagging a paragraph, you should use `<p>`. `<ab>` explicitly has no semantics other than its block-ness. So why can't it nest? Its analogue at the inline level, `<seg>`, can nest. Imagine you're tagging a text consisting of a sequence of emojis that you don't understand, but they're clearly arranged in nested blocks, which you can tell because of indenting. You can't use `<div>` because that's making semantic claims about the nature of the blocks; you would surely want to use `<ab>`, and have it nest.

This ticket arises out of a Stylesheets ticket (https://github.com/TEIC/Stylesheets/issues/328) where we're trying to make the TEI Lite output of odd2lite.xsl actually valid. That output contains lots of nested `<ab>`s; rather than try to rewrite the process to avoid `<ab>`s, which would be quite difficult, it seems to me that we could solve that particular problem by letting `<ab>`s nest, which they should be able to do anyway.

ebeshero:
I added a couple of commits to improve the paragraph examples in CO today, and I believe that responds to all of Council's requests from August. Is this ready to merge to dev for the October release and close #1856 and #1929?


ebeshero:
@sydb @hcayless I'm adding you as reviewers just to make sure all is well, as I kind of went ahead and opened the PR before you answered me in Slack. I am pretty sure we've made the prose and example revisions we discussed.

ebeshero:
@JanelleJenstad @martinascholger I revised this to supply a short passage from _Jane Eyre_ for the introductory example of a paragraph, mostly because we didn't have an example from the Bronte sisters in the Guidelines and because I thought the paragraph might be interesting (people who have read the novle might be surprised that this paragraph comes from there). Anyway, most importantly it's clearly a paragraph this time. As we're going into refrigeration this week, I hope we can review the changes in this branch and see if it's ready for the Guidelines release.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -107,24 +108,17 @@ the TEI document type definition.</p>
     paragraph or between paragraphs, and include bibliographic citations, notes, and 
     lists. The <gi>ab</gi> (anonymous block) element may be used as an alternative to 
     <gi>p</gi> in some kinds of texts. <gi>ab</gi> may be understood as a chunk-level 
-    element which can contain the same elements as <gi>p</gi>, but may represent a 
+    element which can contain the same elements as <gi>p</gi>, but may represent 
     less-constrained textual structures. <gi>ab</gi> may contain other <gi>ab</gi>s, 
     may use the @type attribute, and does not necessarily represent a coherent set of
```
martinascholger: This should be <att>type</att>

martinascholger (changes requested):

martinascholger:
The `<desc>` is missing in `macro.abContent`. I'm not able to leave the comment directly in the code ...

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -85,26 +85,39 @@ organization of the module described here. This should be read in
...
+    element in HTML, which is primarily a block of text, and from the <gi>ab</gi> 
+    (anonymous block) element described in <ptr target="#SASE"/>) which may be used 
+    as an alternative to the paragraph in cases that require a chunk-level container 
+    which is not necessarily a sense unit and which may have different structural properties.</p>
+  <p>The paragraph can be understood in the context of the distinct forms of textual 
+    division discussed in <ptr target="#STBTC"/>: <term>chunk</term>, 
+    <term>phrase</term>, and <term>inter</term>-level. 
+    Chunk-level elements are paragraphs and other elements which have similar 
+    structural properties. Phrase-level elements must be entirely contained within a 
+    paragraph or other chunk-level structure, This type includes emphasized or quoted
```
HelenaSabel:
```suggestion
    paragraph or other chunk-level structure. This type includes emphasized or quoted 
```

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -85,26 +85,39 @@ organization of the module described here. This should be read in
...
+    <term>phrase</term>, and <term>inter</term>-level. 
+    Chunk-level elements are paragraphs and other elements which have similar 
+    structural properties. Phrase-level elements must be entirely contained within a 
+    paragraph or other chunk-level structure, This type includes emphasized or quoted 
+    phrases, names, dates, etc. Inter-level elements can appear either within a 
+    paragraph or between paragraphs, and include bibliographic citations, notes, and 
+    lists. The <gi>ab</gi> (anonymous block) element is an alternative to 
+    <gi>p</gi> to be used in cases where paragraphs are not present, but chunk-level 
+    organization is still needed. <gi>ab</gi> may contain other <gi>ab</gi>s, 
+    may use the @type attribute, and does not necessarily represent a coherent set of
```
HelenaSabel:
```suggestion
    may use the <att>type</att> attribute, and does not necessarily represent a coherent set of 
```

ebeshero:
@martinascholger Want to take a look through this now? I think we've made the requested changes...

martinascholger (approved):
Looks good, thank you!

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -85,26 +85,39 @@ organization of the module described here. This should be read in
...
+    offset with whitespace or an indented first line. But paragraphs are 
+    not simply blocks of text. The paragraph may be thought of as a mid-level 
+    unit of sense, a coherent grouping of sentences. Paragraphs may, in turn, 
+    be grouped into larger divisions, such as chapters. Because it is a unit 
+    of sense rather than simply a block of text, the <gi>p</gi> element in TEI 
+    may contain other structures displayed as blocks, such as lists or quotes. 
+    This distinguishes it from the <ref 
+      target="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p">p</ref> 
+    element in HTML, which is primarily a block of text, and from the <gi>ab</gi> 
+    (anonymous block) element described in <ptr target="#SASE"/>) which may be used
```
sydb: extra close paren

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -85,26 +85,39 @@ organization of the module described here. This should be read in
...
+    phrases, names, dates, etc. Inter-level elements can appear either within a 
+    paragraph or between paragraphs, and include bibliographic citations, notes, and 
+    lists. The <gi>ab</gi> (anonymous block) element is an alternative to 
+    <gi>p</gi> which is useful in cases where paragraphs are not present, but chunk-level 
+    organization is still needed. <gi>ab</gi> may contain other <gi>ab</gi>s, 
+    may use the <att>type</att> attribute, and does not necessarily represent a coherent set of 
+    statements.</p>
+  <p>Paragraphs can contain any of the other elements described within this chapter, 
+    as well as some other elements which are specific to individual text types. Because 
+    paragraphs may appear in different base or additional tag sets, their possible
```
sydb: The concepts of “base tag set” and “additional tag set” were P4 ideas that, it seems to me, are not that relevant any more.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -85,26 +85,39 @@ organization of the module described here. This should be read in
...
+    structural properties. Phrase-level elements must be entirely contained within a 
+    paragraph or other chunk-level structure. This type includes emphasized or quoted 
+    phrases, names, dates, etc. Inter-level elements can appear either within a 
+    paragraph or between paragraphs, and include bibliographic citations, notes, and 
+    lists. The <gi>ab</gi> (anonymous block) element is an alternative to 
+    <gi>p</gi> which is useful in cases where paragraphs are not present, but chunk-level 
+    organization is still needed. <gi>ab</gi> may contain other <gi>ab</gi>s, 
+    may use the <att>type</att> attribute, and does not necessarily represent a coherent set of 
+    statements.</p>
+  <p>Paragraphs can contain any of the other elements described within this chapter,
```
sydb: This is true only if we take “contain” to mean “be a descendant of”; it is not true if we take it to mean “be a child of”. E.g., both `<analytic>` and `<divGen>` are defined in CO (this chapter); both can be a descendant of a `<p>` (p/biblStruct/analytic and p/floatingText/front/divGen); but neither can be a child of `<p>`.

hcayless: That sentence has been there a while. Presumably since before those elements were added. We can remove it.

hcayless (approved):

hcayless:
Think this is ready. Merging before freeze.

Issue #2358, opened 2022-10-14
definitions in the text: markup and handling


bansp:
Consider consistently marking up definitions in the text, so that they could be then (a) referenced across the Guidelines and (b) collectively displayed in some kind of drop-down section within a chapter, and (c) collected and displayed across the entire Guidelines, in the, say, front section, under the header "Definitions" and with references to where they appear.
Note that this is different from descriptions inside individual specs -- here I speak of terminological animals. 

The direct inspiration for this FR is the term "interpretation" that receives its own definition in the FSR chapter. I also noticed that both `<term>` and `<soCalled>` are used to highlight the terms there, and I am not sure how consistent this practice is. So, what I'm thinking of, very schematically, would be something like `<term>` and `<gloss>` (simply because they are already expected to be used like this, compare the [spec for 'gloss'](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-gloss.html)). More specifically, here's a fragment from 18.11:

> In particular, the term ‘interpretation’ when applied to a feature structure is not an interpretation in the model-theoretic sense, but is instead a minimally informative (or equivalently, most general) extension of that feature structure that is consistent with a set of constraints declared by an FSD. In linguistic application, such a system of constraints is the principal means by which the grammar of some natural language is expressed.

I could imagine this fragment encoded in the Guidelines roughly like the following:
```xml
<p> ...
In particular, the term <term xml:id="interpretation-FSR">interpretation</term> when applied to a feature
structure is not an interpretation in the model-theoretic sense, but is instead <gloss ref="#interpretation-FSR">a minimally informative (or equivalently, most general)
extension of that feature structure that is consistent with a set of constraints declared by an FSD</gloss>.  In linguistic application, such a system of constraints is the principal means by which the grammar of some natural language is expressed.
...
</p>
```

That would allow scripts to collect the terms and definientia and display them in a subsection at the beginning of the chapter, and then maybe the definiens could pop-up when the user hovered over `<term ref="#interpretation-FSR">interpretation</term>` (notice the `@ref` rather than `@xml:id`).

In some ODDs (and the crux of the request is that this goes way beyond the TEI Guidelines ODD), I could imagine separate lines such as
```xml
<entry type="definition">
     <term ident="interpretation-FSR">interpretation</term>
    <gloss>a minimally informative (or equivalently, most general)
                extension of that feature structure that is consistent with a set of constraints declared by an FSD</gloss>
</entry>
```

[...]

PR #2359, opened 2022-10-14
Dead iso cat

### LINKED ISSUE #1866: isocat.org links should be updated
The specs for `att.datcat` and `<gram>` contain links to the repository that used to be at isocat.org, but that is now obsolete. It may be sufficient to replace most of them with links to http://www.datcatinfo.net, but some of them are supposed to be links to ISO 12620, which might be better directed to https://www.iso.org/standard/37243.html.

### LINKED ISSUE #2227: Update ISOCat reference to DatCatInfo
From Roberto Rosselli Del Turco via TEI-L (2022-02-02): "in the Guidelines section devoted to Dictionaries there's one reference to the ISOCat standard, but the latter has been superseded by DatCatInfo: http://www.datcatinfo.net/ (with a more precise URL of course) instead of http://www.isocat.org/ in https://tei-c.org/release/doc/tei-p5-doc/en/html/DI.html#index-egXML-d52e79215."

ebeshero:
opened pull request during TEI Council meeting so we can review before October release.


P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -2658,25 +2658,28 @@ equivalently "poj." in Polish, or "Ez." in German, etc., what is actually referr
...
 In order to signal that this variety of surface markers in fact indicate the same
 underlying value, it is possible to align them with an external inventory of standardized
-values. The TEI provides means to align grammatical categories as well as their content
-with the ISOcat reference, which is a Web implementation of <ref target="#ISO-12620">ISO 12620</ref>.</p>
+values. The TEI provides the <ident type="class">att.datcat</ident> attribute class for the purpose of aligning grammatical (or indeed any sort of) categories as well as their values
+with a reference taxonomy of shared data categories.</p>
 <p>In the example below, a fragment of the entry for <foreign>isotope</foreign> cited
-in section <ptr target="#DITPGR"/> is adorned by references to ISOcat definitions for "part
-of speech" (<att>dcr:datcat</att>) and "adjective" (<att>dcr:valueDatcat</att>). Depending
+in section <ptr target="#DITPGR"/> is adorned by references to external definitions for "part
```
martinascholger: mayby standardised definitions instead of external?

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -2658,25 +2658,28 @@ equivalently "poj." in Polish, or "Ez." in German, etc., what is actually referr
...
+of speech" (<att>datcat</att>) and "adjective" (<att>valueDatcat</att>). Depending
 on the status and extent of the dictionary, various strategies may be used to reduce the
-redundancy of the repeated ISOcat references.<egXML xmlns="http://www.tei-c.org/ns/Examples" valid="feasible">
-<entry xmlns:dcr="http://www.isocat.org/ns/dcr">
+redundancy of the potentially redundant references.<egXML xmlns="http://www.tei-c.org/ns/Examples" valid="feasible">
```
martinascholger: "redundancy" and "redundant" in one sentence. Is "potentially redundant" really needed here? I think the sentence is clear without it.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -6,7 +6,7 @@ See the file COPYING.txt for details.
 $Date$
 $Id$
 -->
-<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<?xml-model href="https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-deadISOCat/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
```
martinascholger: please change reference to https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -269,17 +269,29 @@ combinations of feature names and values are used is to provide a
...
+<f name="part_of_speech" datcat="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3">
+<symbol valueDatcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545" value="NN"/>
+</f>
+<!-- ... -->
+</fs>
+</egXML></p>
+<p>Since the above representation takes up a lot of space and quickly becomes redundant and
+error-prone, it is possible to delegate the task of aligning with external repositories to
+elements such as <gi>fLib</gi>, <gi>fvLib</gi>, <gi>fDecl</gi>, or <gi>fsDecl</gi> to reduce the feature
+representation at hand and to increase its readability at the same time, as shown below.
```
martinascholger: as shown in the example below

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -269,17 +269,29 @@ combinations of feature names and values are used is to provide a
...
+ <egXML xmlns="http://www.tei-c.org/ns/Examples">
+<fs>
+<!-- ... -->
+<f name="POS" fVal="#commonNoun"/>
```
martinascholger, line 294: NN instead of #commonNoun?

P5/Source/Specs/gram.xml
```diff
@@ -152,23 +168,22 @@ $Id$
         </valItem>
       </valList>
       <remarks versionDate="2010-07-07" xml:lang="en">
-        <p>A much fuller list of values for the <att>type</att>
-        attribute may be generated from the data category registry
-        accessible from <ptr target="http://www.isocat.org/"/> and
-        maintained by ISO technical committee 37. <!--TC 37/SC 3/WD 12
+        <p>A much fuller list of values for the <att>type</att> attribute may be generated from the
+          data category registry accessible from <ptr target="http://www.isocat.org/"/> and
```
martinascholger: change of the isocat reference is needed

martinascholger (changes requested):

P5/Source/Specs/att.datcat.xml
```diff
@@ -6,52 +6,214 @@ See the file COPYING.txt for details
...
     </attDef>
   </attList>
   <exemplum xml:lang="en">
-    <p>In this example <att>dcr:datcat</att> relates the feature name to the data category
-      "partOfSpeech" and <att>dcr:valueDatcat</att> the feature value to the data category
-      "commonNoun". Both these data categories reside in the ISOcat DCR at <ref target="http://www.isocat.org">www.isocat.org</ref>, which is the DCR used by ISO
-      TC37 and hosted by its registration authority, the MPI for Psycholinguistics in
-      Nijmegen.</p>
-    <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#UND">
+    <p>The below example presents the TEI encoding of the <ref
```
martinascholger: The example below

P5/Source/Specs/att.datcat.xml
```diff
@@ -6,52 +6,214 @@ See the file COPYING.txt for details
...
+        target="https://en.wikipedia.org/wiki/Name%E2%80%93value_pair">name-value pair</ref>
+        "<tag>part-of-speech, common-noun</tag>" (where the name/key is abbreviated as "POS",
+      and the value is symbolized by "NN"). The <att>datcat</att> attribute relates the feature <hi
+        rend="italic">name</hi> to the data category "part of speech", while the attribute
+        <att>valueDatcat</att> relates the feature <hi rend="italic">value</hi> to the data category
+      "common noun". Both these data categories should be defined in an external and preferably open
```
martinascholger: should they be defined in or point to an external ... taxonomy?

P5/Source/Specs/att.datcat.xml
```diff
@@ -6,52 +6,214 @@ See the file COPYING.txt for details
...
   <exemplum xml:lang="en">
-    <p>In this example <att>dcr:datcat</att> relates the feature name to the data category
-      "partOfSpeech" and <att>dcr:valueDatcat</att> the feature value to the data category
-      "commonNoun". Both these data categories reside in the ISOcat DCR at <ref target="http://www.isocat.org">www.isocat.org</ref>, which is the DCR used by ISO
-      TC37 and hosted by its registration authority, the MPI for Psycholinguistics in
-      Nijmegen.</p>
-    <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#UND">
+    <p>The below example presents the TEI encoding of the <ref
+        target="https://en.wikipedia.org/wiki/Name%E2%80%93value_pair">name-value pair</ref>
+        "<tag>part-of-speech, common-noun</tag>" (where the name/key is abbreviated as "POS",
```
martinascholger: this part is difficult to read, but I don't have a good solution yet. At least, "part-of-speech, common-noun" should not be encoded with `<tag>`

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: Sounds better indeed, thanks. I think I'll be fixing this locally and then commit.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: true, fixed

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: By all means, except it will light up in red

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: ok :-)

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: Nope, that's the trick. It's a matter for a separate section (not necessarily of the Guidelines): you can't use "true" tagset labels in the fVal references, because some of them are not well-formed from the point of view of datatype constraints on URIs. So with fVal you are either lucky or you need to work through a redirection. The latter is a safer way. 
I'll demonstrate that separately, I have a little project on a side that studies that.

P5/Source/Specs/gram.xml
bansp: embarrassing... :-) Sure, going to fix that now, thanks :-)

bansp: Actually, this one is a tougher issue. I'll leave some vague statement here, or else I'd have to propose more than I was asked to (and there's no time, in this round, I'm afraid).

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
peterstadler: Yes, red is ok ;)
Otherwise we'll forget about it to change the schema reference back later

martinascholger (changes requested):

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -6,7 +6,7 @@ See the file COPYING.txt for details.
 $Date$
 $Id$
 -->
-<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<?xml-model href="https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-deadISOCat/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
```
peterstadler: as above, change back to 
```suggestion
<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
```

P5/Source/Guidelines/en/FS-FeatureStructures.xml
peterstadler: ok, but where does the "#commonNoun" refer to? Do we need an explanatory sentence?

martinascholger: +1 -- I'm now in the rendered version where it is easier to see which explanation belongs to which example.

P5/Source/Specs/att.datcat.xml
```diff
@@ -6,52 +6,214 @@ See the file COPYING.txt for details
...
+      feature value library, discussed in chapter <ptr target="#FS"/>.</p>
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
       <fs>
-        <f name="POS" dcr:datcat="http://www.isocat.org/datcat/DC-1345" fVal="#commonNoun" dcr:valueDatcat="http://www.isocat.org/datcat/DC-1256"/>
+        <f name="POS" fVal="#commonNoun"/>
+        <!-- ... -->
       </fs>
     </egXML>
+    <p>The assumption here is that the relevant feature values are collected in a place that the
+      annotation document in question has access to -- preferably, a single document per linguistic
```
martinascholger: please use — instead of -- (also a bit further down)

martinascholger (changes requested):

P5/Source/Specs/gram.xml
bansp: because there is no _reliable_ source to recommend nowadays, and the only one I can think of is the Lex0 suggestion at https://dariah-eric.github.io/lexicalresources/pages/TEILex0/spec.html#TEI.gram (unfortunately, it's hard-coded)

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: done, thanks, gosh, I'm lost in this system, changed everything locally, about to commit and push -- I hope that's OK

P5/Source/Specs/att.datcat.xml
bansp: right, I have sinned here. Using `<code>` for this makes it bolded, and thus a bit horrible. I could use `<hi>` with monotype, except I don't know what the "standardized" value for that is, across the Guidelines

P5/Source/Specs/gram.xml
bansp: So I got a bit stuck here (well, for 40 minutes, it seems), trying to find a better solution than just deleting the `<remarks>`.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: I have added one more sentence there, due to Peter's suggestion that I can't see anywhere except my mail. (I thought it would just show up here.) The sentence points to the following section where `#commonNoun` is exemplified.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: Done for both chapters.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: I'm not sure what "+1" means in this context :-)

P5/Source/Specs/att.datcat.xml
bansp: aren't both grammatical? :-) But sure, if it's tickled your eye, I'll modify the wording.

P5/Source/Specs/att.datcat.xml
bansp: DCs -- defined. References -- point. I'll see how I can improve the wording.

P5/Source/Specs/gram.xml
bansp: I'm resolving this tentatively -- there is still work to do (_systemic_ work!) but I suggest that (unless you have a better solution) we leave this with the bland content that I have replaced the reference to ISOCat with.

P5/Source/Specs/att.datcat.xml
bansp: Oh dang, that's a search-and-replace in files. Will do.

bansp: Hmm, the scope for this is wider than just the files I've modified -- just sayin'...

P5/Source/Specs/att.datcat.xml
bansp: I'd say it's all good. Will see, maybe split this into two `<p>`s -- but I wouldn't like to omit the info.

P5/Source/Specs/att.datcat.xml
bansp: ... ah, I used `<hi rend="mono"` elsewhere here, and was going to use that for that abused `<tag>` as well. Except "mono" doesn't work, siiigh, I'll try to search across the Stylesheets to see if they handle that in any way...

bansp: Well, I'll use `<code>` and blame the horrible outcome on the Stylesheets, I guess. I don't want to search through them again...

P5/Source/Specs/att.datcat.xml
bansp: done

P5/Source/Specs/att.datcat.xml
bansp: used `hi style="font-family:monospace"`

P5/Source/Specs/att.datcat.xml
bansp: I have reworded this slightly but I won't be able to see how it reads, because the build won't compile due to the proper reference to the validation script now...

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: I will tentatively resolve this, to see how that affects the entire process (it's the last unresolved issue)

bansp:
This is now up for re-review, if that's the procedure, but I'm afraid it is going to bug out due to the namespace issue. And I'm also not at all sure how to handle the JSON file that is apparently conflicting.

bansp:
Should I merge the current dev branch into this, or will you be rebasing the changes?

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -352,8 +367,22 @@ each one represented as an identifiable <gi>fs</gi> element within a
 Other <gi>f</gi> elements may invoke them by reference, using the
 <att>fVal</att> attribute; for example, one might use them in a
 feature value pair such as: <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#UND"><f name="dental-fricative" fVal="#T.DF"/> </egXML> rather than expanding the hierarchy of the
-component phonological features explicitly.  </p>
-<p>Feature structures stored in this way  may also be associated with
+component phonological features explicitly.</p>
+<p>The feature structure that concludes section <ptr target="#FSSY"/> above, identifying the
+value of some part of speech to be a "common noun" can be used in tandem with a
```
HelenaSabel: I’d recommend using an element instead of the straight quotes. Maybe `<soCalled>common noun</soCalled>` (or  `<q>`). Same comment goes for line 277 above, and 1625 below.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -352,8 +367,22 @@ each one represented as an identifiable <gi>fs</gi> element within a
...
+feature-value library, which offers a way to encode a grammatical <term>tagset</term>, in this case
+containing labels for parts of speech:
+<egXML xmlns="http://www.tei-c.org/ns/Examples">
+<fvLib n="POS values">
+<symbol xml:id="commonNoun" value="NN" datcat="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3"/>
+<symbol xml:id="properNoun" value="NP" datcat="http://hdl.handle.net/11459/CCR_C-1371_fbebd9ec-a7f4-9a36-d6e9-88ee16b944ae"/>
+<!-- ... -->
+</fvLib>
+</egXML>
+Such a feature-value library combines the standard short symbolic label for a part of speech (e.g., "NN") with a
```
HelenaSabel:
```suggestion
Such a feature-value library combines the standard short symbolic label for a part of speech (e.g., <val>NN</val>) with a 
```

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+        >指定された要素の内容または指定された属性の値を，ISOcatの適切かつ単純なデータカテゴリ（または複数のデータカテゴリ）と対応付けるPID（永続的識別子）を含む．</desc>
+      <datatype maxOccurs="unbounded">
+        <dataRef key="teidata.pointer"/>
+      </datatype>
+    </attDef>
+    <attDef ident="targetDatcat" usage="opt">
+      <desc versionDate="2022-09-17" xml:lang="en">provides a definition of, and/or general
+        information about, information structure of an object referenced or modeled by the
+        containing element, by reference to an external taxonomy or ontology. This attribute has the
+        characteristics of the @datcat attribute, except that it addresses not its containing
```
HelenaSabel:
```suggestion
        characteristics of the <att>datcat</att> attribute, except that it addresses not its containing
```

P5/Source/Guidelines/en/FS-FeatureStructures.xml
sydb: I agree completely, and the same goes for lines 2664 & 2665 in DI chapter (previous file), too. I do not think `<soCalled>` is appropriate, though.n (But will be happy to change my mind if someone argues convincingly that the part of speech being discussed is _not_ a common noun, rather said encoding is in error; or some such.) My instinct is either `<q>` or `<term>`?
In any case, though, the comma afterwards is missing (before “can” which perhaps should be “may”), and the sentence itself feels a bit run-onny:

> The feature structure that concludes section [FSSY] above, identifying the value of some part of speech to be a “common noun” can be used in tandem with a feature-value library, which offers a way to encode a grammatical “tagset”, in this case containing labels for parts of speech: [EXAMPLE]

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+        <symbol xml:id="commonNoun" value="NN" datcat="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3"/>
+        <symbol xml:id="properNoun" value="NP" datcat="http://hdl.handle.net/11459/CCR_C-1371_fbebd9ec-a7f4-9a36-d6e9-88ee16b944ae"/>
+        <!-- ... -->
+      </fvLib>
+    </egXML>
+    <p>Note that these Guidelines do not prescribe a specific choice between <att>datcat</att> and
+        <att>valueDatcat</att> in such cases. The former is the generic way of referencing a data
+      category, whereas the latter is more specific, in that it references a data category that
+      represents a value. The choice between them comes into play where a single element — or a
+      tight element complex, such as the <gi>f</gi>/<gi>symbol</gi> complex illustrated above --
```
HelenaSabel:
```suggestion
      tight element complex, such as the <gi>f</gi>/<gi>symbol</gi> complex illustrated above —
```

bansp: ouch, I've searched for ' -- ', my bad... thanks!

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: sigh, true, I wonder how many straight quotes I have left there... that may take a while ;-)

bansp:
Hey, I'm learning to do this online, and that preserves the attribution to the reviewer -- thanks a lot, I've just gained something.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: Tried to use `<q>` for quoted bits, and `<soCalled>` for, e.g., names of concepts. Wow, semantic markup... cool. :-)

bansp: And it's just struck me that maybe (just maybe) some of those could be `<term>`s, but I think that sorting that out needs a bit of quiet and focus, and some target framework, such as that sketched in #2358 -- where you'd, for example, get a pop-up over a `<term>`, citing its definition.

bansp: I believe I have eliminated simple quotes in the spec and in the FS chapter, with `<q>` where the quotes were straightforward. I even ended up removing quotes where they were not really needed, it seems. 
There are two sentences in FS that feel "run-onny" (if I interpret that correctly) -- in both cases they talk about some functionality that should/may accompany a suggestion exemplified in an earlier section. The potentially unstable word in both is "concludes", because one could imagine the target section to grow, one day. But today, they are true, and I think that when the FS chapter receives a solid revision, these two instances will be modified if needed.

bansp: ... and I've touched the DI chapter as well.

peterstadler:
Release team looked at this and requests feedback from some native speakers and experienced Guidelines editors (@sydb @ebeshero @hcayless). We might get this into the release 4.5.0 when this is done on this weekend 😬

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -2653,30 +2653,33 @@ sometimes informally stated values for the particular descriptive features, the
...
-for instance, the value of the grammatical category of Number is provided as "sg.", "sing.", "Singular", or
-equivalently "poj." in Polish, or "Ez." in German, etc., what is actually referred to is always the same 
+for instance, the value of the grammatical category of Number is provided as <q>sg.</q>, <q>sing.</q>, <q>Singular</q>, or
+equivalently <q>poj.</q> in Polish, or <q>Ez.</q> in German, etc., what is actually referred to is always the same 
 grammatical value that can be rendered with a plethora of markers, depending on the publisher, language, or lexicographic tradition.
 In order to signal that this variety of surface markers in fact indicate the same
 underlying value, it is possible to align them with an external inventory of standardized
-values. The TEI provides means to align grammatical categories as well as their content
-with the ISOcat reference, which is a Web implementation of <ref target="#ISO-12620">ISO 12620</ref>.</p>
+values. The TEI provides the <ident type="class">att.datcat</ident> attribute class for the purpose of aligning grammatical (or indeed any sort of) categories as well as their values
```
peterstadler, line 2661: the "(or indeed any sort of) categories" sounds very broad and might overlap with `@ana`. Was that on purpose? Then we'd need some explanatory sentence about the difference of those attributes, methinks.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -2653,30 +2653,33 @@ sometimes informally stated values for the particular descriptive features, the
...
+with a reference taxonomy of shared data categories.</p>
 <p>In the example below, a fragment of the entry for <foreign>isotope</foreign> cited
-in section <ptr target="#DITPGR"/> is adorned by references to ISOcat definitions for "part
-of speech" (<att>dcr:datcat</att>) and "adjective" (<att>dcr:valueDatcat</att>). Depending
+in section <ptr target="#DITPGR"/> is adorned by references to standardised definitions for <soCalled>part
```
peterstadler: Other occurrences have "standardized":
```suggestion
in section <ptr target="#DITPGR"/> is adorned by references to standardized definitions for <soCalled>part
```

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -269,20 +269,35 @@ combinations of feature names and values are used is to provide a
...
+standardized external taxonomies.</note> In the following example, both
 the feature <val>part_of_speech</val> and its value
-<val>#commonNoun</val> are aligned with the respective definitions
-provided by <ref target="#ISO-12620">ISO DCR (Data Category
-Registry)</ref>, as implemented by ISOcat.
-<egXML xmlns="http://www.tei-c.org/ns/Examples" valid="feasible" source="#UND">
-<fs xmlns:dcr="http://www.isocat.org/ns/dcr">
-<!--...-->
-<f name="part_of_speech" dcr:datcat="http://www.isocat.org/datcat/DC-1345" fVal="#commonNoun" dcr:valueDatcat="http://www.isocat.org/datcat/DC-1256"/>
+<val>NN</val> (standing for "common noun") are aligned with the respective definitions
```
peterstadler:
```suggestion
<val>NN</val> (standing for <q>common noun</q>) are aligned with the respective definitions
```

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
   <exemplum xml:lang="en">
-    <p>In this example <att>dcr:datcat</att> relates the feature name to the data category
-      "partOfSpeech" and <att>dcr:valueDatcat</att> the feature value to the data category
-      "commonNoun". Both these data categories reside in the ISOcat DCR at <ref target="http://www.isocat.org">www.isocat.org</ref>, which is the DCR used by ISO
-      TC37 and hosted by its registration authority, the MPI for Psycholinguistics in
-      Nijmegen.</p>
-    <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#UND">
+    <p>The example below presents the TEI encoding of the <ref
+        target="https://en.wikipedia.org/wiki/Name%E2%80%93value_pair">name-value pair</ref> <q><hi
+        style="font-family:monospace">&lt;part of speech, common noun></hi></q>, where the name (key) <q>part
```
peterstadler: Is this styling intended here? do you want `<code>`?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      should be defined in an external and preferably open reference taxonomy or ontology.</p>
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
+      <fs>
+        <f name="POS" datcat="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3">
+          <symbol valueDatcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545" value="NN"/>
+        </f>
+        <!-- ... -->
+      </fs>
+    </egXML>
+    <p>"NN" is the symbol for common noun used e.g. in the <ref
```
peterstadler:
```suggestion
    <p><q>NN</q> is the symbol for common noun used e.g. in the <ref
```

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
+      <gramGrp>
+        <pos datcat="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3"
+          valueDatcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545">NN</pos>
+      </gramGrp>
+    </egXML>
+    <p>Again, the issue of efficiency and redundancy of this type of interoperable markup demands
+      that the references to the particular data categories are provided either in the header of the
+      dictionary or at the level of ODD, in the element <gi>equiv</gi> (child of <gi>elementSpec</gi>), which constitutes an
+      equivalent of the <ident>att.datcat</ident> class.</p>
```
peterstadler: On first read, I stumbled here, and I wonder, whether the subordinate clause "in the element <equiv> …" is helpful or confusing. I think I tend to the latter and would simply drop it, but I'm happy to be overruled!

peterstadler (changes requested):

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: It is meant to be broad, because the need to fix references goes beyond linguistic applications (wouldn't FRBR apply? it's a bit hard for me to gauge the need, not being an expert in that field, but these attributes _are_ intended for general use, also where they complement `<equiv>`). I'd rather not put time into this in this round, if possible (I got behind with all the other deadlines), so I can delete the brackets, but then I'd be telling only half the story. Is there a chance to leave this as-is and expand afterwards, when I know more about other fields in this regard, or maybe an expert on that happens to float by, and we catch them and put them to work? ;-)
And `@ana` is different -- it's for analysis, while datcats are for fixing references, establishing identities or confirming non-identities (again, FRBR could be an analogue, I think).

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: Yeah, I looked at what Martina suggested, thought I'd need to use that spelling, used it, and then forgot to check the other occurrences... I'll gladly just "commit suggestion" here, if there's no orthographic guidelines on that.

P5/Source/Specs/att.datcat.xml
bansp: This is an itch, I admit. I initially used `<code>` and was a bit horrified by the result -- bold type is not a good device there. So I went around that. Is there a way to use code in such a way that it doesn't get bolded? I tried to locate the processing of the attributes of `<code>` in the Stylesheets, but either gave up or something distracted me.

P5/Source/Specs/att.datcat.xml
bansp: omg, I thought I'd caught them all... thanks...

P5/Source/Specs/att.datcat.xml
bansp: OK, will give it a closer look after I get a covid shot in half an hour. The results may be unexpected... ;-)

bansp: Right, I think I should rephrase this, the more so that the header is only one such place; fsdDecl is another. And 'constitutes' is maybe too lofty, not sure.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
sydb: Personally I am OK with leaving the parenthetical. @peterstadler’s point that we may need some explanation of the difference between `@ana` and `@datcat` is reasonable, but may not be essential unless users express confusion.

The part I am uncomfortable with is “… categories as well as their values”. It used to be “content”, not “values”. I do not find either intuitive, and have to read further (e.g., the tagdoc) and look at examples to figure out what is going on. I do not really have a better suggestion at the moment, though. So unless someone else does, perhaps we should just pin this by putting `<?tei winita better wording needed ?>` in there.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
sydb: Yes, use the “ize” version:
> … use the "-ize" suffix instead of "-ise" for words that are a Latinized version of a Greek suffix because while both spellings are given in the Oxford English Dictionary, only the "-ize" form is used in American English.

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -2653,30 +2653,33 @@ sometimes informally stated values for the particular descriptive features, the
...
+of speech</soCalled> (<att>datcat</att>) and <soCalled>adjective</soCalled> (<att>valueDatcat</att>). Depending
```
sydb: `<soCalled>`? Are we trying to assert that they are not usually called “part of speech” and “adjective”, or to establish that we do not think they should be called “part of speech” and “adjective”, or some such?

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
```diff
@@ -2653,30 +2653,33 @@ sometimes informally stated values for the particular descriptive features, the
...
-<entry xmlns:dcr="http://www.isocat.org/ns/dcr">
+redundancy of references.<egXML xmlns="http://www.tei-c.org/ns/Examples" valid="feasible">
+<entry>
 <!--...-->
 <form>
 <orth>isotope</orth>
 </form>
 <gramGrp>
-<pos dcr:datcat="http://www.isocat.org/datcat/DC-1345" dcr:valueDatcat="http://www.isocat.org/datcat/DC-1230">adj</pos>
+<pos
```
sydb, line 2674: Wondering aloud here …
Wouldn’t this be easier for a reader to grok if we used some clear indirection? E.g.
~~~xml
      <listPrefixDef>
        <prefixDef ident="pb" matchPattern="pos" replacementPattern="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3"/>
        <prefixDef ident="pb" matchPattern="adj" replacementPattern="http://hdl.handle.net/11459/CCR_C-1230_23653c21-fca1-edf8-fd7c-3df2d6499157"/>
      </listPrefixDef>
      <!-- ... -->
      <entry>
        <!--...-->
        <form>
          <orth>isotope</orth>
        </form>
        <gramGrp>
          <pos datcat="pb:pos" valueDatcat="pb:adj">adj</pos>
        </gramGrp>
        <!--...-->
      </entry>
~~~

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -269,20 +269,35 @@ combinations of feature names and values are used is to provide a
...
+<fs>
+<f name="part_of_speech" datcat="http://hdl.handle.net/11459/CCR_C-396_5a972b93-2294-ab5c-a541-7c344c5f26c3">
+<symbol valueDatcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545" value="NN"/>
+</f>
+<!-- ... -->
+</fs>
+</egXML></p>
+<p>Since the above representation takes up a lot of space and quickly becomes redundant and
+error-prone, it is possible to delegate the task of aligning with external repositories to
+elements such as <gi>fLib</gi>, <gi>fvLib</gi>, <gi>fDecl</gi>, or <gi>fsDecl</gi> to reduce the feature
```
sydb, line 289: Possible delegation to `<fLib>`, `<fvLib>`, `<fdecl>`, or `<fsDecl>` is listed here, but only to `<fvLib>` or `<taxonomy>`, below (~10 lines down).

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -1176,15 +1205,14 @@ using an XML identifier are discussed in <ptr target="#SAUR"/></note>
 way of accomplishing this is to add an XML identifier to each
 <gi>fsDecl</gi> element in <ident type="file">example.xml</ident>:
 <egXML xmlns="http://www.tei-c.org/ns/Examples" valid="feasible" source="#UND">
-           <!-- ... -->
-           <fsdDecl>
-           <fsDecl type="gpsg" xml:id="GPSG">
+<fsdDecl>
```
sydb, line 1208: The odd indentation introduced into this example makes no difference in the output using our current Stylesheets. But if I get my way someday soon the exact indentation that is used in the XML file will be reproduced in the output. This this will, if and when that happens, become ugly.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -1593,6 +1621,27 @@ in a construction.  Since PFORM is specified above as an open set,
 This example makes use of a negated value:    <code>&lt;vNot&gt;&lt;string/&gt;&lt;/vNot&gt;</code>
 subsumes any string that is not the empty
 string.</p>
+<p>For the reduced feature structure that concludes section <ptr target="#FSSY"/> above and 
+identifies the value of some part of speech to be a common noun, it is possible to align 
+the concept of <soCalled>part of speech</soCalled> with its definition and persistent identifier thanks to the
```
sydb: Delete “of” — redundant, no?

Furthermore, “thanks to” is somewhat colloquial; and it does not seem necessary to bring in the class here, only the attribute. How about
~~~xml
… it is possible to align the concept <q>part of speech</q> with its definition and persistent identifier using the <att>targetDatcat</att> attribute, which connects the modeled XML object with the appropriate locus in a reference taxonomy, as shown below:
~~~
?

P5/Source/Guidelines/en/FS-FeatureStructures.xml
```diff
@@ -1593,6 +1621,27 @@ in a construction.  Since PFORM is specified above as an open set,
...
+<vAlt>
+<symbol value="NN" datcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545"/>
+<symbol value="NP" datcat="http://hdl.handle.net/11459/CCR_C-1371_fbebd9ec-a7f4-9a36-d6e9-88ee16b944ae"/>
+   <!-- ... -->
+</vAlt>
+</vRange>
+</fDecl>
+</egXML>
+The above example declares the feature <q>POS</q> as instantiating the corresponding concept defined in a reference 
+taxonomy / ontology, and defines the range of values of the feature at hand by listing the appropriate
```
sydb, line 1642: I realize this is my hobby horse, and it may be the case that no one else will object to the solidus (‘/’, U+002F), or worse (from my point of view) some people will outright prefer it. But I _hate_ use of a solidus to mean “either ‘and’ or ‘or’, I am too lazy to tell you which”. In this case we should either pick “taxonomy” or “ontology” and drop the other, or use “or” IMHO.

P5/Source/Specs/att.datcat.xml
bansp: Thanks, Peter -- you made me realise that I forgot about something. Must have been in a hurry to get to features, and then forgot about this part of the system. And, as it is, the system is not complete in the Guidelines, and one way to go about it would be to delete this part of remarks altogether -- simply not mention the lack of efficiency, and attack the issue after the release, with a slow process.
Another possibility would be to try to amend what I forgot by adding `tagUsage` to att.datcat, and this way to save at least part of the picture, while begging questions about another part, but... questions aren't bad, are they. And I'm thinking in parallel with writing this reply, and my current thinking is to try to add `tagUsage`, and if the reviewers don't approve, to offer dropping this `<p>` element altogether as an alternative.

bansp: I have addressed this as above, in commit 3b05c99a4dcd97fa1db90580c3c80b717515b03b. And in doing so, I realised that I can rearrange some of the secondary `egXML`s to be inside the `p` children of `exemplum` -- it had been a pain to me for a while that (I thought) I could only use one `egXML` per `exemplum`. But no. So I'd like to push one more commit, merely rearranging the elements of examples slightly, for better readability.

lb42 (commented):
"rather than get repeated" is barbarous. Try "rather than being repeated".

P5/Source/Specs/att.datcat.xml
sydb: But it _is_ code. Can we try `<code style="font-weight: normal; font-family: monospace;">`?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+    <p>The example below presents the TEI encoding of the <ref
+        target="https://en.wikipedia.org/wiki/Name%E2%80%93value_pair">name-value pair</ref> <q><hi
+        style="font-family:monospace">&lt;part of speech, common noun></hi></q>, where the name (key) <q>part
+      of speech</q> is abbreviated as <q>POS</q>, and the value, <q>common noun</q> is symbolized by <q>NN</q>. The
+      entire name-value pair is encoded by means of the element <gi>f</gi>. In TEI XML, that element
+      acts as the container, labeled with the <att>name</att> attribute. Its contents may be complex
+      or simple. In the case at hand, the content is the symbol <q>NN</q>.</p>
+    <p>The <att>datcat</att> attribute relates the feature <hi rend="italic">name</hi> (i.e., the key) to the data
+      category <q>part of speech</q>, while the attribute <att>valueDatcat</att> relates the feature <hi
+        rend="italic">value</hi> to the data category <soCalled>common noun</soCalled>. Both these data categories
```
sydb: Mildly uncomfortable with the `<hi>`s, but maybe we should leave them for expediency. But the `<soCalled>` seems outright wrong … it _is_ a common noun, isn’t it?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+          <symbol valueDatcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545" value="NN"/>
+        </f>
+        <!-- ... -->
+      </fs>
+    </egXML>
+    <p><q>NN</q> is the symbol for common noun used e.g. in the <ref
+        target="http://ucrel.lancaster.ac.uk/claws7tags.html">CLAWS-7 tagset</ref> defined by the
+      University Centre for Computer Corpus Research on Language at the University of Lancaster. The
+      very same data category used for tagging an early version of the British National Corpus, and
+      coming from the so-called <ref target="http://www.natcorp.ox.ac.uk/docs/c5spec.html">BNC Basic
```
sydb: Why is the _BNC Basic Tagset_ “so-called”?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+        (C5) tagset</ref>, uses the symbol <q>NN0</q> (rather than <q>NN</q>). Making these values semantically
+      interoperable / interchangeable would be extremely difficult without a human expert if they
```
sydb: I still hate the solidus.

* interoperable
* interchangeable
* interoperable or interchangeable
* interoperable or even interchangeable
* Data interchange of these values would be
* Data interchange, let alone data interoperability, of these values would be

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      were not anchored in a single point of an established reference taxonomy of morphosyntactic
+      data categories. In the case at hand, the string <q><ref
+        target="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545"
+        >http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545</ref></q> is both
+      a persistent identifier of the data category in question, as well as the address of the place where a shared definition
```
sydb: 1. as well as the web address of the place where a shared definition of CN is provided.
1. a pointer to a shared definition of CN.
1. de-referenceable URL that points to a shared definition of CN.

I like (2) myself.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      of <q>common noun</q> is provided.</p>
+    <p>While the symbols <q>NN</q>, <q>NN0</q>, and many others (often coming from languages other than
+      English) are implicitly members of the container category <q>part of speech</q>, it is sometimes
+      useful not to rely on such an implicit relationship but rather use an explicit identifier
+      for that data category, to distinguish it from other morphosyntactic data categories, such
+      as <soCalled>gender</soCalled>, <soCalled>tense</soCalled>, etc. For that purpose, the above example uses the <att>datcat</att>
```
sydb: Again with the `<soCalled>`s …

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
       <fs>
-        <f name="POS" dcr:datcat="http://www.isocat.org/datcat/DC-1345" fVal="#commonNoun" dcr:valueDatcat="http://www.isocat.org/datcat/DC-1256"/>
+        <f name="POS" fVal="#commonNoun"/>
+        <!-- ... -->
       </fs>
     </egXML>
+    <p>The assumption here is that the relevant feature values are collected in a place that the
+      annotation document in question has access to — preferably, a single document per linguistic
+      resource, for example a <gi>fsdDecl</gi> that is XIncluded as a sibling of <gi>text</gi> or a
```
sydb: I, for one, pronounce “fsdDecl” as “eff-ess-dee-decl”, and thus the article should be “an”.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      child of <gi>encodingDesc</gi>; a <gi>taxonomy</gi> available resource-wide (e.g., in a shared
+      header) is also an option. The example below presents a <gi>fvLib</gi> element that collects the
```
sydb: ditto: “an”

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      relevant feature values (most of them omitted).</p>
+  </exemplum>
+  <exemplum xml:lang="en">
+    <p>This example shows one way of encoding a <hi rend="bold">tagset</hi>, i.e., an established inventory of values of
```
sydb: `<hi>`? Should be `<term>`, no?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+          <vAlt>
+            <symbol value="NN" datcat="http://hdl.handle.net/11459/CCR_C-1256_7ec6083c-23d4-224d-6f94-eecbe6861545"/>
+            <symbol value="NP" datcat="http://hdl.handle.net/11459/CCR_C-1371_fbebd9ec-a7f4-9a36-d6e9-88ee16b944ae"/>
+            <!-- ... -->
+          </vAlt>
+        </vRange>
+      </fDecl>
+    </egXML>
+    <p>Above, the modeled feature uses <att>targetDatcat</att>, because if it used
+        <att>datcat</att> then we would be saying that the <gi>fDecl</gi> is an instance of the part-of-speech container data category,
```
sydb: I think this wording is suboptimal, but having a hard time improving it. Perhaps
~~~xml
… because if it used <att>datcat</att> it would be asserting that it is an instance of the part-of-speech container data category, whereas it is not — it models a container (<gi>f</gi>) that encodes a part of speech.
~~~
?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      whereas it is not — the <gi>fDecl</gi> models a container (<gi>f</gi>) that encodes a part
+      of speech. Note also that it is the <gi>f</gi> that is modeled above, not its values, which
+      are used as direct references to data categories, hence the use of <att>datcat</att> in the
```
sydb: Maybe change comma after “categories” to semicolon?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+        <gi>symbol</gi> element.</p>
+  </exemplum>
+  
+  <remarks versionDate="2022-09-17" xml:lang="en">
+    <p>The TEI Abstract Model can be expressed as a hierarchy of attribute-value matrices (AVMs)
+      of various types and of various levels of complexity, nested or grouped in various ways. At
+      the most abstract level, an AVM consists of an information container and the value
+      ('contents') of that container.</p>
```
sydb: `<term`, `<q>`, or perhaps `<soCalled>`, no?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+    <p>A simple example of an XML serialization of such structures is, on the one hand, the opening
+      and closing tags that delimit and name the container, and, on the other, the content enclosed
+      by the two tags that constitues the value. An analogous example is an
+      attribute name and the value of that attribute.</p>
+    <p>In a TEI XML example of two equivalent serializations expressing the name-value pair
+      <hi style="font-family:monospace">&lt;part-of-speech,common-noun></hi>, namely 
+      <hi style="font-family:monospace">&lt;pos>commonNoun&lt;/pos></hi> and 
+      <hi style="font-family:monospace">pos="common-noun"</hi>, one would
+      classify the element <gi>pos</gi> and the attribute <att>pos</att> as containers (mapping onto
+      the first member of the relevant name-value pair), while the <hi style="font-family:monospace">text()</hi>
```
sydb: Don’t we still call that stuff the XDM calls a text node (accessed by `text()`) “character data” or even “parsed character data”? I am not sure.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      content of <gi>pos</gi> or the string/symbolic value of <att>pos</att> would be seen as
```
sydb: Does “string/symbolic” actually add anything here?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      mapping onto the second member of the pair.</p>
+    <p>The <ident type="class">att.datcat</ident> class provides means of addressing the containers
+      and their values, while at the same time providing a way to interpret them in the context of
+      external taxonomies or ontologies. Aligning e.g. both the <gi>pos</gi> element and the
+        <att>pos</att> attribute with the same value of an external reference point (i.e., an entry
+      in an agreed taxonomy) affirms the identity of the concept serialised by both the element container
+      and the attribute container, and optionally provides a definition of that concept (in the case at hand,
+      the concept <soCalled>part of speech</soCalled>).</p>
```
sydb: _Again_ with the authorial distance! It _is_ the concept “part of speech”, no?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+    <p>The value of the <ident type="class">att.datcat</ident> attributes should be a PID
+      (persistent identifier) that points to a specific — and, ideally, shared — taxonomy or
+      ontology. Among the resources that to a smaller or greater extent can be used as inventories
```
sydb: 1. “lesser”, not “smaller”
2. Probably easier to read if that snippet is moved to after “can”.

Thus:
> Among the resources that can, to a lesser or greater extent, be used as inventories

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      of (more or less) standardized linguistic categories, are the GOLD ontology, <ref
```
sydb: I don’t think there should be a comma after “categories”.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+        target="https://www.clarin.eu/content/clarin-concept-registry">CLARIN CCR</ref>, <ref
+        target="https://github.com/acoli-repo/olia">OLiA</ref> or
```
sydb: Just double-checking: you intend OLiA and TermWeb DatCatInfo to be alternatives to one another, yes?

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+    </egXML>
+    <p>Above, the modeled feature uses <att>targetDatcat</att>, because if it used
+        <att>datcat</att> then we would be saying that the <gi>fDecl</gi> is an instance of the part-of-speech container data category,
+      whereas it is not — the <gi>fDecl</gi> models a container (<gi>f</gi>) that encodes a part
+      of speech. Note also that it is the <gi>f</gi> that is modeled above, not its values, which
+      are used as direct references to data categories, hence the use of <att>datcat</att> in the
+        <gi>symbol</gi> element.</p>
+  </exemplum>
+  
+  <remarks versionDate="2022-09-17" xml:lang="en">
```
sydb, line 223: This entire `<remarks>` seems more important than a remark in a tagdoc should be. Seems to me it belongs in the main prose of the _Guidelines_. But to be honest, I am not sure exactly where it would go.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      the concept <soCalled>part of speech</soCalled>).</p>
+    <p>The value of the <ident type="class">att.datcat</ident> attributes should be a PID
+      (persistent identifier) that points to a specific — and, ideally, shared — taxonomy or
+      ontology. Among the resources that to a smaller or greater extent can be used as inventories
+      of (more or less) standardized linguistic categories, are the GOLD ontology, <ref
+        target="https://www.clarin.eu/content/clarin-concept-registry">CLARIN CCR</ref>, <ref
+        target="https://github.com/acoli-repo/olia">OLiA</ref> or 
+      <ref target="https://datcatinfo.net/">TermWeb's DatCatInfo</ref>, and also the 
+      <ref target="https://universaldependencies.org/">Universal
+      Dependencies</ref> inventory, on the assumption that its URI are going to persist. Apart from these
```
sydb: Disagreement in number: either “its URIs are going to” or “its URI will”.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      global resources, it is imaginable to address a project-specific taxonomy as well, if global
+      interoperability is not the goal.</p>
```
sydb: Subjunctive. (Or is it conditional?) Either way, I am wondering if “were” is more correct than “is”.

A project-specific taxonomy might be useful for other reasons, too, no? Most notably if a project wanted a taxonomy different from those globally available, typically because more or less precision than provided by the globally available ones is desired.

I am also quite confused by the “as well” and the “global interoperability not the goal”. Do you mean “as well” as “we can imagine a project specific taxonomy in addition to our imagining of global taxonomies” or “we can imagine a TEI user pointing to both a project specific taxonomy and a global taxonomy”? I suspect the former, but either way this ambiguity is problematic.

Thus perhaps “It is imaginable that a project may choose to address a local taxonomy instead, but this risks losing the enormous advantage of global interchangeability.” or some such.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      ontology. Among the resources that to a smaller or greater extent can be used as inventories
+      of (more or less) standardized linguistic categories, are the GOLD ontology, <ref
+        target="https://www.clarin.eu/content/clarin-concept-registry">CLARIN CCR</ref>, <ref
+        target="https://github.com/acoli-repo/olia">OLiA</ref> or 
+      <ref target="https://datcatinfo.net/">TermWeb's DatCatInfo</ref>, and also the 
+      <ref target="https://universaldependencies.org/">Universal
+      Dependencies</ref> inventory, on the assumption that its URI are going to persist. Apart from these
+      global resources, it is imaginable to address a project-specific taxonomy as well, if global
+      interoperability is not the goal.</p>  
+    <p>Historically, <att>datcat</att> and <att>valueDatcat</att> originate from (now obsolete) ISO
```
sydb: Need article: insert “the” before “(now”.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      12620:2009 standard, describing the data model and procedures for a Data Category Registry
+      (DCR). The current version of the standard, ISO 12620-1, does not standardize the
```
sydb: I would use “current version of that standard”.

P5/Source/Specs/att.datcat.xml
```diff
@@ -7,51 +7,221 @@ $Date$
...
+      serialization of pointers, merely mentioning the TEI <ident type="class">att.datcat</ident> as
+      an example.</p> 
+    <p>Note that no constraint prevents the occurrence of a combination of att.datcat attributes:
```
sydb: `<ident type="class">`

sydb (changes requested):

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: Syd, you're being so delightfully Socratic here... ;-) This item relates to my doubts and musings in #2358 , where for a while I considered using `soCalled` together with `term`, but then realised that my reasons have to do with the rendering of these elements rather than their semantic import.

P5/Source/Specs/att.datcat.xml
bansp: Super, thanks.

P5/Source/Specs/att.datcat.xml
bansp: Definitely :-) Thanks.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: Ah, I just 'fixed' (or so I thought) it in passing, because it looked so bad in the chapter that I thought I should do something about it :-/
The was some bad indentation, and I just tried to make the result better. I can happily withdraw that.

P5/Source/Specs/att.datcat.xml
bansp: One tradition has been to capitalize them. I am not sure it would be accepted, so I looked for something that highlights them without scare quotes. I'll try them without any markup, then.

P5/Source/Specs/att.datcat.xml
bansp: My mental pronunciation started with "feature system ..." but, right, a wider group would just go for the letter-by-letter spelling.

P5/Source/Specs/att.datcat.xml
bansp: yes, I don't know what I was thinking when encoding that.

P5/Source/Specs/att.datcat.xml
bansp: Too many "it"s spoil it, I think. And the first of them is my fault. It is not the modeled feature that uses the attribute, but rather the modeling element. Tsk.

bansp: Changing to 
```xml
 Above, the <gi>fDecl</gi> uses <att>targetDatcat</att>, because if it were to use
        <att>datcat</att>, it would be asserting that it is an instance of the container data
      category <term rend="italic">part of speech</term>, whereas it is not — it models a container (<gi>f</gi>)
      that encodes a part of speech.
```
... while cringing a little bit.

P5/Source/Specs/att.datcat.xml
bansp: Done.

P5/Source/Specs/att.datcat.xml
bansp: I might scream. But then, maybe no-quotes is the best option, after all. I'm too quotey, I discover now.

P5/Source/Specs/att.datcat.xml
bansp: Modified, thanks.

P5/Source/Specs/att.datcat.xml
bansp: Removed :-)

P5/Source/Specs/att.datcat.xml
bansp: Also a term. But, true, no "disclaiming of responsibility" (I should have looked the spec up earlier, rather than going by the wording of so-called soCalled...)

P5/Source/Specs/att.datcat.xml
bansp: Right, I forgot the commas. Taking your version.

P5/Source/Specs/att.datcat.xml
bansp: OK

P5/Source/Specs/att.datcat.xml
bansp: comma added... :-)

P5/Source/Specs/att.datcat.xml
bansp: I was thinking the same, and then I thought that it would simply be a nice step forward to start by having it all in a single place, and then be able to speculate about whether it should also sneak into the text of the Guidelines and where.

P5/Source/Specs/att.datcat.xml
bansp: Handled before (caught my eye at some point). Thanks nevertheless.

P5/Source/Specs/att.datcat.xml
bansp: done

P5/Source/Specs/att.datcat.xml
bansp: ok

P5/Source/Specs/att.datcat.xml
bansp: yup :-)

P5/Source/Specs/att.datcat.xml
bansp: (and I meant a simple conditional here)

P5/Source/Specs/att.datcat.xml
bansp: done

P5/Source/Specs/att.datcat.xml
bansp: Tough choice :-) Followed your intuition.

P5/Source/Specs/att.datcat.xml
bansp: _semantic_ is an important modifier here as well. I'll just take the first option, this time.

P5/Source/Specs/att.datcat.xml
bansp: It's my command of English, I'm afraid. Removed, thanks.

P5/Source/Specs/att.datcat.xml
bansp: Changed to `<term rend="italic">`; I wonder.

bansp: this is following a probably idiosyncratic intuition or habit to use italics for mentioning terms and bold type for introducing terms with an accompanying definition-solidus-gloss...

bansp: and the `<hi>`s are just for contrastive focus.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: Sure thing, it's "or" here.

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: modified, thanks

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: Oh, so far, my impression was that you usually want to conserve space and let the formatter do some automatic work, except for the "feasible" cases, where it doesn't. Maybe I misinterpreted. Not sure what to do here. I shouldn't have touched it. Let me please leave it spoiled, and return to this example and others at some later point. 
Lesson learned: stay within your scope.  :-\

P5/Source/Guidelines/en/FS-FeatureStructures.xml
bansp: But that makes sense: first, there is a general statement on what can get aligned to reduce redundancy of individual feature representations. Then, a specific statement about the specific example mentions `fvLib` that is illustrated, and `taxonomy` that is not illustrated (but I felt it shouldn't not be mentioned).

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: replaced with `term`

P5/Source/Guidelines/en/DI-PrintDictionaries.xml
bansp: It would be. Describing this properly feels a bit over my head at this hour, however, so this may be the only one of your remarks that I will not implement in this round. Although it's so true that not implementing it feels a bit painful.

bansp: I have put this nice snippet into the datcat spec, to preserve it.

PR #2360, opened 2022-10-18
Address 2300 — no text inside `<facsimile>` (at least, not in `<line>` or `<zone>`) 

### LINKED ISSUE #2300: restrict embedded transcription to sourceDoc / exclude text in zone and line as part of facsimile
The specs state "`facsimile` contains a representation of some written source in the form of a set of images rather than as transcribed or encoded text." https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-facsimile.html
The current content model of `zone` and `line` is identical in `facsimile` and in `sourceDoc` which makes it possible for users to add text to these elements in a `facsimile` structure and by this contradicting the specs. Would it be an idea to align the schema with the specs by adding a schematron constraint on `(zone|line)[ancestor::facsimile]` exluding non-whitespace content?

sydb:
* Add a constraint (in Schematron) to require that there NOT be text content (other than whitespace-only text nodes, of course) inside 'zone' or 'line' elements that are themselves inside a 'facsimile' element.
 * Add a test that should fail that rule to detest.xml.
 * Change the expected results to match new outputs.

### DIFF
--- a/P5/Source/Specs/facsimile.xml
+++ b/P5/Source/Specs/facsimile.xml
@@ -7,7 +7,7 @@ $Date$
 $Id$
 -->
 <?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
-<elementSpec xmlns="http://www.tei-c.org/ns/1.0" ident="facsimile" module="transcr">
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0" ident="facsimile" module="transcr" xmlns:sch="http://purl.oclc.org/dsdl/schematron">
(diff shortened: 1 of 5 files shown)


raffazizzi (approved):

P5/Source/Specs/facsimile.xml
```diff
@@ -28,13 +28,35 @@ ou encodé.</desc>
...
     </sequence>
   </content>
+  <constraintSpec scheme="schematron" ident="no_facsimile_text_nodes">
+    <constraint>
+      <sch:rule context="tei:facsimile//tei:line | tei:facsimile//tei:zone">
+        <sch:report test="child::text()[ normalize-space(.) ne '']">
+  	  A facsimile element represents a text with images, thus
+	  transcribed text should not be present within it.
+	</sch:report>
+        <!--
```
martinascholger, line 45: I suggest to remove the comment. an issue has already been opened.

peterstadler: @martinascholger It's issue #2363, I believe? (for the record)

My 2 cents here are that the comment doesn't hurt, yet I think it's helpful and can be removed when issue #2363 is done

Issue #2314, opened 2022-06-19
TEI example is old-formatted


kzhr:
A TEI example cited below seems to be old-formatted and poorly displayed. I'm not sure what it should be.

https://tei-c.org/release/doc/tei-p5-doc/en/html/DI.html#index-egXML-d54e79590

A bit confusing is that this example lacks a source citation. If I remember correctly, it should have a source citation in early 2020, but I can't find recent changes around here.

sydb:
Good grief! This example is FUBAR. Need to dig through old versions of DI to see if a version that is _not_ messed up can be found. Will try to do that early this week.

(Thank you, @kzhr !)

sydb:
(DI-PrintDictionaries.xml line 3038.)

lb42:
It's not an example of TEI encoding, nor is it "FUBAR" ! (not an expression I would have expected from Syd ) It's an example of a non-TEI XML encoding which can't be transferred directly without using entryFree. That's what the surrounding text says anyway. If there;s anything to fix here, it's the formatting. Maybe wrapping this in a CDATA section would be a good idea?

lb42:
P.S. And there is a source citation too: but it's been given as a `<ref>` following the example instead of on the @source attribute

sydb:
Indeed, @lb42, good point. I just read the prose and came back here to say essentially the same thing, and found you beat me to it!

I think this example should probably be wrapped in an `<eg>` (rather than `<egXML>`), using the `@source` attribute as @lb42 points out. Using a CDATA marked section (as was the case before 2007-09-29) instead of entity references would be my preference, but someday some process will convert it to entity references, anyway. (Or so says the pessimist in me.)

PR #2362, opened 2022-10-20
Address #2314

sydb:
Per my suggestion on the [ticket](https://github.com/TEIC/TEI/issues/2314), changed example of non-XML encoding from `<egXML>` to `<eg>`, and moved citation to its `@source` attr.

Note: The little anchor that shows up for the “link-to-me” feature is the wrong character (i.e., is different than the one used for `<egXML>`s); and the “bibliography” link is not showing up. Both of these are Stylesheets problems, and neither one should prevent us from incorporating this change, IMHO.


martinascholger (approved):
Looks good to me

HelenaSabel (approved):

PR #2375, opened 2022-11-30
Update encoding of examples

### LINKED ISSUE #2225: missing space
While it is not _necessarily_ incorrect encoding, the example of `<cl>` in its tagdoc is missing a space between “so,” and “They”. It is not _necessarily_ incorrect because a project _could_ decide that, by default, `break="yes"` on `<cl>`. But that would be a bad thing to exemplify without explaining it. So either way this example needs to be updated. 

(Personally, I think the fix is just to insert a space before the `<cl>` start tag.)

sabineseifert:
Reorganized encoding of `cl` as the relative clause is only the first part of the sentence and not the whole sentence. This way, the encoding of `l` could also be added in the English example.
For the French example, reorganized encoding of `cl` as the relative clause and moved type and function attributes.
Could not check or update example in zh-TW due to lack of language skills.

### DIFF
--- a/P5/Source/Specs/cl.xml
+++ b/P5/Source/Specs/cl.xml
@@ -31,18 +31,18 @@ $Id$
   <content>
     <macroRef key="macro.phraseSeq"/>
   </content>
-  <exemplum xml:lang="en">
+  <exemplum versionDate="2022-11-30" xml:lang="en">
     <egXML xmlns="http://www.tei-c.org/ns/Examples">
-      <cl type="relative" function="clause_modifier">Which frightened
- both the heroes so,<cl>They quite forgot their quarrel.</cl>
-         </cl>
+      <l><cl type="relative" function="clause_modifier">Which frightened
+          both the heroes so,</cl></l>
+      <l><cl>They quite forgot their quarrel.</cl></l>
     </egXML>
   </exemplum>
-  <exemplum versionDate="2008-04-06" xml:lang="fr">
+  <exemplum versionDate="2022-11-30" xml:lang="fr">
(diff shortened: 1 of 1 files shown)


peterstadler:
pinging @HelenaSabel for review ;)

P5/Source/Specs/cl.xml
```diff
@@ -31,18 +31,18 @@ $Id$
...
   </exemplum>
-  <exemplum versionDate="2008-04-06" xml:lang="fr">
+  <exemplum versionDate="2022-11-30" xml:lang="fr">
     <egXML xmlns="http://www.tei-c.org/ns/Examples">
-      <cl type="relative" function="proposition_relative_déterminative"> Il nous rejoindra dans
-          les jours<cl>qui viennent.</cl>
-         </cl>
+      <cl>Il nous rejoindra dans
+          les jours</cl>
+      <cl type="relative" function="proposition_relative_déterminative">qui viennent.</cl>
```
HelenaSabel:
```suggestion
      <cl>Il nous rejoindra dans
          les jours
      <cl type="relative" function="proposition_relative_déterminative">qui viennent.</cl>
      </cl>
```

HelenaSabel (changes requested):
I added a suggestion concerning the French example: with a relative clause, I think the `<cl>` elements must be nested (it seems incorrect to me to have a clause segmentation that has the antecedent and the relative in two different clauses, but with no hierarchy).

peterstadler:
Thanks @HelenaSabel for the review! @sabineseifert  pulled in your suggestion and I think this is ready to be merged now

PR #2377, opened 2022-12-09
Changed titles to refs

### LINKED ISSUE #2368: ref= attribute of <title> element being ignored
The _Guidelines_ both support and in a few cases use an `@ref` on `<title>` to point to a bibliographic citation. But the Stylesheets just ignore it.

trishaoconnor:
Changed all title elements with ref attributes to ref elements with target attributes to fix issue #2368 with @sydb.


ebeshero (approved):
@sydb This all looks fine to me! @HelenaSabel I think you should just merge this in. Sorry for the delayed review!

PR #2384, opened 2022-12-21
Update example and bibliographic item

### LINKED ISSUE #2144: wrong attribution for example
The first example for @key in https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.canonical.html is attributed to 'No source, made up for these Guidelines.'  but the id number and name are suspiciously similar to http://nzetc.victoria.ac.nz/tm/scholarly/name-427308.html so I'm pretty sure the example comes from there.

@Conal-Tuohy may have some memory of how this came about; I certainly don't.

sabineseifert:
Updated the original example and added bibliographic reference for this example


sabineseifert:
Thank you, @sydb, for your revisions.  Can we now turn this draft PR to a proper PR for review by others?

sydb:
Yes, @sabineseifert, I think so. (And I am not sure I realized it was a draft. :-)

rettinghaus (commented):
Looks good! 

@sabineseifert Could you please rebase or merge latest develop to resolve the conflicts?

ebeshero:
We seem to have lost track of this PR, which should have been part of the 4.6.0 release--it just seems like it needed to an update to the branch. We'll get it in the next release.

PR #2385, opened 2022-12-28
Fixed `xml:lang` values in a summary for msPart examples for #2252

trishaoconnor:
Fixed the initial problem brought up by @MarjorieBurghart on ticket #2252. Updated the languages in the summary in msPart example as discussed on the ticket with @sydb. This pull request does not address the other issues highlighted in this ticket.

PR #2386, opened 2023-01-05
Fix examples of IETF language tags

adunning:
The text specifies that tags in `<textLang>` should follow IETF BCP 47, but several of these examples did not use codes from the IETF registry.

See <https://r12a.github.io/app-subtags/> for an enhanced version of the subtag registry.


sydb:
Excellent, thank you.
But the bad news is that PR #2385 already addresses 1 of these sets of errors. I am not at all sure what is the best way to proceed vis-vis git. I suspect that we can just pull in this PR and then when #2385 gets pulled in git will just do the right thing and merge the results, but I am not sure. So I am going to assign myself to review this, then ask about the git implications first before it is actually pulled.

sydb (approved):
I checked the IETF registry, @adunning’s corrections seem spot on to me.

sydb:
On advice of a few folks smarter than I, I went ahead and merged these changes into dev directly, then merged dev into trishaoconnor_2252.

adunning:
Thank you!

Issue #2298, opened 2022-05-14
description vs gloss in `@type` of `<divGen>`


sydb:
I think some of the `<desc>` elements inside `<valItem>`s in the definition of `@type` in the element specification of `<divGen>` (i.e., in P5/Source/Specs/divGen.xml) should be `<gloss>` elements instead.

lb42:
I think this is a systematic problem actually

sydb:
It strikes me as quite possible you are correct, @lb42. So …

I have just written a routine to make a table of descriptions and glosses. Seems to me any description that contains that which should have been a gloss will be pretty short. So the table includes a column for the approximate length of the description. Sorting the table by this column is somewhat instructive. At first glance it seems there definitely are other cases, but only a few dozen, not hundreds. I put a [copy of the output](https://bauman.zapto.org/~syd/temp/4TEICouncil/descs_and_glosses.html) on my basement server so folks can take a look.

sydb:
Note — I have updated the table to today’s dev version of the _Guidelines_, and @trishaoconnor and I plan to tackle this, probably starting next week.

PR #2388, opened 2023-01-23
Changed descriptions to glosses and removed underscores from glosses

trishaoconnor:
@sydb and I made small changes, changing descriptions to glosses as highlighted in #2298.


sabineseifert (approved):
Seems fine to me!
Just in 'domain', the values could be both gloss or description, the difference here is not that clear cut as in 'att.linguistic' and 'unclear'. But I'm sure you (@trishaoconnor and @sydb) talked this through and I am ok with the solution you found.

Issue #2279, opened 2022-04-17
Constrain translatable elements in TEI ODDs


hcayless:
To make it easier to build and maintain a translation interface for the TEI documentation, a few additional constraints would help:

1. Require that the English version of a desc, gloss, or remarks come first. This is a convention followed everywhere in the Specs as far as I can tell, but I don't think it is enforced. It is easier to look at preceding siblings for the "main" version than to have to look around.
2. Prevent multiple glosses and descs in the same language. I found a single case where a Japanese translation was added when an existing one was already there. There is already a rule like this in place for remarks.

sydb:
I submit that these are changes that should be made to p5odds.odd (where, BTW, the "only-one-remark-per-lang" constraint comes from), not to the _Guidelines_ themselves, in which case I am 100% in favor.

I get the duplicate in same language constraint, makes perfect sense. However, need to define “same language”. One (easy)  way to do so is to say “the values of `@xml:lang` match”, which I think makes sense. But if someone wants to make an argument that different variants or extensions should be considered the same, speak now …

While I have no objection to the “English first” constraint, I don’t get it — I thought translators were using an interface, not reading the tagdocs directly. In which case,
~~~xml
<xsl:apply-templates select="*[@xml:lang eq 'en']"/>
<xsl:apply-templates select="*[@xml:lang ne 'en']"/>
~~~
seems a pretty easy way to go.

hcayless:
@sydb imagine you have a desc in German, and you want to know if it's out of date. It's easier if you can just look back at the preceding English desc rather than have to look at both preceding siblings and following siblings.

And yes, I agree about where to make the changes.

sydb:
I have implemented this in branch sydb_2279. I found that there were 10 errors our new constraints found, and was able to fix 6 of them myself quite easily. HOWEVER we have trouble in this here paradise. In the following cases, there **is no** English `<desc>`, `<gloss>`, or `<remarks>`, thus it _cannot_ come first, and the constraint that the English must come first fails, and the build will break.

- the `<remarks>` for the element specification of `<extent>`
- the `<desc>` for the value item "foreign" of attribute definition for `@type` of `<q>`
- the `<remarks>` for the element specification of `<speaker>`

Also
 - the element specification for `<availability>` has two (slightly different) French glosses. (I would bet the 2021 version is to be preferred over the 2007 version, but as I do not speak French, and GoolgeTranslate says both mean “availability” …)

[...]

PR #2393, opened 2023-02-03
constrain att.translatable elements when used in the Guidelines

sydb:
This PR addresses #2279 by adding a couple of `<constraintSpec>`s to p5odds.odd, and then making some changes to the _Guidelines_ so that they remain valid even with the new constraint.

Note that the reviewers should particularly check the `<remarks xml:lang="en">` of speaker.xml.

BTW, I do not remotely think all 4 of you need to review this. One full-throated approval should be sufficient.


P5/Source/Specs/speaker.xml
```diff
@@ -88,12 +88,20 @@
       </list>
     </egXML>
   </exemplum>
+  <remarks xml:lang="en" versionDate="2023-01-13">
+    <p>This element may be used to transcribe which character is
+    speaking in a play as indicated by the source text; the
```
joeytakeda: Could "play" change to "dramatic text" (as per the `desc`)?

P5/Source/Specs/speaker.xml
```diff
@@ -88,12 +88,20 @@
...
+    <att>who</att> attribute of an <gi>sp</gi> element may be used to
+    point to another element (typically a <gi>castItem</gi>) which
```
joeytakeda: Specifying `<castItem>` seems slightly restrictive to me as not all dramatic texts have a defined `<castList>` and there are no examples in the GLs (as far as I can tell) of the `<castItem>` bearing an `@xml:id` or of a `sp/@who` pointing at one. Depending on what others think, I'd be inclined to say that the parenthetical here probably isn't necessary; if it is, then I'd suggest broadening to something like `typically a <gi>castItem</gi> or <gi>person</gi>`).

P5/p5odds.odd
```diff
@@ -496,17 +496,80 @@ $Id$
...
+                                 | tei:attDef/tei:remarks[1]
+                                 | tei:valItem/tei:desc[1]
+                                 | tei:valItem/tei:gloss[1]
+                                 | tei:valItem/tei:remarks[1]
+                                 | tei:moduleSpec/tei:desc[1]
+                                 | tei:moduleSpec/tei:gloss[1]
+                                 | tei:moduleSpec/tei:remarks[1]
+                                  ">
+                <sch:let name="lang" value="normalize-space( @xml:lang )"/>
+                <sch:assert test="matches( $lang, '^en-?' )">The first <sch:value-of select="name(.)"/> child of a <sch:value-of select="name(..)"/> should have an @xml:lang that indicates it is in English.</sch:assert>
```
joeytakeda: Since the above rule enforces that there must be an `@xml:lang`, could this be simplified to `lang('en')` (which should capture all variants)?

joeytakeda (changes requested):
A few notes in a few places, which are just minor queries (which of course can be taken or left as necessary!)

P5/Source/Specs/speaker.xml
sydb: Done. (Should arrive next push.)

P5/Source/Specs/speaker.xml
sydb: While I agree that you have noticed a problem, the solution I would prefer is to add an example of not just a `sp/@who` pointing at a `<castItem>`, but an example that demonstrates an encoding of `<sp>` pointing at a `<castItem>` for a cast member who did not appear in the source text.

P5/p5odds.odd
sydb: Well, yes, I suppose so. I have for some time now leaned away from using the lang() function because one of my processors (I do not remember which one, nor, more importantly, which version) used to choke on it. (In the sense that it _always_ returned false, not that it crashed.) But last I remember having that problem was years ago (before the pandemic started). So unless someone else pops up and says “yes, avoid lang()”, I think you are right, we should just go for it. (Especially given that we include a copy of Saxon in our Stylesheets repo, it matters far more what _that_ version of Saxon does than any other processor.)

sydb: As no one did pop up and say “avoid lang()”, I have now implemented @joeytakeda’s suggestion (in cc21f9a52).

P5/Source/Specs/speaker.xml
sydb: Hmmm … good catch, @joeytakeda. Turns out even though we _say_

> In the following example from Hamlet, speeches (`<sp>`) in the body of the play are linked to `<castItem>` elements in the `<castList>` using the `@who` attribute.

they are not. They are linked to the `<role>` element in the `<castItem>` (in the `<castList>`).

So I am planning to change “castItem” to “role” (there are 13 examples of that in the _Guidelines_) on this ticket; and raise separate tickets to consider the prose-encoding mis-match mentioned immediately above, and an example of referring to a speech from a character not listed in the cast list. If someone else would like to also raise a ticket suggesting that the parenthetical be dropped or include `<person>`, that is not unreasonable.

joeytakeda (approved):

P5/Source/Specs/speaker.xml
JanelleJenstad: See #2407

joeytakeda:
Per F2F: @sydb should bring this branch up to date and, assuming all is well, feel free to merge.

sydb:
Merging per Council F2F …

PR #2401, opened 2023-03-12
Fixing permissions in actions (should fix #2399) 

### LINKED ISSUE #2399: GitHub jobs failing due to permissions error
In GitHub, the test job fail when running the Tests from the Stylesheets repo due to a permissions error:

https://github.com/TEIC/TEI/actions/runs/4224874009/jobs/7336394765#step:6:72

From what I can tell, this is a fairly widespread issue for containers running in GitHub runners due to a recent debian security upgrade (see threads here: https://github.com/actions/runner-images/issues/6775 and https://github.com/actions/checkout/issues/1169) 

Following the advice from https://github.com/actions/checkout/issues/1169, I _think_ marking the directory as safe in Git after the [checkout step](https://github.com/TEIC/TEI/blob/b49da7f1651746181737b125a591621f7438cdc7/.github/workflows/test.yml#L22-L27) would work to solve the issue until `actions/checkout` implements a fix:

```yaml
      - name: Checkout Stylesheets repo
        uses: actions/checkout@v3
        with:
          repository: TEIC/Stylesheets
          ref: dev
          path: Stylesheets
      - name: Fix permissions
         run: git config --system --add safe.directory /github/workspace
```

joeytakeda:
* test.yml:
    * Add permissions config step to resolve 'dubious permissions' error
* generate-spec-lists.yml:
    * Upgrade action/checkout to v3
    * Add names and ws to the yaml (to align with test.yml)
 
Note that I haven't added the permissions step to generate-spec-lists.yml, as I don't think it will be necessary—however, I haven't been able to figure out a way to test for the dubious permissions error locally (or at least reliably), since I can't replicate the error (running the steps in a Docker container doesn't yield the error and neither does running the action using [`act`](https://github.com/nektos/act)).

### DIFF
--- a/.github/workflows/generate-spec-lists.yml
+++ b/.github/workflows/generate-spec-lists.yml
@@ -10,9 +10,17 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/generate-spec-lists
-      - uses: stefanzweifel/git-auto-commit-action@v4
+      - name: Checkout TEI Repository
+        uses: actions/checkout@v3
+        
+      - name: Fix permissions   
+        run: git config --global --add safe.directory /__w/TEI/TEI
+        
+      - name: Generate JSON Spec list        
+        uses: ./.github/actions/generate-spec-lists
+        
+      - name: Auto-commit Spec list  
+        uses: stefanzweifel/git-auto-commit-action@v4
         with:
           commit_user_name: ${{ github.actor }}
           commit_message: Re-generated spec lists.
--- a/.github/workflows/generate-spec-lists.yml
+++ b/.github/workflows/generate-spec-lists.yml
@@ -12,13 +12,10 @@ jobs:
     steps:
       - name: Checkout TEI Repository
         uses: actions/checkout@v3
-        
-      - name: Fix permissions   
-        run: git config --global --add safe.directory /__w/TEI/TEI
-        
+
       - name: Generate JSON Spec list        
         uses: ./.github/actions/generate-spec-lists
-        
+
       - name: Auto-commit Spec list  
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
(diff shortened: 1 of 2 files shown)


ebeshero:
@joeytakeda I merged this branch into my failing branch, and indeed, it worked so that the tests could complete running: https://github.com/TEIC/TEI/actions/runs/4411611402/jobs/7730289074 

So I think this is a helpful repair to merge in to dev--and should indeed repair what's broken from the most recent commit there. @hcayless or @raffazizzi any issues with my merging to dev as requested?

raffazizzi (approved):

raffazizzi:
@joeytakeda this looks fine to me and the checks pass. 

I was a little concerned that the  `git config --global --add safe.directory /__w/TEI/TEI` may fail if the directory name were to change in the future, but I tested adding a non-existent safe.directory locally and it doesn't seem to be an issue. If the directory name *does* change in the future, I figure the future council will have to update the path.

I would maybe remove the comment ` # Temporary fix to deal with #2399` before merging.

hcayless (approved):

hcayless:
Would `git config --global --add safe.directory "$GITHUB_WORKSPACE"` work instead? and avoid the hard-coded directory path?

joeytakeda:
Thanks @hcayless @raffazizzi! 

> I was a little concerned that the git config --global --add safe.directory /__w/TEI/TEI may fail if the directory name were to change in the future, but I tested adding a non-existent safe.directory locally and it doesn't seem to be an issue. If the directory name does change in the future, I figure the future council will have to update the path.

> Would git config --global --add safe.directory "$GITHUB_WORKSPACE" work instead? and avoid the hard-coded directory path?

Good point — I'll change that now.

> I would maybe remove the comment  # Temporary fix to deal with #2399 before merging.

I added that comment so we can flag it for removal at some point (since this is really a bug with the action/checkout@v3 that I would hope would be solved in a future iteration) — but it's not a very helpful comment, so I will adjust to try and make it a bit more useful

Issue #1773, opened 2018-05-24
Improve biblStruct examples


jamescummings:
c.f. https://listserv.brown.edu/archives/cgi-bin/wa?A2=TEI-L;7d360e0c.1805 

which suggests improving our biblStruct examples.

Putting a ticket it to remind us to do so with the suggestions provided, but lower urgency than some other issues.

martinascholger:
Good examples using specific attributes. How about choosing two or three and add a note which type of publication is described (manuscript, article in a volume/journal, etc.)?

vbigot-juloux:
@jamescummings: Chuck only looked to `biblStruct` element page and not the examples listed. Then, maybe an additional question should be: is the “Show all” link is enough visible? Maybe a note needs to be added before the example like “For further examples, click on ‘Show all’ link” or something like that.

There are indeed already a lot of good examples using core elements e.g. `analytic` `series`, `monogr` (c.f. http://www.tei-c.org/release/doc/tei-p5-doc/en/html/examples-biblStruct.html). As @martinascholger wrote, it needs only examples with specific attributes--currently there are 2 examples of attributes within `biblStruct`: `<biblStruct type="patent">` and `<biblStruct type="book">`--but no note added yet.  Thus, I suggest completing the current examples with additional attributes.
However, we can also add one example with one `@type` on the `biblStruct` element page.

trishaoconnor:
@sydb and I were reviewing the examples for inclusion on the `<biblStruct>` element page. We like the example under [[3.12.2.1 Analytic, Monographic, and Series Levels](https://tei-c.org/release/doc/tei-p5-doc/en/html/CO.html#COBICOL)]. We think it would be a good example to include on the `<biblStruct>` element page if it were re-encoded using `<relatedItem>`. 

We would like to consult a German speaker (@martinascholger?) to advise on how best to re-encode the below example using `<relatedItem>`.  

```
<biblStruct>
 <analytic>
  <author>
   <persName>
    <surname>Thaller</surname>
    <forename>Manfred</forename>
   </persName>
  </author>
  <title level="a">A Draft Proposal for a Standard for the
     Coding of Machine Readable Sources</title>
 </analytic>
 <monogr>
  <title level="j">Historical Social Research</title>
  <imprint>
   <date when="1986-10">October 1986</date>
  </imprint>
  <biblScope unit="volume">40</biblScope>
  <biblScope unit="page" from="3" to="46">3-46</biblScope>
 </monogr>
[...]

PR #2405, opened 2023-03-21
Added new example of the <biblStruct> element with a type attribute #1773

trishaoconnor:
@sydb and I added a new example of the `<biblStruct>` element with a type attribute for #1773.


HelenaSabel (approved):

P5/Source/Specs/biblStruct.xml
```diff
@@ -68,6 +68,39 @@
...
+          <author>
+            <forename>Susanne</forename>
+            <surname>Craig</surname>
+          </author>
+          <author>
+            <forename>Russ</forename>
+            <surname>Buettner</surname>
+          </author>
+          <title type="main">Trump Took Part in Suspect Schemes to Evade Tax Bills</title>
+          <title type="sub">Behind the Myth of a Self-Made Billionaire, a Vast Inheritance From His Father.</title>
```
joeytakeda: Should the terminal period here be deleted? (I tend to think that titles should not have concluding punctuation in a biblStruct, but I can't find the print edition to see if there's actually a period at the end of the subtitle

joeytakeda (commented):
Thanks, @trishaoconnor ! Just one small query re: trailing period on the subtitle and then it seems good to merge to me

P5/Source/Specs/biblStruct.xml
trishaoconnor: Thanks, @joeytakeda! Well spotted. I missed that. @sydb kindly took pictures of the print article for me and the subtitle does not have concluding punctuation. I will update that now.

Issue #2396, opened 2023-02-11
make `<calendar>` a member of att.typed?


sydb:
The `<calendar>` element is (obviously) repeatable. And I, for one, can imagine categorizations thereof.
The first that jumps to mind is "fictional" vs "actual" (or some such). But my current use case would be dividing calendar systems up into those that can easily be converted to (proleptic) Gregorian, those that cannot be converted to (proleptic) Gregorian, and those that _might_ be convertable to (proleptic) Gregorian.
I cannot, at the moment, see any reason that `@type` should not be allowed on `<calendar>`.

trishaoconnor:
The att.typed membership is merged into dev. 

Council suggests adding a description in the prose to clarify the purpose of `@type` in `<calendar>`.
@sydb proposes the following description for inclusion into the prose:

```
A <gi>calendar</gi> element may also have a <att>type</att> (and
  <att>subtype</att>) attribute in order to categorize the calendar in
  any convenient way. In the following example, <att>type</att> is
  used to differentiate real calendar systems in actual use from those
  used in a fictional environment.
```

sydb:
I not only suggested it (at @martinascholger’s suggestion on the PR), I committed it to the branch. Thus that prose is already in dev.

PR #2406, opened 2023-03-21
Added `@type` to `<calendar>` and added an example of `@type` to prose `#HD44` for #2396

trishaoconnor:
Added `@type` to `<calendar>` to specify between real or actual and fictional calendar as per the example currently listed in the prose description for `<calendar>` in the Header (`#HD44'). Also added `@type` to the '<calendar>` example in the prose description in the Header `#HD44` for #2396. 

Needs to be reviewed by Council because this issue has not been discussed in detail prior to pushing this pull request. @jamescummings and @martindholmes any thoughts on adding `@type` to `<calendar>`?

### DIFF
--- a/P5/Source/Guidelines/en/HD-Header.xml
+++ b/P5/Source/Guidelines/en/HD-Header.xml
@@ -2339,8 +2339,8 @@ either the <att>calendar</att> attribute on <gi>date</gi> or the
   code for it as the value of its <att>xml:id</att> attribute.
     <egXML xmlns="http://www.tei-c.org/ns/Examples">
       <calendarDesc>
-        <calendar xml:id="Gregorian"><p>Gregorian calendar</p></calendar>
-        <calendar xml:id="Stardate"><p>Fictional Stardate (from Star Trek series)</p></calendar>
+        <calendar xml:id="Gregorian" type="actual"><p>Gregorian calendar</p></calendar>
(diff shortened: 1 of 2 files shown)


martindholmes (approved):
The PR looks fine, but as you say, the ticket itself has had little or no discussion, so Council should all take a look before merging. For the record, I am in favour of it myself.

sydb:
While I do not remotely think _all_ members of Council need weigh in, I do want at least one, preferably two, more positive reviews before merging. Thus pinging @jamescummings, @sabineseifert, and @martinascholger.

martinascholger (approved):
Looks good to me. I just wonder if there should be a sentence in HD saying something like "The type attribute could be used to [specify|distinguish] between real or actual and fictitious calendar“.

sydb:
My instinct is “yes”, updating the prose to discuss it a little is a good idea. However, we can do that next week. The att.typed membership needs to be merged today or tomorrow.
That said, how about
~~~xml
  A <gi>calendar</gi> element may also have a <att>type</att> (and
  <att>subtype</att>) attribute in order to categorize the calendar in
  any convenient way. In the following example, <att>type</att> is
  used to differentiate real calendar systems in actual use from those
  used in a fictional environment.
~~~
?

jamescummings:
All looks good to me for the record.

jamescummings (approved):
Looks good. 👍

sabineseifert (approved):
Looks good to me as well, also the addition in the prose description.

PR #2410, opened 2023-03-30
“Deprecation” message: <content> gets 1 and only 1 child

sydb:
Here is what I did:

- Created Schematron constraint, as a separate file for easy testing
- Ran it on all .xml, .tei, .odd, and .plodd files in the ATOP repository (there are a lot of ODD files there to be used as tests)
- Found that it seemed to work perfectly — it found lots of cases of 2+ children of `<content>` that were either all TEI elements or all RELAX NG elements; also found 2 cases in which there was a mix. I spot-checked 2–3 of each of the former cases and the 2 mixes, too.
- Moved the Schematron into P5/Sourece/Specs/content.xml in `<constraintSpec ident="only_1_child_of_content">`.
- That constraint needs the RELAX NG annotation namespace to be declared, so added an `<sch:ns>` to TEI.xml (in "c2"), binding the namespace to the prefix “rna:”.
- Built the _Guidelines_ with `make clean validate html-web test exemplars` in the Docker environment.
- Found 1 failure — 2 examples in USE had 2+ children of `<content>`.
- Fixed those errors. (Also tweaked some whitespace in sp.xml while I was at it.)
- Built again, no errors.

Here is what the warning messages look like (whitespace altered for readability) …

2+ TEI children:
~~~
In the near future the <content> element will be limited to 1 and
only 1 child element. This <content> element is in an elementSpec
with an @ident of "constraintSpec" and has 2 children, which
could be wrapped in a <sequence> element.
~~~
 
2+ RELAX NG children:
~~~
In the near future the <content> element will be limited to 1 and
only 1 child element. This <content> element is in a moduleRef
with a @url of "http://www.w3.org/Math/RelaxNG/mathml3/mathml3.rng"
and has 2 children, which could be wrapped in an <rng:div> element.
~~~

2+ children, not all TEI, not all RELAX NG:
~~~
In the near future the <content> element will be limited to 1 and
only 1 child element. This <content> element is in an elementSpec
with an @ident of "publicationStmt" and has 3 children, but those
children are neither all TEI elements nor all RELAX NG elements,
and thus this <content> is invalid and can not be easily rectified.
~~~

What I did *not* do was use role="warning" on the `<sch:assert>` and `<sch:report>` elements. I probably should have, and may well fix that and commit shortly.

### DIFF
--- a/P5/Source/Specs/content.xml
+++ b/P5/Source/Specs/content.xml
@@ -54,13 +54,13 @@ $Id$
       <sch:let name="msg_part09" value="../@ident | ../@key | ../@url"/>
       <sch:let name="msg_part10" value="concat('&quot; and has ', $tot_kids, ' children,')"/>
       <sch:let name="msgs_1to10" value="concat( $msg_part01, $msg_part02, $msg_part03, $msg_part04, $msg_part05, $msg_part06, $msg_part07, $msg_part08, $msg_part09, $msg_part10 )"></sch:let>
-      <sch:report test="$tei_kids eq $tot_kids">
+      <sch:report test="$tei_kids eq $tot_kids" role="warning">
         <sch:value-of select="$msgs_1to10"/> which could be wrapped in a &lt;sequence> element.
       </sch:report>
-      <sch:report test="$rng_kids eq $tot_kids">
+      <sch:report test="$rng_kids eq $tot_kids" role="warning">
         <sch:value-of select="$msgs_1to10"/> which could be wrapped in an &lt;rng:div> element.
       </sch:report>
-      <sch:assert test="$tei_kids eq $tot_kids  or  $rng_kids eq $tot_kids">
+      <sch:assert test="$tei_kids eq $tot_kids  or  $rng_kids eq $tot_kids"   role="warning">
         <sch:value-of select="$msgs_1to10"/> but those children are neither all TEI elements nor
         all RELAX NG elements, and thus this &lt;content> is invalid and can not be easily rectified.
       </sch:assert>
--- a/P5/Source/Specs/content.xml
+++ b/P5/Source/Specs/content.xml
@@ -34,6 +34,39 @@ $Id$
       <anyElement minOccurs="1" maxOccurs="unbounded" require="http://relaxng.org/ns/compatibility/annotations/1.0 http://relaxng.org/ns/structure/1.0"/>
     </alternate>
   </content>
+  <constraintSpec ident="only_1_child_of_content" scheme="schematron">
+    <desc>A temporary constraint to give users a warning that in the
+    future the content of <gi>content</gi> will be restricted to 1 and
+    only 1 child element.</desc>
+    <constraint>
+    <sch:rule context="tei:content[ *[2] ]">
+      <sch:let name="tot_kids" value="count( * )"/>
+      <sch:let name="tei_kids" value="count( tei:* )"/>
+      <sch:let name="rng_kids" value="count( rng:* | rna:* )"/>
+      <sch:let name="msg_part01" value="'In the near future the &lt;content> element will be limited to 1 and only 1 child element. '"/>
+      <sch:let name="msg_part02" value="'This &lt;content> element is in '"/>
+      <sch:let name="msg_part03" value="if ( local-name(..) eq 'elementSpec' ) then 'an ' else 'a '"/>
+      <sch:let name="msg_part04" value="concat( local-name(..), ' with ' )"/>
+      <sch:let name="msg_part05" value="if ( parent::tei:moduleRef/@key ) then 'a @key'    else ''"/>
+      <sch:let name="msg_part06" value="if ( parent::tei:moduleRef/@url ) then 'a @url'    else ''"/>
+      <sch:let name="msg_part07" value="if ( parent::tei:*/@ident )       then 'an @ident' else ''"/>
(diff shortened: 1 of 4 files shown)


sydb:
More:

- Added role="warning"
- Improved XSLT to fix — added error checking so it only works iff all children are either TEI or RELAX NG (see below)
- Ran that XSLT on all the files I used for testing, above, expecting error msgs for the mixed two, and got them
  - Note: I flattened the hierarchy, so due to name collisions only 413 of the 644 files actually used 
- Spot-checked 4 of the output files, they seemed correct
- Validated all those fixed files with the Schematron — I should only get errors where the content of `<content>` was mixed, and that’s what I got

~~~xml
<?xml version="1.0" encoding="UTF-8"?>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
  xmlns="http://www.tei-c.org/ns/1.0"
  xmlns:tei="http://www.tei-c.org/ns/1.0"
  xmlns:rng="http://relaxng.org/ns/structure/1.0"
  xmlns:rna="http://relaxng.org/ns/compatibility/annotations/1.0"
  xpath-default-namespace="http://www.tei-c.org/ns/1.0"
  exclude-result-prefixes="#all"
  version="3.0">
  
  <xsl:output method="xml" indent="yes"/>
  <xsl:mode on-no-match="shallow-copy"/>
  
  <xsl:template match="content[ *[2] ]" as="element(content)">
    <xsl:variable name="wrapperGI">
      <xsl:choose>
        <xsl:when test="tei:* and not( rng:* | rna:* )">sequence</xsl:when>
        <xsl:when test="( rng:* | rna:* ) and not( tei:* )">rng:div</xsl:when>
        <xsl:otherwise>
          <xsl:sequence select="'HUMAN_FIX_THIS'"/>
          <xsl:message>ERROR: content element with children that are neither all TEI nor all RELAX NG; contents will be wrapped in an invalid &lt;HUMAN_FIX_THIS> element.</xsl:message>
        </xsl:otherwise>
      </xsl:choose>
    </xsl:variable>
    <xsl:copy>
      <xsl:apply-templates select="@*"/>
      <xsl:element name="{$wrapperGI}">
        <xsl:apply-templates select="node()"/>
      </xsl:element>
    </xsl:copy>
  </xsl:template>
  
</xsl:stylesheet>
~~~

P5/Source/Specs/content.xml
```diff
@@ -34,6 +34,39 @@ $Id$
...
+      <sch:let name="msg_part07" value="if ( parent::tei:*/@ident )       then 'an @ident' else ''"/>
+      <sch:let name="msg_part08" value="' of &quot;'"/>
+      <sch:let name="msg_part09" value="../@ident | ../@key | ../@url"/>
+      <sch:let name="msg_part10" value="concat('&quot; and has ', $tot_kids, ' children,')"/>
+      <sch:let name="msgs_1to10" value="concat( $msg_part01, $msg_part02, $msg_part03, $msg_part04, $msg_part05, $msg_part06, $msg_part07, $msg_part08, $msg_part09, $msg_part10 )"></sch:let>
+      <sch:report test="$tei_kids eq $tot_kids" role="warning">
+        <sch:value-of select="$msgs_1to10"/> which could be wrapped in a &lt;sequence> element.
+      </sch:report>
+      <sch:report test="$rng_kids eq $tot_kids" role="warning">
+        <sch:value-of select="$msgs_1to10"/> which could be wrapped in an &lt;rng:div> element.
```
ebeshero, line 61: @sydb This is clever how you structured the message depending on parent elements! I think you covered all the possible parents, yes?

ebeshero (approved):
@sydb This looks ready to merge to me. I am impressed with your customized Schematron messaging!

P5/Source/Specs/content.xml
sydb: Yup, `<moduleRef>` might have a `@key` or a `@url`, but `<dataSpec>`, `<elementSpec>`, and `<macroSpec>` all have a required `@ident`.

P5/Source/Specs/content.xml
```diff
@@ -34,6 +34,39 @@ $Id$
...
+      </sch:report>
+      <sch:assert test="$tei_kids eq $tot_kids  or  $rng_kids eq $tot_kids"   role="warning">
```
hcayless, line 63: I may just not be understanding the logic here, but shouldn't the test be something like `$tei_kids gt 0 and $rng_kids gt 0`? Or maybe I'm missing something. Is the point that you don't want both PureODD and RNG children of content, yes?

ebeshero: @hcayless I think he's just filtering the messages based on whether the multiple children are RNG or TEI. The point is to issue a customized warning depending on the situation that soon we'll only be permitting one and only one child of `<content>`. So I'm happy with it.

sydb: I am reasonably confident that
~~~xml
<sch:assert test="$tei_kids eq $tot_kids  or  $rng_kids eq $tot_kids"
~~~
tests exactly the same conditions that
~~~xml
<sch:report test="$tei_kids gt 0  and  $rng_kids gt 0">
~~~
would.
In truth, given that your formulation (@hcayless) uses `<report>`, not `<assert>`, and the other constraints are expressed with `<report>` (and given that it does not require a `not()`), I kinda like your formulation a bit better. But I wouldn’t bother changing it at this stage. We only expect this constraint to be around for a year at most, anyway.

Issue #2411, opened 2023-04-04
newly deprecated practice still extant … <content>s with 2 kids


sydb:
@martindholmes discovered (perhaps inadvertently) that Exemplars/tei_its.odd and Exemplars/tei_customization.odd both have a case of a `<content>` with 2 child elements, which we now warn about.

sydb:
Note — per comment on previous PR associated with this ticket, it is remaining open so that the following files can be “repaired”.
~~~
 Test/test-newodd.xml: 10
 Test/testdocbook3.odd: 1
 Test/testdocbook2.odd: 1
 Test/testnym.odd: 1
 Test/testdocbook1.odd: 1
 Test/testdocbook5.odd: 1
 Test/testdocbook4.odd: 3
~~~
(The numbers are how many `<content>` elements have 2+ children.)

PR #2412, opened 2023-04-04
abide by “`<content>` has 1 and only 1 child” rule, step 1

sydb:
This PR addresses #2411 by wrapping the content of `<content>` elements that have 2+ child elements in either `<rng:div>` or `<sequence>` in Exemplars and a Test/ file that is used by `make test`.

NOTE:
There are still 7 files in Test/ that have 1 or more `<content>` elements that have 2 or more child elements:
~~~
 Test/test-newodd.xml: 10
 Test/testdocbook3.odd: 1
 Test/testdocbook2.odd: 1
 Test/testnym.odd: 1
 Test/testdocbook1.odd: 1
 Test/testdocbook5.odd: 1
 Test/testdocbook4.odd: 3
~~~
However, it seems these are not executed by `make test`, and thus will not affect the build. Thus, **for now** I have not changed them. If & when this PR is merged & closed _the ticket should remain open_ so that those other files can be repaired.

### DIFF
--- a/P5/Exemplars/tei_its.odd
+++ b/P5/Exemplars/tei_its.odd
@@ -69,12 +69,14 @@
       <moduleRef key="textstructure"/>
       <moduleRef url="https://www.tei-c.org/release/xml/tei/Exemplars/its.rng">
       	<content>
-	  <rng:define name="tei_model.teiHeaderPart" combine="choice">
-	    <rng:ref name="its_rules"/>
-	  </rng:define>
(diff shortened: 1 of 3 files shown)


ebeshero (approved):
This looks all correct to me based on our discussions--it corrects the problem in our files and announces the deprecation.

martindholmes (approved):
Looks good.

PR #2416, opened 2023-04-13
guidelines.xsl generation update

sydb:
I got caught once again by the fact that P5/Utilities/guidelines.xsl is a generated file. (And in front of my mentee, making it even more embarrassing.) So the main point of this PR is to make sure there is a very very obvious warning in that file that says it is a derived file. But while I was at it, I converted the generation mechanism to XSLT.

Various changes include:
 * Use new XSLT pgm guidelines_model_2_executable.xslt instead of using Perl (to change paths to imported files)
 * Add XML declaration to output
 * Allow XSLT processor to indent output (but not for guidelines-latex, as I do not know enough LaTeX to know that is OK)
 * Use XSLT 3
 * Add warning comments in derived output
 * Correct lots of mistyped "UTF-8"s

### DIFF
--- a/P5/Makefile
+++ b/P5/Makefile
@@ -78,10 +78,7 @@ html-web.stamp:  check.stamp p5.xml  Utilities/guidelines.xsl.model
 	@# remove vestiges of previous run
 	rm -rf Guidelines-web
 	@# generate XSL stylesheet by modifying paths in our model stylesheet
-	perl -p -e \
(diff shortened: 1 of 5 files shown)


sydb:
BTW, I have run `make clean validate html-web test exemplars teiwebsiteguidelines pdf epub` on this branch in a Docker environment, and it all worked.

raffazizzi:
F2F@Guelph subgroup: this looks good and @joeytakeda can go ahead and merge

martinascholger (approved):

PR #2418, opened 2023-04-17
(a) French version of tutorial (b) minor typos in release notes

lb42:


Documents/pureODD/chainingTuto/metadata.xml
```diff
@@ -11,6 +11,7 @@
             <p>Part of TEI Chaining tutorial</p>
         </sourceDesc>
     </fileDesc>
+    <xenoData xmlns:dc="dublineCore"><dc:wibble/></xenoData>
```
raffazizzi, line 14: If this is just an example of xenoData, could we use a real Dublin Core element, with the appropriate namespace?

Documents/pureODD/chainingTuto/motherODD.xml
```diff
@@ -24,11 +25,10 @@
         
 <div type="schemaSpec">
             <head>The Mother ODD specification</head>
```
raffazizzi, line 27: This needs translation.

Documents/pureODD/howtoGenerate-fr.xml
```diff
@@ -376,7 +376,7 @@
                particulièrement utiles dans le processus de définition d’un schéma utile pour un
                ensemble de documents TEI. </p>
          </div>
-         <div>
+<!--         <div>
```
raffazizzi, line 379: Could you just remove this commented out block?

raffazizzi (changes requested):

Documents/pureODD/chainingTuto/metadata.xml
lb42: I think this was an accidental change, so it could be removed. Or, as you suggest:  change it to 
```
<xenoData xmlns:dc="http://dublincore.org/documents/dcmi-namespace/">
<dc:author>Jules Verne</dc:author>
</xenoData> 
```

Documents/pureODD/howtoGenerate-fr.xml
lb42: It's gone.

Documents/pureODD/chainingTuto/motherODD.xml
lb42: I dont think so. It's not used by either of the two French tutorials.

Issue #2214, opened 2021-12-01
sequence of top level elements within msDesc


schassan:
The general structure of a manuscript description following the German cataloguing rules puts the description of the contents at the end, right after physDesc, history, and literature and before the msParts. The schema declaration (snippet) would look like this:

(
     msIdentifier,
     model.headLike*,
     (
        model.pLike+
      | (
           physDesc?,
           history?,
           additional?,
           msContents?,
           ( msPart* | msFrag* )
        )
     )
  )

A brief discussion on TEI-MS-SIG has shown that in other regions than only in German speaking countries different order of elements is preferable too.

In the past we dealt with this problem by just changing the sequence of contents in the transformations while preparing the output of the descriptions, on the web and/or for print. But in the future we will have to change the file structure itself as we are about to implement an editor which will save the file "as shown" (WYSIWYG, the hard way). :(

I guess, in order to allow for the current structure *and* the German one, we need to loose the rules for the sequence or define the latter as another option?

It would be necessary to keep the limit of occurences of physDesc, history, additional, and msContents at One though.

peterstadler:
I'm all in favor for loosening this sequential constraint 👍

lb42:
I seem to recall that the last time this was proposed was one of the rare occasions in which members of the ms description workgroup almost came to blows.

sydb:
## An Analysis Expressed in (mostly) Formal Syntax
Here is my take on the problem @schassan is expressing and the various solutions. If anyone knows how to get GitHub to syntax-color the code block of RELAX NG Compact Syntax below, please let me know (or just edit this post). [JC: there isn't one for rnc highlighting, but you could do something else like js. ;-) ]
~~~js
start = element msDescs_for_2214 { msDesc_current, msDesc_schassan, msDesc_2214_alt, msDesc_2214_fac, msDesc_2214_dtd, msDesc_2214_mso }

# ---------------------------------------------

# Current content model in P5
msDesc_current =
  element msDesc {
    msIdentifier,
    model.headLike*,
    (
       model.pLike+
       |
       ( msContents?, physDesc?, history?, additional?, (msPart* | msFrag*) )
    )
  }

# What @schassan wants to use
msDesc_schassan =
  element msDesc {
    msIdentifier,
    model.headLike*,
    (
       model.pLike+
       |
       ( physDesc?, history?, additional?, msContents?, (msPart* | msFrag*) )
    )
  }

# Simple alternation of the two
msDesc_2214_alt =
[...]

PR #2423, opened 2023-04-28
changes the content model for msDesc, msPart, and msFrag in order to …

schassan:
…allow for any order of child elements, after msIdentifier and head

This pull request answers to issue #2214, allowing an even more flexible order of child elements of msDesc, msPart, and msFrag, as in the original issue. This turned out to be necessary within the Handschriftenportal, but as well was favoured by the MS-SIG in it's 2022 SIG meeting in Newcastle.

### DIFF
--- a/P5/Source/Specs/msDesc.xml
+++ b/P5/Source/Specs/msDesc.xml
@@ -36,7 +36,7 @@ $Id$
       <classRef key="model.headLike" minOccurs="0" maxOccurs="unbounded"/>
       <alternate>
         <classRef key="model.pLike" minOccurs="1" maxOccurs="unbounded"/>
-        <sequence>
+        <alternate>
           <elementRef key="msContents" minOccurs="0"/>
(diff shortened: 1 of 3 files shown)


sydb (changes requested):
I do not have time to actually test right now, but on first glance, @schassan , I do not think this content model is going to what you want. I _think_ this content model will allow only 1 of the elements `<msContents>` OR `<physDesc>` OR `<history>` OR `<additional>` to occur. But there are _lots_ of cases where a manuscript description should have both, say, `<physDesc>` and `<history>`.
I will try to poke at this more carefully (including experimenting with `@preserveOrder`) in the next week.

sydb:
@schassan and MS-SIG — see [today’s comment](https://github.com/TEIC/TEI/issues/2214#issuecomment-1528886927) on #2214.

sydb:
#2495 addressed this, so closing.

Issue #2381, opened 2022-12-14
`<content>` should have 1 and only 1 child


sydb:
I am raising this ticket here (at request of ATOP task force) for discussion of the issue.

The _Guidelines_ do not actually say what a series of children of `<content>` _means_. We all know (and the Stylesheets behave as though) a series of child elements of `<content>` means the same thing as the same set of child elements inside a `<sequence>`, itself the only child of the `<content>`. That is,
~~~xml
  <content>
    <elementRef key="a"/>
    <elementRef key="b"/>
    <elementRef key="c"/>
    <alternate>
      <elementRef key="x"/>
      <elementRef key="y"/>
      <elementRef key="z"/>
    </alternate>
  </content>
~~~
means exactly the same as
~~~xml
  <content>
    <sequence>                  <!-- ← explict sequence 😌 -->
      <elementRef key="a"/>
      <elementRef key="b"/>
      <elementRef key="c"/>
      <alternate>
        <elementRef key="x"/>
        <elementRef key="y"/>
        <elementRef key="z"/>
      </alternate>
    </sequence>
  </content>
~~~
The former is a bit easier to write; the latter is explicit and gives a place for explicit quantification (`@minOccurs` & `@maxOccurs`). More importantly, the former utterly fails when generating DTDs, you have to use the latter.

The content model for `<content>` is currently
~~~xml
  <content>
    <alternate>
      <elementRef key="valList" minOccurs="1" maxOccurs="1"/>
      <classRef key="model.contentPart" minOccurs="1" maxOccurs="unbounded"/>
      <anyElement minOccurs="1" maxOccurs="unbounded" require="http://relaxng.org/ns/compatibility/annotations/1.0 http://relaxng.org/ns/structure/1.0"/>
    </alternate>
  </content>
~~~
We were thinking that perhaps it should be
~~~xml
  <content>
    <alternate>
      <elementRef key="valList" minOccurs="1" maxOccurs="1"/>
      <classRef key="model.contentPart" minOccurs="1" maxOccurs="1"/>
      <anyElement minOccurs="1" maxOccurs="unbounded" require="http://relaxng.org/ns/compatibility/annotations/1.0 http://relaxng.org/ns/structure/1.0"/>
    </alternate>
  </content>
~~~
[...]

PR #2428, opened 2023-05-06
more on 1 kid for content

sydb:
More on #2411 (from #2381): Wrap contents of content elements that have > 1 child in rng:div


sydb:
I have just merged latest from 'dev' branch into this one. Could we merge this in pretty soon? It seems like a pretty simple update.

hcayless:
Looks good to me.

PR #2429, opened 2023-05-07
Address 1710 by changing module= values

sydb:
NOTE — This PR is here for @raffazizzi & @martinascholger to use as their basis for working on the ticket. (That is, I am not really asking that this be merged into dev as it is untested.)Just automatically changed based on table on today’s comment on the ticket (#1710).
(That is, I did the easy part, I have _not_ gone through the prose to see if there is any discussion that needs updating. I bet there is very little, but there could be some.)


raffazizzi:
F2F@Guelph #1710 is blocked for now so we are closing this.

Issue #1860, opened 2019-03-03
Encoding RDF relationships in TEI (TEI+RDFa and alternatives)


chiarcos:
Motivation is to achieve a representation of RDF relations in the TEI which is unambiguous in vocabulary and semantics. Note that this does not pertain to cases where native TEI vocabulary elements could be interpreted as triples, but to cases that are not covered by TEI semantics, e.g., the linking between a passage in a edition and a terminology repository or a CTS urn. A similar restriction can be found in the definition of &lt;link>.

At the moment, there are at least three different possibilities to express RDF triples inline in TEI:
&lt;relation> (#311)
&lt;fs>
&lt;link>

Each of these are problematic as they conflate pre-RDF and RDF semantics, and that they are analogy-driven ("tag abuse") rather than explicitly defined. The currently preferred solution with &lt;relation> is restricted to named entities, example 4 in the guidelines thus breaks the TEI schema (see my comment on #311).

Several alternatives are possible (see email thread in http://tei-l.970651.n3.nabble.com/Best-practice-for-W3C-Web-Annotations-generated-based-on-TEI-names-and-dates-module-tags-td4031445.html). One possibility, RDFa, has great appeal due to being an established W3C standard that comes with off-the-shelf tooling (e.g., https://www.w3.org/2012/pyRdfa/ and http://www.sparql.org/sparql.html which can directly run against TEI documents or derived XML formats that maintain [rather than generate] RDFa information). 

In the past, RDFa has been ruled out, partially because of fears it would evolve and this would have a negative impact on the TEI (http://tei-l.970651.n3.nabble.com/TEI-and-RDFa-was-Re-SAWS-and-LOD-was-Re-Cross-references-among-segs-in-TEI-td4025195.html). Since its W3C standardization (2015, https://www.w3.org/TR/rdfa-core/), this risk does no longer exist. 

In 2018, two successful applications of TEI+RDFa in two independent projects have been reported (http://lrec-conf.org/workshops/lrec2018/W23/pdf/10_W23.pdf, http://e-spacio.uned.es/fez/eserv/bibliuned:363-Pruiz3/Ruiz_Fabo_Pablo_DISCO.pdf), thus motivating project-independent specifications, ideally as part of the TEI. I suggest to follow the modeling of https://github.com/postdataproject/disco/#rdfa-attributes.

Note1: This is a follow-up to #311, but a different approach.

[...]

PR #2431, opened 2023-05-08
first draft for a TEI with RDFa exemplar

HelenaSabel:
See discussion in #1860


sabineseifert:
Council discussion at VF2F 16 March 2024: 
also consult with Ontologies SIG

martinascholger:
@HelenaSabel and I have revised and updated the exemplar document [P5/Exemplars/tei_rdfa.odd](https://github.com/TEIC/TEI/pull/2431/files#diff-066b36a9df8479a622d63c42f6d820a87d9c0d3f6d9d34496dd0ba31d16256a2). To facilitate the review process, we have added the schema generated from this exemplar ([P5/Exemplars/tei_rdfa.rng](https://github.com/TEIC/TEI/pull/2431/files#diff-ed62585644dc17c5792a2eca61321bde2fe7a4e2c9ecf301a3fa0ac68629ebfa)) so that it can be easily associated to any other files for testing. We also added a TEI example ([P5/Exemplars/tei_rdfa.tei](https://github.com/TEIC/TEI/pull/2431/files#diff-de45d261be0468b8f04401e9c85def6bcd9f5a9ecadbcc1ada0841c9237c1f66)).

Next steps: 
- [x] add documentation and examples to the newly created odd
- [ ] add odd to the customization page (https://tei-c.org/guidelines/customization/)
- [ ] update make file for generating .rng automatically
- [ ] decide where in the Guidelines the exemplar should be mentioned. 

@chiarcos, could you please have a look at it and make comments?

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+        <moduleRef key="textcrit"/>
+        <moduleRef key="namesdates"/>
+        <moduleRef key="figures"/>
+        <moduleRef key="corpus"/>
+        <moduleRef key="linking"/>
+        <moduleRef key="analysis"/>
+        <moduleRef key="iso-fs"/>
+        <moduleRef key="nets"/>
+        <moduleRef key="certainty"/>        
+        <moduleRef key="tagdocs" except="elementSpec moduleRef schemaSpec" />
```
sydb: My instinct is that this should be
~~~xml
<moduleRef key="tagdocs" include="att code eg egXML gi ident tag val"/>
~~~

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+        -->
+
+        <classSpec type="atts" ident="att.global.analytic.rdfa" mode="add">
+          <desc xml:lang="en" versionDate="2020-07-01">provides additional global attributes for
+            embedding RDF following the RDFa Core 1.1 specification</desc>
+          <attList>
+            <attDef mode="add" ident="about">
+              <gloss xml:lang="en" versionDate="2020-07-01">rdfa-about</gloss>
+              <desc xml:lang="en" versionDate="2020-07-01">a SafeCURIEorCURIEorIRI, used for stating
+                what the data is about (a 'subject' in RDF terminology)</desc>
```
sydb: Should “'subject'” be `<term>subject</term>`?

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+              <datatype>
+                <dataRef key="teidata.enumerated"/>
```
sydb: Huh?
We expect an enumerated list of possible values?

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+          Hook the new class `att.global.analytic.rdfa` into the existing
+          class `att.global.analytic`
+        -->
+        <classSpec type="atts" ident="att.global.analytic" mode="change">
+          <classes mode="change">
+            <memberOf key="att.global.analytic.rdfa" mode="add"/>
+          </classes>
+        </classSpec>
+        
+        <elementSpec ident="citeData" mode="change">
```
hcayless, line 223: Deleting `@property` from `<citeData>` isn't a viable solution going forward. The attribute on citeData is (almost) precisely the same thing, except that it doesn't play a role in defining a proper triple (it specifies how to use the document to construct triples). Do these attributes have to be absolutely global? Can we simply override the definition in this case, as we do in various places with `@type`, for example. I feel like some thinking about the scope of RDFa in TEI might be worthwhile.

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+            </attDef>
+            <attDef mode="add" ident="datatype">
+              <gloss xml:lang="en" versionDate="2020-07-01">rdfa-datatype</gloss>
+              <desc xml:lang="en" versionDate="2020-07-01">a TERMorCURIEorAbsIRI representing a
+                datatype, to express the datatype of a literal</desc>
+              <datatype>
+                <dataRef key="teidata.enumerated"/>
+              </datatype>
+            </attDef>
+            <attDef mode="add" ident="href">
```
hcayless: `@href` and `@src` below are the HTML attributes on `<a>` and ``, for example. We shouldn't add them. The RDFa spec is explaining how these built in HTML attributes interact with RDF—namely they can act in the same way as `@resource`. The corresponding TEI attributes are `@target` and `@url`, though there are many, many more attributes in TEI that could potentially interact with RDFa. See https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-teidata.pointer.html for a rather dizzying list.

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+            <attDef mode="add" ident="rev">
+              <gloss xml:lang="en" versionDate="2020-07-01">rdfa-rev</gloss>
+              <desc xml:lang="en" versionDate="2020-07-01">a white space separated list of
+                TERMorCURIEorAbsIRIs, used for expressing reverse relationships between two
+                resources (also 'predicates')</desc>
+              <datatype maxOccurs="unbounded">
+                <dataRef key="teidata.enumerated"/>
+              </datatype>
+            </attDef>
+            <attDef mode="add" ident="src">
```
hcayless: See comment on `@href` above. We shouldn't add this.

P5/Exemplars/tei_rdfa.odd
```diff
@@ -0,0 +1,227 @@
...
+              <desc xml:lang="en" versionDate="2020-07-01">An attribute used to indicate that the
+                object associated with a rel or property attribute on the same element is to be
+                added to the list for that predicate. The value of this attribute MUST be ignored.
+                Presence of this attribute causes a list to be created if it does not already
+                exist.</desc>
+              <datatype>
+                <dataRef key="teidata.nullOrName"/>
+              </datatype>
+            </attDef>
+            <attDef mode="add" ident="prefix">
```
hcayless, line 123: `@prefix` probably isn't needed in a TEI context since we can use `@xmlns` to declare prefixes. Non-XHTML HTML can't, hence the need for the attribute.

P5/Exemplars/tei_rdfa.tei
```diff
@@ -0,0 +1,63 @@
...
+            <p><!-- supply publication information--> </p>
+         </publicationStmt>
+         <sourceDesc>
+            <p><!-- supply information about the source --></p>
+         </sourceDesc>
+      </fileDesc>
+      <profileDesc>
+         <particDesc>
+            <listPerson>
+               <person xml:id="disco_004e" about="disco:004e" typeof="foaf:Person">
```
hcayless, line 20: I could make an argument that the `@about` attribute is redundant here....maybe something to think about.

P5/Exemplars/tei_rdfa.odd
HelenaSabel: We argue that is a RDFa-native way to declare prefixes that would be interpreted by existing out-of-the-box tools that extract the triples

PR #2435, opened 2023-05-11
Moved calendar to its own class

### LINKED ISSUE #2045: calendar= should not be in att.datable
The `@calendar` attribute is currently a member of att.datable, but should not be.

This is because `@calendar` is, by definition, about “the date represented by the content of this element”. But among the members of att.datable we find several elements whose content is not a date (e.g., `<orgName>` or `<unitDef>`), and **lots** of elements whose content need not include a date (e.g., `<acquisition>` or `<climate>`).

So it is clear to Council that `@calendar` should split out of att.datable. There are some issues still to be resolved, however:
- should `@calendar` be put into a class of its own? (My vote is “yes”)
- if so, what should that class be named? (“att.dated”, maybe?)
- which elements should retain `@calendar` and which should lose it?

### LINKED ISSUE #2432: `<docDate>` should be in att.datable and lose its bespoke `@when`
In reviewing #2045, @ebeshero, @trishaoconnor, and I were surprised by (and thankful for) @martindholmes ' [comment](https://github.com/TEIC/TEI/issues/2045#issuecomment-731271583) that `<docDate>` has its own `@when` and is not part of `att.datable`.  This seems like an error that can be fairly easily resolved.

raffazizzi:
Moved calendar to its own class and deprecated on all att.datable members except `<date>`, `<time>`, `<origDate>`, and `<docDate>`. Fixes #2045. 

I was not able to use `@validUntil` on `@calendar` because it's not to exclude `@calendar` from membership to `att.datable` on the four elements where it needs to remain. So there is a schematron check targeting all the elements it will be removed from.

After the deprecation period, we will need to make `<date>`, `<time>`, `<origDate>`, and `<docDate>` members of the new `att.calendarSystem`. 

If anyone has suggestions on how to do this better, please let me know!

Also made `<docDate>` member of att.datable. Fixes #2432.


P5/Source/Specs/att.calendarSystem.xml
```diff
@@ -0,0 +1,76 @@
...
+      <exemplum versionDate="2008-04-06" xml:lang="fr">
+        <egXML xmlns="http://www.tei-c.org/ns/Examples" source="#fr-ex-Bouiller_Rapport"> L'année
+          1960 fut, en vertu du calendrier grégorien, bissextile ; le 22 juin tomba ainsi le jour de
+          l'été, le <date calendar="#gregorian">22 juin</date>. </egXML>
+      </exemplum>
+      <remarks xml:lang="en" versionDate="2015-05-29">
+        <p>Note that the <att>calendar</att> attribute (unlike <att>datingMethod</att> defined in
+          <ident type="class">att.datable.custom</ident>) defines the calendar system of the date
+          in the original material defined by the parent element, <emph>not</emph> the calendar to
+          which the date is normalized.</p>
```
joeytakeda: Should we update this wording to give a bit clearer guidance as to when to use `@calendar` and when to use `@datingMethod`? Maybe something like:

```
<p>The <att>@calendar</att> attribute defines the calendar system of the date in
 the original material defined by the parent element, and should only be used when the parent
 element contains textual content that represents a date. This differs from
 the <att>datingMethod</att> attribute, which refers to the calendar system
 of the date represented by the attributes defined in
 <ident type="class">att.datable.custom</ident>).</p>
```

P5/Source/Specs/att.datable.xml
```diff
@@ -49,6 +58,20 @@ $Id$
...
+            <sch:let name="me_phrase"
+              value="concat( 'Attribute calendar on ', local-name(.))"/>
+            <sch:assert test="'2024-11-11' cast as xs:date ge current-date()">
+              <sch:value-of select="concat( $me_phrase, ' construct is outdated (as of 2024-11-11); ODD processors may ignore it, and its use is no longer supported' )"/>
```
joeytakeda, line 67: Should we add a link to the GH issue so that users can see the discussion? https://github.com/TEIC/TEI/issues/2045

P5/Source/Specs/att.datable.xml
```diff
@@ -84,7 +107,7 @@ $Id$
     </attDef>
     <attDef ident="period" usage="opt">
       <desc versionDate="2021-04-19" xml:lang="en">supplies pointers to one or more definitions of 
-        named periods of time (typically <gi>category</gi>s or <gi>calendar</gi>s) within which the datable item is understood to have occurred.</desc>
+        named periods of time (typically <gi>category</gi>s, <gi>date</gi>s or <gi>event</gi>s) within which the datable item is understood to have occurred.</desc>
```
joeytakeda, line 110: Depending on preference—comma before "or"? (I can't seem to find any reference to the Oxford Comma in [TCW24](https://teic.github.io/TCW/tcw24) and a brief search of the GL seems to show a near even split)

joeytakeda (commented):
This seems to be the best way to do all of these deprecations, as far as I can tell—just some minor questions/queries about wording, but otherwise, everything looks good to me.

P5/Source/Specs/att.datable.xml
ebeshero: We should prefer the Oxford comma. But I think we can do better than the awkwardly pluralized element names. Maybe this, removing the parentheses: 

```xml
supplies pointers to one or more definitions 
of named periods of time, typically the <gi>category</gi>, 
<gi>date</gi>, or <gi>event</gi> elements 
indicating when the datable item is 
understood to have occurred. 
```

P5/Source/Specs/origDate.xml
```diff
@@ -25,7 +25,8 @@ $Id$
   <desc versionDate="2007-01-21" xml:lang="it">contiene una data espressa in una forma qualsiasi utilizzata per indicare la data di origine di un manoscritto o di una sua parte</desc>
   <classes>
     <memberOf key="att.global"/>
-    <memberOf key="att.datable"/>
+    <memberOf key="att.datable" />
+    <!--<memberOf key="att.calendarSystem"/>--> <!-- Uncomment this after @calendar is deprecated on att.datable -->
```
ebeshero, line 29: I wonder if this is necessary. Can we uncomment it now on dev branch or make the new class available before the deprecation period ends?

P5/Source/Specs/docDate.xml
```diff
@@ -31,54 +31,13 @@ on a title page or in a dateline.</desc>
     <memberOf key="model.titlepagePart"/>
     <memberOf key="model.divWrapper"/>
     <memberOf key="model.pLike.front"/>
+    <memberOf key="att.datable"/>
+    <!--<memberOf key="att.calendarSystem"/>--> <!-- Uncomment  this after @calendar is deprecated on att.datable -->
```
ebeshero, line 35: Can it simply be added to the new class now without issue?

P5/Source/Specs/date.xml
```diff
@@ -25,6 +25,7 @@
     <memberOf key="model.publicationStmtPart.detail"/>
     <memberOf key="att.canonical"/>
     <memberOf key="att.datable"/>
+    <!--<memberOf key="att.calendarSystem"/>--> <!-- Uncomment  this after @calendar is deprecated on att.datable -->
```
ebeshero, line 28: Can the new class simply be made available during the deprecation period?

ebeshero (commented):
I raised some questions and suggested a rewording. I am wondering why we need to comment out the new class membership for now. Can the new class coexist with the old one during the deprecation period?

raffazizzi:
@ebeshero unfortunately having both old and new classes at the same time would define the attribute twice, breaking the build/schema

ebeshero:
Sure, an attribute can't be a member of multiple classes unless they are inherited...
I guess I was wondering if the new version could be on dev branch safely, but probably not given the length of the deprecation period.

P5/Source/Specs/att.datable.xml
sydb: > Should we add a link to the GH issue so that users can see the discussion? #2045

I do not think it is a good idea for the _Guidelines_ to refer to specific issues or discussions except perhaps as a footnote.

P5/Source/Specs/att.datable.xml
sydb: I agree completely the Oxford comma should be used. But the rewording to avoid those icky plural element names alters the semantics. (From “occured within” to “when occurred”.)
How about
~~~xml
supplies pointers to one or more definitions of named periods of time,
within which the datable item is understood to have occurred;
typically each named period is encoded as a <gi>category</gi>, 
<gi>date</gi>, or <gi>event</gi>.
~~~
?

ebeshero: @sydb point taken about changing the semantics, but please, let's not make a complicated run-on with the comma and semicolon punctuation. How about this?

```xml
supplies pointers to one or more definitions 
of named periods of time, typically the <gi>category</gi>, 
<gi>date</gi>, or <gi>event</gi> elements that describe a datable item.
```

ebeshero: ^^^ edited one more time to simplify.

ebeshero:
Really, I wish we did not have to handle this with deprecation, since it's a change involving class memberships and the improvements explain the usage of `@calendar` so much more clearly! Just reviewing the decision for a moment, I suppose we are deprecating an overuse of `@calendar` that has always been problematic but wasn't explained sufficiently. Of course the deprecation is simply to warn people this is going away on some elements.

If we want the improvements to be available as soon as possible, I don't like the persistence of `@calendar` in the old class much at all.

ebeshero:
Not sure if this is viable: What if we revised / updated classes now (no commenting out of what we intend to do), and for the deprecation period, we temporarily supply `@calendar` on the elements where it is to be removed--just without its class and with the deprecation warning notice?

P5/Source/Specs/att.calendarSystem.xml
sydb: a) Yes, @joeytakeda is right, we should improve the wording.
b) I think his suggestion is better, but could still be improved.

The following is still a bit off, I think, but is perhaps closer.

~~~xml
   Note that the <att>calendar</att> attribute declares the calendar
   system used to interpret the textual content of an element, content
   presumably transcribed from an original source. It does
   <emph>not</emph> modify the interpretation of the normalization
   attributes provided by <ident type="class">att.datable.w3c</ident>,
   <ident type="class">att.datable.iso</ident>, or <ident
   type="class">att.datable.custom</ident>. Attributes from those first two
   classes are always interpreted as Gregorian or proleptic Gregorian
   dates, as per the respective standards on which they are based. The
   calender system used to interpret the last (<ident
   type="class">att.datable.custom</ident>) may be specified with
   <att>datingMethod</att>.
~~~

P5/Source/Specs/att.datable.xml
sydb: OK.

P5/Source/Specs/att.datable.xml
```diff
@@ -49,6 +58,20 @@ $Id$
...
...r] | tei:district[@calendar] | tei:editor[@calendar] | tei:education[@calendar] | tei:event[@calendar] | tei:faith[@calendar] | tei:floruit[@calendar] | tei:funder[@calendar] | tei:gender[@calendar] | tei:geogFeat[@calendar] | tei:geogName[@calendar] | tei:idno[@calendar] | tei:langKnowledge[@calendar] | tei:langKnown[@calendar] | tei:licence[@calendar] | tei:location[@calendar] | tei:meeting[@calendar] | tei:name[@calendar] | tei:nationality[@calendar] | tei:objectName[@calendar] | tei:occupation[@calendar] | tei:offset[@calendar] | tei:orgName[@calendar] | tei:origPlace[@calendar] | tei:origin[@calendar] | tei:persName[@calendar] | tei:persPronouns[@calendar] | tei:placeName[@calendar] | tei:population[@calendar] | tei:precision[@calendar] | tei:principal[@calendar] | tei:provenance[@calendar] | tei:region[@calendar] | tei:relation[@calendar] | tei:residence[@calendar] | tei:resp[@calendar] | tei:seal[@calendar] | tei:settlement[@calendar] | tei:sex[@calendar] | tei:socecStatus[@calendar] | tei:sponsor[@calendar] | tei:stamp[@calendar] | tei:state[@calendar] | tei:terrain[@calendar] | tei:title[@calendar] | tei:trait[@calendar] | tei:unitDecl[@calendar] | tei:unitDef[@calendar]">
```
sydb: This would be a lot easier to read as
~~~xml
<sch:rule context="tei:*[@calendar][ not( self::tei:date | self::tei:time | self::tei:origDate ) ]">
~~~

sydb (changes requested):
Overall looks good to me. I think the remarks of new att.calednarSystem and the description of `@period` need improvement before release, but otherwise good to go.

Issue #2285, opened 2022-04-22
more on `<altIdent>`


sydb:
### introduction

If you thought #2049 was more than enough fixes for [`<altIdent>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-altIdent.html), think again. Now that we are on the road to improving its content and where it can go in a TEI document, it is time to start thinking through its semantics and required constraints more carefully. (This, BTW, comes up acutely in the context of the [atop](https://github.com/teic/atop) project.)

After #2049 has been completed, `<altIdent>` will only be allowed as a child of `<attDef>`, `<classSpec>`, `<constraintSpec>`, `<dataSpec>`, `<elementSpec>`, `<macroSpec>`, or `<valItem>`. In all cases it may appear 0 or more times, and may or may not have an `@xml:lang`.

We (or at least, I) do not know _why_ `<altIdent>` is repeatable. As far as I can tell, there is **no** occurrence of a 2nd `<altIdent>` child of _any_ element anywhere in the _Guidelines_ (including examples) or in the Exemplars/ directory. (I searched for `//*[ tei:altIdent[2] | teix:altIdent[2] ]` in all files within the P5/ hierarchy with file extensions ".xml", ".odd", or ".tei".)

So it seems to me we need a discussion about what a 2nd (or 3rd or 4th) `<altIdent>` might mean. Heck, I don’t even know what the _1st_ one means for anything other than an `<elementSpec>`, an `<attDef>`, or a `<valItem>`!

### singleton

In those three constructs (`<elementSpec>`, `<attDef>`, `<valItem>`) it means “this is the name or value of this thing that you (the processor) should put in the schema; you (the human) should, as usual, use the `@ident` of the parent element refer to it from within the ODD(-chain), but do not expect it in the output schema(s).” I have tested, and at least the “use this name in the schema” part works for all 3 of them.

But what does it mean for `<dataSpec>`? The name of a datatype never makes it into the RELAX NG schema, so the fact that you can call it something different in the ODD file is irrelevant, isn’t it? I have not tested, but seems to me the same is true for `<constraintSpec>`.

[...]

PR #2440, opened 2023-06-07
Restrict where `<altIdent>` can go in tei_customization and p5odds

sydb:
In response to Council’s decision of 2023-05-09 re: #2285, restrict where `<altIdent>` is allowed to only `<elementSpec>`, `<attDef>`, and `<valItem>` in p5odds and tei_customization:

**p5odds**
Just added a `<constraintSpec>`

**tei_customization**
Moved `<altIdent>` out of model.identSynonyms and then add it back into content models of `<attDef>` and `<valItem>`. While I was there also:

- Removed the `<constraintSpec>` that limited `<altIdent>` to an NCName (TEI P5 now does this)
- Updated the ODD file itself to be based on 4.6.0
- Increased the version number to 0.9.0b — I am guessing we forgot to increase it to 1.0.0 on actual release. Sigh.

I have run `make test` in the P5 directory, and all tests passed. What I have not done is actually test that p5odds and tei_customization do the right thing.


HelenaSabel:
Hello @sydb!

The schema `tei_odds.rng` (generated with `make exemplars`) is not firing errors in a test file that contains `<altIdent>` as a child of `<attDef>` and as a child of `<constraintSpec>`.

In the case of `tei_customization.rng`, I tested a few ODDs and I always get the error `conflicting ID-types for attribute "defaultPhase" of element "schema" from namespace "http://purl.oclc.org/dsdl/schematron"` but this also happens with the version in `dev`.

sydb:
Heh-heh. The test for p5odds is not in the RELAX NG, it is in the Schematron. However, I just tested and found a bug: I used `../tei:attDef` where I should have said `parent::tei:attDef`. I plan to check in a fix in a few minutes.

In the meantime, the error you are getting from tei_customization is the result of having DTD compatibility mode turned on. If you are using `jing` on the commandline add a `-i` switch; if you are using oXygen _uncheck_ the “Check ID/IDREF” box in the Options > Preferences > XML > XML Parser > RELAX NG pane.
See https://journals.openedition.org/jtei/2573#annexe.

sydb:
Fixed in [251161c](https://github.com/TEIC/TEI/pull/2440/commits/251161ca67aa877c89f656f5813da9f361b519eb).

HelenaSabel (approved):

PR #2441, opened 2023-06-10
remove `<remarks>` from "deprecation-should-be-explained"

### LINKED ISSUE #2433: Cannot set validUntil on remark
Setting `@validUntil` on `<remark>` causes the error:
```
assertion failed: A deprecated construct should include, whenever            possible, an explanation, but this remarks does not have a child            <desc type ="deprecationInfo" >
```

... but `<remark>` cannot have a child `<desc>`. This should probably be a warning, or it should be removed from elements that cannot contain `<desc>`.

sydb:
Added `<remarks>` to the list of elements for which the "deprecation-should-be-explained" constraint should **not** be fired.
Fixes #2433, I hope.


joeytakeda (approved):
Looks good to me! (in fact, I'm mildly surprised that the schematron worked without the `tei:` prefix to begin with)

martindholmes:
Looks good to me.

martinascholger (approved):

Issue #2173, opened 2021-08-20
Add warning when using <constraintSpec> inside <classSpec> 


HelenaSabel:
As a temporary solution for [Stylesheets ticket #519](https://github.com/TEIC/Stylesheets/issues/519), a warning should be fired whenever a `<constraintSpec>` is used inside `<classSpec>` without an explicit `@context`.

sydb:
Actually, since `<constraintSpec>` is permitted in _lots_ of (mostly ridiculous) places,[1] methinks the warning should be fired whenever a `<constraintSpec>` is used without an explicit `@context` anywhere an intelligent context is not generated by the Stylesheets.

So the first step in resolving this ticket is generating the list of places the Stylesheets _do_ provide an intelligent context. Schematron is processed in at least 2 different ways, I submit that if the context is generate in either method, we should not issue a warning.

[...]

PR #2442, opened 2023-06-17
First crack at addressing #2173 in a serious way.

sydb:
Holy smoke. This has been a _lot_ of work.

Branch name is “sydb_2173”. It builds on my local system (in a docker container), and passes the tests established on [the ticket](https://github.com/TEIC/TEI/issues/2173). Only possible problem I have noticed is the "unique_xmlIDs" constraint of tei_customization is not contextualizable. That is on purpose, of course. The current Stylesheets generate `"@xml:id"` as the value of `@context`, thus firing on _all_ occurrences of `@xml:id` (whether the element it appears on got it via att.global or not), which is the desired effect. If necessary we can just add an `<sch:rule context="@xml:id">` wrapper. (It would be added in P5/Utilities/TEI-to-tei_customization.xslt, around line 1465.)

Things I had to do to make this (putting the constraint demonstrated in the ODD attached to the ticket into constraintSpec.xml) work follow, not in any particular order.

1. Added more informative messages to P5/Makefile and P5/antbuilder.xml
1. In the 2nd `@test` in the new "context_for_constraint" constraint (the major complex one — “I am inside a `<specGrp>`”, “and there are one or more references to one or more of my `<scpecGrp>` ancestors” …) I added a clause to exclude cases of `<elementSpec>` inside `<specGrp>`. (There are 47 such cases in the _Guideliens_.)
1. I added a predicate to the `@context` of the `<sch:rule>` so it is not fired if the `<constraintSpec>` is a descendant of an `<egXML>`. (This is only necessary because of how the build process does testing, I think. If you look at p5examples.isosch.xsl (very carefully) you will find that the `tei:` prefix is bound to the Examples namespace.) 
1. Within an XPath comment in that same `@test` I changed “don't” to “don’t” (i.e., U+0027 → U+2019), because the U+0027 caused Utilities/iso_dsdl_include.xsl to mis-count the parentheses, and thus broke the build.
1. I tweaked some wording in TD. (I did not fix the incorrect prose about `<sch:ns>`.)
1. I replaced expected-results/ValidatorLog.xml. Note also that in doing so I added the new comments generated by Utilities/validator.xsl and deleted a whole bunch of old warnings that, as far as I know, are long since past their usefulness.


sydb:
This PR will probably never be needed, as we expect Council to simply require `@context` in all cases of `<sch:assert>` or `<sch:report>` within a `<constraint>`.
But some of the work here may be useful. So we should kill this PR when the above has been done.

P5/Source/Specs/constraintSpec.xml
```diff
@@ -83,6 +83,148 @@ $Id$
...
+                                         or &lt;metaMark>, it is probably appropriate to
+                                         check. :)
+                              )
+                            )
+                          )">
+        this constraintSpec <sch:value-of select="
+        if (@ident)
+        then concat('(&quot;', @ident,'&quot;)')
+        else ''"/> has no @context, but is in a specGrp that has at
+        least 1 reference to it that is not contextulizable.</sch:report>
```
raffazizzi, line 224: typo: contextulizable (missing "a") and I would spell out 1 to "one"

sabineseifert:
Full Council discussion at VF2F 16 March 2024:
- PR can be closed “won’t fix” because [#2513](https://github.com/TEIC/TEI/pull/2513) is a superset of this 
- branch should not be deleted yet, as there are some other changes (particularly to Makefile and buid.xml) that we may want to keep

martindholmes:
@sydb The Council decision in 2024 says this PR should be closed. Do you agree?

martindholmes (commented):
Judging by the discussion, this PR should just be closed since it's no longer needed.

Issue #2426, opened 2023-05-03
Usage of `<dataRef>`+`<dataSpec>` and `<macroRef>`+`<macroSpec>`


martindholmes:
The ATOP team notes that:

- A `<dataSpec>` can be pointed to using a `<dataRef>` in the context of an attribute definition; but
- `<dataSpec>` allows the use of `<elementRef>` or `<anyElement>`, which would result in an element being included in the content model of an attribute, which is of course invalid.

We suggest that it would be useful to:

- Clarify in the prose that `<dataRef>` and `<dataSpec>` should be used for the definition of attributes, while `<macroRef>` and `<macroSpec>` should be used for the definition of content models for elements.
- Schematron should be added to prevent the use of `<anyElement>` or `<elementRef>` as a descendant of `<content>` in the context of a `<dataSpec>`.

This would not catch all potential errors, of course, because ODD-chaining could still result in elements ending up in the content model of attributes, but the ODD processor can catch these. But the most straightforward cases would be caught by Schematron.

ebeshero:
From Council discussion: Consider whether we should prevent macroRef from being used in `<attDef>`? That would prevent the other way of putting an element inside an attribute.

joeytakeda:
Thinking about this the other way: what _should_ a `dataSpec` allow as a child of `content`? Right now, it allows:

* alternate 
* anyElement 
* classRef 
* dataRef 
* elementRef 
* empty 
* macroRef 
* sequence 
* textNode

I would suggest that nothing other than `<dataRef>` makes sense; the only reason I can see a `dataRef` being used in an element would be to specify a `<valList>` for an element's content (which is handled via `dataSpec/valList` anyway).

sydb:
@ebeshero — At first blush keeping `<macroRef>` out of `<attDef>` seems like a good idea. But I do not think it _is_ allowed, anyway.

@joeytakeda — Good thought experiment. But I claim `<alternate>`, `<sequence>`, `<textNode>`, and even `<empty>` are useful. Imagine, e.g., that I want a datatype that gives me latitude & longitude with leading zeroes in degrees to 2 decimal places and elevation in meters:
~~~xml
        <dataSpec ident="lle">
          <content>
            <sequence>
              <dataRef name="float" restriction="-?[0-9][0-9][0-9]\.[0-9][0-9]"/>
              <dataRef name="float" restriction="-?[0-9][0-9][0-9]\.[0-9][0-9]"/>
[...]

PR #2443, opened 2023-06-18
Address #2426 — do not allow elements or classes inside a datatype

sydb:
What I did:
 * Added para to "TD-datatypes" (22.5.3.1) asserting that, while they may be used in a content model, because datatypes are for attr vals, cannot have element or attr.
 * Changed example that shows defining “more than one” occurrence of a datatype from `@target` of att.pointing to `@columns` of `<layout>`. I did this because I think the URI example is particularly problematic, as someone who does not carefully read a modern version of teidata.pointer but knows how xs:anyURI is defined could be quite confused.
 * Made clear in list of ways a `<dataRef>` refers to a dataype that TEI datatypes are defined by the `<dataSpec>` element.
 * Added a `role="warn"` constraint to `<dataSpec>`.
 * Corrected half a dozen references to XML Schema Part 2: Datatypes Second Edition in dataRef and dataFacet tagdocs.

Note that I used an `@role` of "warn" on the constraint itself, on the theory that we should provide a sort of deprecation period. Feel free to disagree. (Either way — no deprecation needed, remove "warn", or re-word warning to make it a real deprecation.)

### DIFF
--- a/P5/Source/Guidelines/en/TD-DocumentationElements.xml
+++ b/P5/Source/Guidelines/en/TD-DocumentationElements.xml
@@ -995,17 +995,19 @@ to mark any technical term, thus:
         <p>The <soCalled>datatype</soCalled> (i.e. the kind of value) for an attribute may be specified using
           the elements <gi>datatype</gi> and <gi>dataRef</gi>. A datatype may be defined in any of the
           following three ways: <list>
-            <item>by reference to an existing TEI datatype definition; </item>
-            <item>by use of its name in the widely used schema datatype library maintained by the W3C as part
-              of the definition of its schema language; </item>
+            <item>by reference to an existing TEI datatype definition, itself defined by a <gi>dataSpec</gi>;</item>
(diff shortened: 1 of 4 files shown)


sydb:
The ticket suggests we add prose that says, essentially “Macros are for use in content models, not attr values”. I did not do so, because `<macroRef>` cannot occur inside `<attDef>`, anyway.

raffazizzi (approved):

raffazizzi:
Reviewed files and tested schematron constraint. It all looks good to me. Build checks should be fixed before merging.

sydb:
The build check failures are “rnv: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by rnv)”, so I kinda doubt it has anything to do with this PR. How do we ask Mr. Jenkins to re-run the checks? (Do I just go in and change some whitespace somewhere? :-)

HelenaSabel:
Everything looks great. 
Build failed because of #2455. Tests pass in my local machine so I think it’s safe to merge.

PR #2446, opened 2023-07-10
Fix spacing

rettinghaus:
This PR just cleans up some unnecessary/unwanted doubled spaces in Specs and Guidelines.

### DIFF
--- a/P5/Source/Guidelines/en/CC-LanguageCorpora.xml
+++ b/P5/Source/Guidelines/en/CC-LanguageCorpora.xml
@@ -232,7 +232,7 @@ text--><!-- JC: 2018-07-20 ... uh no it can't! -->.  For further discussion, see
  </p>
 <p>All composite texts share the characteristic that their different
 component texts may be of structurally similar or dissimilar types.  If
-all  component texts may all be encoded using the same  module,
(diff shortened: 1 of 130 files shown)


P5/Source/Guidelines/en/CC-LanguageCorpora.xml
```diff
@@ -733,7 +733,7 @@ way:
       <activity>radio performance</activity>
    </setting>
 </settingDesc></egXML></p>
-<p>Again, a  more detailed encoding for  places is feasible if the
+<p>Again, a mre detailed encoding forr pces is feasible if the
```
peterstadler, line 736: This PR seems to introduce various typos. Probably your regex messed something up?

peterstadler (changes requested):
In general, I don't believe we need to clean up the double whitespaces since this is not significant whitespace in XML.

Second, the current PR introduces various typos that must not be merged!

P5/Source/Guidelines/en/CC-LanguageCorpora.xml
rettinghaus: Weird.

PR #2450, opened 2023-07-29
Update unit.xml

knagasaki:
Add a Japanese translation


martindholmes (approved):
Looks OK. It specifies only "units" rather than "units of measurement", but I would guess they couldn't really be units of anything else!

PR #2452, opened 2023-08-02
Update links to https

rettinghaus:
This small PR updates links to the IETF, IANA, and Unicode websites to be secure.


sydb (approved):
I tested one URL from for each host; the new ones worked, so I think this is ready to merge.
Thank you @rettinghaus !!

sydb:
Note that the check that fails is dying on a C library — has nothing to do with the changes @rettinghaus made.

PR #2453, opened 2023-08-06
remove `@scheme` requirement when `<constraintSpec>` is mode=change or delete

sydb:
* Made `@scheme` of `<constraintSpec>` optional (per closed schema)
 * Made `@scheme` of `<constraintSpec>` required iff mode=add (default) or mode=replace (per open schema)
 * Moved declarations of Schematron namespaces from constraintSpec to TEI in order to make above work
 * Added a case in detest.odd to test new constraints

Note for reviewing: I did this work on a machine that does not have spell-checking installed for Emacs (yet). So probably worth reading the prose changes carefully.


sydb:
Note that the tests here are still failing on “/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found”, so probably nothing to do with the changes.

This branch passed all the tests that my Mac laptop runs, but I do not have a docker container on this machine.

P5/Test/expected-results/detest_odd_schematron.log
```diff
@@ -1,11 +1,11 @@
-Buildfile: /TEI/P5/Test/antruntest.xml
+Buildfile: /Users/syd/Documents/TEI_gitHub/P5/Test/antruntest.xml
```
joeytakeda: I presume that the expected results here should not be user specific -- should these all be `/TEI/` ?

PR #2473, opened 2023-09-12
Adjustments for deprecation of @calendar

raffazizzi:
To be discussed at the next council meeting because we need to decide whether to do Step 0 below which would cause a delay to the deprecation period.

Optional Step 0: 

- announce that there will be an upcoming change for ODD writers who want to include `@calendar` via `att.datable`, because:

Step 1:

- make `date`, `docDate`, `origDate`, and `time` members of `att.calendarSystem` 
- remove `@calendar` from `att.datable`
- define `@calendar` on the 40+ elements that it needs to be removed from and provide `@validUntil` on those attributes. This will make sure that the deprecation notice appears on the guidelines as well as in the generated schema.

Step 2:

- complete process by removing `@calendar` from the elements.


raffazizzi:
During the 2023-10/13 council call we decided to:
* proceed with Step 1 on this branch to be merged into `dev` when ready;
* warn ODD users on TEI-L about the forthcoming change (likely coming in the November 23 release)
* open new ticket to discuss deprecation of changes that affect ODD writing even when they don't directly affect the generated TEI schema. Another example can be found in [the previous release (4.6.0)](https://tei-c.org/release/doc/tei-p5-doc/readme-4.6.0.html): "The model.glossLike class (which contained altIdent, equiv, and gloss) has been renamed to model.identSynonyms."

P5/Exemplars/tei_jtei.odd
```diff
@@ -1993,9 +1993,9 @@
               <attDef ident="group" mode="delete"/>
             </attList>
           </classSpec>
+	  <!-- We should be removing @calendar here!! -->
```
raffazizzi: @sydb, this comment can go now, right?

P5/Exemplars/tei_lite.odd
```diff
@@ -2950,11 +2950,7 @@ $Id$
               <attDef ident="to" mode="delete"/>
             </attList>
           </classSpec>
-          <classSpec type="atts" ident="att.datable" module="tei" mode="change">
-            <attList>
-              <attDef ident="calendar" mode="delete"/>
-            </attList>
-          </classSpec>
+	  <!-- We should be removing @calendar here!! -->
```
raffazizzi: Ditto, this can go, @sydb ?

P5/Exemplars/tei_jtei.odd
sydb: Uhhh … do we not need to do what it says, first? I mean these exemplars are not supposed to have a `@calendar` attribute.

P5/Exemplars/tei_lite.odd
sydb: ditto.

P5/Exemplars/tei_jtei.odd
raffazizzi: But it's done AFAIK! The attribute is no longer in `att.datable` in this PR's code.

sydb (approved):
Tests pass (including tei_jtei and tei_lite), and otherwise seems to look OK.

Issue #2407, opened 2023-03-22
prose example mis-match for `@who` (of `<sp>`)


sydb:
In att.ascribed.xml the paragraph of description in the 1st `<exemplum>` says
~~~xml
        <p>In the following example from Hamlet, speeches (<gi>sp</gi>) in the body of the play 
        are linked to <gi>castItem</gi> elements in the <gi>castList</gi> using the <att>who</att>
        attribute.</p>
~~~
But if you look at the example, the `@who` attributes actually point to the `<role>` elements, not the `<castItem>`s.

JanelleJenstad:
Easy fix (I think). I will revise the example and chapter text after Council F2F.

sydb:
Note — PR with a proposed fix is #2477.

ebeshero:
Merged and closed. Small editorial changes like this can be pushed directly to dev.

PR #2477, opened 2023-09-29
Replaced castItem with role in the exemplum for `#2407`

trishaoconnor:
Updated the exemplum in att.ascribed.xml with @sydb to replace `<castItem>` with `<role>` because the `@who` attributes point to the `<role>` elements and not the `<castItem>` elements.


sydb:
Note — this is a fix for #2407.

ebeshero:
This repair looks perfectly fine to me. I think it should be merged to dev immediately.

Issue #2480, opened 2023-10-02
Build is broken because of inclusion of att.calendarSystem


martindholmes:
It looks like att.calendarSystem.xml was inadvertently included in ND when it's not supposed to be, which has broken the documentation build.

martindholmes:
I've created PR #2481 just to get the build working again.

ebeshero:
I think we've fully resolved this ticket and the larger att.calendarSystem situation with these two merged PRs, so I'm closing this ticket. Please re-open if we've missed anything!

https://github.com/TEIC/TEI/pull/2481 
https://github.com/TEIC/TEI/pull/2473

PR #2481, opened 2023-10-02
Comment out att.calendarSystem.xml inclusion.

martindholmes:
This is a quick fix for issue #2480 to get the build working again.


sydb:
OK, so we merge this now, @ebeshero will put a note in agenda for upcoming meeting to remember to revert this if & when att.calendarSystem is fixed. (The reminder may not be necessary, as we cannot test it _without_ uncommenting this, but …)

Issue #2469, opened 2023-09-06
samplingDecl gloss seems too partial to corpora


raffazizzi:
The current (v 4.6.0) gloss of `<samplingDecl>` says (emphasis mine): 
> contains a prose description of the rationale and methods used in sampling texts in the **creation of a corpus or collection**.

[A paper](https://teimec2023.uni-paderborn.de/contributions/137.html) at TEIMEC23 suggested using `<samplingDecl>` to describe the completeness of the encoding of _one_ document, the sampling being justified by analytical reasons: they were interested in only encoding parts of a document, while being explicit about what they were leaving out.

The gloss for `<samplingDecl>` doesn't seem to support this use of the element, but [the guidelines](https://tei-c.org/release/doc/tei-p5-doc/en/html/HD.html#HD52) are more permissive (emphasis mine):

> The [samplingDecl](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-samplingDecl.html) element may be used to describe, in prose, the rationale and methods used in selecting texts, **or parts of text**, for inclusion in the resource.

I think the gloss should be updated accordingly. For example:

"contains a prose description of the rationale and methods used in sampling parts of a text or in sampling texts in the creation of a corpus or collection."

lb42:
The description in the prose is much better also because it avoids the repetition of the word 'sampling'

joeytakeda:
This was fixed in #2485

PR #2485, opened 2023-10-10
Updated the description for samplingDecl element according to #2469

trishaoconnor:
Updated the description for samplingDecl element with @sydb according to #2469:

"contains a prose description of the rationale and methods used in selecting texts, or parts of a text, for inclusion in the resource."


lb42 (commented):
How about  "...used in selecting material for inclusion in the tei resource." ?  Plenty of corpora include  non textual stuff like pictures  (or dont)

sydb:
For the record, I like @lb42’s suggestion:

> contains a prose description of the rationale and methods used in selecting material for inclusion in the TEI resource.

But more imporantly, the `@versionDate` needs to be updated.

raffazizzi (approved):

Issue #2070, opened 2020-12-04
Make file points at non-existent files


martindholmes:
With @sydb: The P5 build tries to retrieve the VERSION files from all previous versions of TEI (in the dist-doc.stamp target, which ends up invoking a Stylesheets file Stylesheets/profiles/readme/html/to.xsl), pointing to paths such as:

https://www.tei-c.org/Vault/P5/4.1.0/VERSION

for every single version of P5. However, those files are now in a different location:

https://www.tei-c.org/Vault/P5/4.1.0/doc/tei-p5-doc/VERSION

So the fix needs to be made in the Stylesheets, but this affects only the P5 build (as far as we know).

sydb:
Note to @npcole or anyone else trying to tackle this: seems to me the problematic part of file `Stylesheets/profiles/readme/html/to.xsl` is roughly line 66.

sydb:
@martindholmes (and @npcole) —
Once we fix this (which I believe I am on the verge of doing), how do we test it?

martindholmes:
@sydb I think the only test would be to make a local build of P5 and see if the readmes are retrieved or not.

ebeshero:
@martindholmes @sydb Just did a local build to see if I could test this...and I'm not sure which file to check for the URLs?

martindholmes:
@ebeshero I'm not sure either now. @martinascholger how do you produce the readmes for each of the releases? The ones linked here: 

https://tei-c.org/guidelines/p5/

peterstadler:
@ebeshero , @martindholmes the issue is just about the warning
```
BUILD: Make web pages for release notes
for i in ReleaseNotes/readme*xml; do \
ANT_OPTS="-Xss2m -Xmx752m -Djava.awt.headless=true" ant -q -f /var/jenkins_home/jobs/Stylesheets-dev/lastSuccessful/archive/dist/xml/tei/stylesheet/html/build-to.xml -lib Utilities/lib/saxon10he.jar -Dprofiledir=/var/jenkins_home/jobs/Stylesheets-dev/lastSuccessful/archive/dist/xml/tei/stylesheet/profiles -Dprofile=readme \
-DinputFile=`pwd`/$i -DoutputFile=`pwd`/release/tei-p5-doc/share/doc/tei-p5-doc/`basename $i .xml`.html; \
done
     [echo] XSLT generate HTML files (language ${lang})
Warning: XML resolver not found; external catalogs will be ignored
[...]

PR #2494, opened 2023-10-23
README encoding improvements ala #2070

sydb:
Per #2070 step 2, change encoding of class names from `<ident>` (with or without `type="class"`) to `<name type="class">`; and, in the case of 2.2.0 element names and class names from plain text to `<gi>` and `<name type="class">`, respectively. I did _not_ fix any of the other individual problems noted in #2070 note [2], yet, but might get to it before this is merged. (None of those changes _have_ to be done in conjunction with the changes here, but we should do them now lest we forget.)

**WARNING**
This PR _cannot_ be considered in isolation, it must be combined with the [corresponding PR in the Stylesheets repo](https://github.com/TEIC/Stylesheets/pull/638).


sydb:
In 29935ed1e I fixed most of the little things listed in note [2] on the ticket, #2070. I do not claim all errors are fixed, but I do claim that this branch is now significantly better than dev.

PR #2495, opened 2023-10-23
relax restriction on order of top level elements within msDesc (and friends)

sydb:
See #2214.


sydb:
Pinging @schassan, @holfordm, MS-SIGgers, @trishaoconnor, and @jamescummings to review this PR ASAP so we can hopefully get it into the _Guidelines_ next week.

jamescummings (approved):
The pull request looks good to me (I read through the changes -- though I only skimmed over the testing bit -- and have not generated a schema to test it). 

My one question is whether the Schematron constraint also fires when these elements are used inside msPart and msFrag? (I think it should, right? The reason for the constraint is the same.)

sydb:
Yes, @jamescummings, the Schematron fires on `<msContents>`, `<physDesc>`, `<history>`, and `<additional>` wherever they occur. Luckily those elements should (currently) be constrained to 0 or 1 occurrence wherever they occur.

Issue #2132, opened 2021-04-12
Additional version/date attributes for gaiji description


747:
Due to the gradual and time-consuming procedure of Han character standardization into Unicode, an unencoded Han gaiji will likely have multiple identities as well as go through property changes during and after the standardization process. In order to maximize the stability of text body using the gaiji module, the character/glyph description needs the capacity to record the update history for traceable collation.

Thus we suggest:

- Extended versioning: the existing `@version` is limited to the Unicode Standard version and insufficient to support modifiable properties. We will need attributes to delimit the start and end points in Unicode version (perhaps as `@verFrom` and `@verTo`). Versioning systems outside Unicode should better be supported as well for those regional or specialized character sets that are (still) widely used.

- Datable elements: to further support non-version-based change items, we should allow `att.datable` attributes to subelements of `<char>`/`<glyph>`. Whether the version-based and date-based attributes can coexist is subject to discussion.

An example is like (`<mapping>`s contain numeral notations instead of real code points for visibility):

``` xml
<char xml:id="myChar">
  <localProp name="Name" value="A LOCAL GAIJI" />
  <unihanProp name="kIRG_USource" value="U-012345" verFrom="1X.0" verTo="1Y.0" />
  <unihanProp name="kIRG_SSource" value="S-567890" verFrom="1Z.0" />
  <mapping type="internal">0xABCD</mapping>
  <mapping type="PUA" from="2012-01-01" to="2018-03-31">U+FXXX</mapping>
  <mapping type="standard" from="2018-04-01" to="2019-10-15">U+YYYYY</mapping>
  <mapping type="standard" from="2019-10-16">U+YYYYY U+E0100</mapping>
</char>
```

A hypothetical sample for the local version attribute:

``` xml
  <localProp name="reading" value="MEOW" localVer="1.1.0" />
  <localProp name="reading" value="OINK" localVer="1.2.0" />
```

It could be also dated with start and end e.g. using `@localVerFrom` and `@localVerTo`.

**By-Question**: How do we handle [IVD (UTS #37)](http://www.unicode.org/reports/tr37/) versions and properties, which are not linked to the Unicode proper's in any ways?

747:
From: https://github.com/TEIC/TEI/issues/1805#issuecomment-783100905

sydb:
Makes perfect sense to me that multiple version numbers would need to be recorded. But will it ever be necessary to specify anything other than
1. No version # at all (not very interesting)
2. A single version number (already supported)
[...]

PR #2511, opened 2023-11-29
2 of 3 requested improvements to gaiji descriptions

sydb:
Per #2132 this PR does two things:

- Adds `@scheme` to att.gaijiProp with a default value of "Unicode".
- Adds `<mapping>` , `<localProp>`, `<unicodeProp>`, and `<unihanProp>` to att.datable.

It does _not_ do the main item requested on that ticket (provide a mechanism for providing a range of versions to att.gaijiProp) because we have not actually figured exactly how to do that yet.


martinascholger (approved):
Looks good

Issue #2510, opened 2023-11-29
`@context` should be required for Schematron rules in ODD


martindholmes:
Following discussion by the ATOP group and by a larger group of attendees at the TEI Members Meeting in 2023, there is a general consensus that the current practice of allowing Schematron rules to be written without an explicit `@context` and then having an ODD processor attempt to infer a suitable context for the rule is unworkable. Therefore `<sch:rule>` should be required when there is `<sch:assert>` or `<sch:report>`, and `@context` should be required on it (as it is by the Schematron schema itself).

sydb:
@martindholmes and I discussed the caveat that only `<constraintSpec scheme="schematron">`s that contain an assertion (via either  `<sch:assert>` or `<sch:report>`) are required to have an explicit context (via `<sch:rule>` with `@context`). At first blush, it might seem unnecessary. When would there be a `<constraintSpec>` that does _not_ contain an assertion, other than one that contains only namespace declarations (which soon enough will be in a separate element altogether — see #2330)? Turns out there are at least two answers to that question:

- Abstract patterns do not have assertions in them. (And soon the _TEI Guidelines_ themselves will be using abstract patterns.)
- On occasion folks use `<constraintSpec>` to contain only some combination of `<sch:let>` and `<sqf:fix>` (which itself contains XSLT). There are 2 cases in the set of ODDs people contributed to ATOP for testing.

sydb:
Should have been closed when #2513 was merged, no?

PR #2513, opened 2023-12-15
issue deprecation warning when assertion is made without context

sydb:
Work on #2510.

Since this does not use the `@validUntil` deprecation mechanism, does it need to be added to Appendix G by hand?

I chose 2024-06-30 as end-of-deprecation date on my own accord. But since there is basically no difference between deprecation and error, here (although I suppose we could add role=warning), should we just go straight to error?

Anyway, it passes all tests on my local system both in & out of docker.

Reviewers: You will definitely want to ignore whitespace if you try comparing code. (Other than the code of `<constraintSpec>`, it is mostly a kinda pointless activity, IMHO.)


sydb:
P.S. Since after this almost all Schematron was expressed with a namespace prefix (e.g., `<sch:assert>`) rather than an explicit namespace binding (e.g. `<assert xmlns="http://purl.oclc.org/dsdl/schematron">`), I changed the last few, as well. So now **all** Schematron in the Source/ directory uses namespace prefixes.

martinascholger:
@martindholmes the tests are passing now, could you please have a look at it?

martindholmes (approved):
This looks good to me. But I really wish we could keep PRs to a single set of changes; this seems to be changing namespace mechanisms and whitespace as well as handling the `@context` problem. When there's so much change in one batch, it's really hard to review the PR. Could that be a recommendation somewhere?

martinascholger:
Thanks @martindholmes for the quick review. You are right, different changes should not be tackled all at once.  We are trying to stick to our policy better.

Issue #2487, opened 2023-10-13
Deprecate and then remove `<re>` for `#2456`


trishaoconnor:
While working with @sydb on #2456, the discussion with @laurentromary highlighted that `<re>` is no longer used in the context of [TEI Lex 0](https://dariah-eric.github.io/lexicalresources/pages/TEILex0/TEILex0.html) and may no longer be needed at all anymore. 

Given that `<entry>` is recursive, the proposal is to implement a long deprecation period for `<re>` (to facilitate legacy usage of `<re>`) with the suggestion that the value `"relatedEntry"` be used on `<entry>`.

ebeshero:
Appears related to #2488, so I'm adding the same ticket assignees as we sprint to the 4.7 release...

PR #2521, opened 2024-01-10
Deprecated the `<re>` element

trishaoconnor:
Deprecated the `<re>` element for #2487 and in examples replaced `<re>` with `<entry>` element with the `@type` set to "relatedEntry".

### DIFF
--- a/P5/Source/Specs/re.xml
+++ b/P5/Source/Specs/re.xml
@@ -72,7 +72,7 @@ $Id$
           <def>of, relating to, or affecting a nerve or the nervous system</def>
         </sense>
         <sense n="2"> ... </sense>
-        <re>
+        <entry type="relatedEntry">
           <form>
             <orth>neurally</orth>
             <pron extent="suff">-ə-lē</pron>
@@ -80,7 +80,7 @@ $Id$
           <gramGrp>
             <pos>adv</pos>
           </gramGrp>
-        </re>
+        </entry>
       </entry>
     </egXML>
   </exemplum>
@@ -97,7 +97,7 @@ $Id$
         <sense>
           <def>Faire fonctionner un siphon, transvaser (un liquide) au moyen d'un siphon.</def>
         </sense>
-        <re>
+        <entry type="relatedEntry">
           <form>
             <orth>siphonnage</orth>
             <pron extent="suff">[as]</pron>
@@ -105,7 +105,7 @@ $Id$
           <gramGrp>
             <pos>n.</pos>
           </gramGrp>
-        </re>
+        </entry>
       </entry>
     </egXML>
   </exemplum>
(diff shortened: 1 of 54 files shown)


laurentromary:
I have checked this pull request and it all seems like we want it. Thanks!

P5/Source/Specs/re.xml
```diff
@@ -7,26 +7,34 @@ $Date$
 $Id$
 -->
 <?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
-<elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="dictionaries" ident="re">
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0" module="dictionaries" ident="re"
+  validUntil="2026-03-10">
+  <desc xml:lang="en" type="deprecationInfo" versionDate="2023-01-10"> Because an <gi>entry</gi> can
+    now occur inside an <gi>entry</gi>, the <gi>re</gi> element is no longer needed. These
+    Guidelines recommend using an <gi>entry</gi> with a <att>type</att> attribute value of
+      <val>related</val> instead.</desc>
```
HelenaSabel:
```suggestion
      <val>relatedEntry</val> instead.</desc>
```

HelenaSabel (changes requested):
My first comment concerns commit 61247e54d5412e5f1a2e46adb76ef92879c66b3a: I think it might have been included in this PR by mistake.

Then I made a suggestion so that the recommended value for `@type` in the deprecation note matches the examples. @trishaoconnor: if you agree you can click on “commit suggestion” :)

trishaoconnor:
Thank you @HelenaSabel for spotting these mistakes. My apologies for not recognising that the commit was added to this PR. That was an accident.  I have clicked "commit suggestion" so the deprecation note now matches the examples.

HelenaSabel:
@trishaoconnor: do you want to try to delete commit https://github.com/TEIC/TEI/commit/61247e54d5412e5f1a2e46adb76ef92879c66b3a? I think you could try:
```
git revert --strategy resolve 61247e54d5412e5f1a2e46adb76ef92879c66b3a
```
That will delete the commit (without altering the history). But if @sydb, who is assigned to this PR, thinks it’s ok to merge it as it is, fine with me :)

sydb:
Yes, the global addition of `mode="add"` was added to this PR by mistake. But since it is a separate commit, and has no effect on anything anyway, I would just not worry about it. (Except to _avoid_ a “squash and merge”.)
But that may just be me being lazy — feel free to separate to a separate PR if y’all think it important.

sydb:
An attempt to ping @GusRiva on the TEI repo. Please ignore.

HelenaSabel (approved):

PR #2525, opened 2024-01-20
Use reference to model.eventLike, not direct references to its members

sydb:
This is a pretty non-problematic change. Passes all tests on my system. I think _anyone_ can review, and once there has been at least one positive review _anyone_ can pull.


raffazizzi:
Tests ran without error on my machine. Code looks straightforward. Merging.

Issue #2514, opened 2023-12-15
time to update header comments?


sydb:
I was getting annoyed at seeing old Subversion keywords not being substituted in our header comments. And then I noticed someone had (in a few files) put them all on one line. How much better!

joeytakeda:
Any reason why those SVN keywords can't just be removed entirely?

sydb:
No, @joeytakeda no reason not to remove them entirely. (That was the plan.)

Back in December I tried fixing this (in a branch[1]). But using an XML tool on our source files is a bit problematic, because at times there are whitespace consequences we would like to avoid. So the results are _almost_ right, but not quite, and merging that branch now would be a big headache.

So today I took another look. First, I checked and verified that the comments we are interested in replacing are _always_ a child of the document node, and _never_ have any preceding siblings (i.e., are always the 1st child of the document node); furthermore _none_ of them contain any of the characters `>`, `!`, `@`, `#`, `%` or TAB (U+0009).

I then wrote a C program (which I will attach later if anyone is interested[2]) to read in a TEI file and write out the same file with the whitespace of the 1st comment regularized. It regularized newline to `!`, carriage return to `@`, and compresses strings of blanks. (I never ended up using `#` or `%`.) It does not really matter what regularization it performed, it just had to make sure the 1st comment was all on one line, so it could be easily processed. In truth it has a bug. It outputs `-->-->` at the end of the comment instead of just `-->`. I started to debug it and then thought “What am I doing? It does not matter at all what it ends in, I am just going to replace it (using Perl), anyway.”.

So I ran the C program on all 877 source files, and then used a Perl one-liner to replace all the initial comments to a new copyright comment. Both of these processes were _very_ fast, < 2 s each. Which is a good thing, because I had to repeat the process when I realized I had spelled “licenses” wrong. The new copyright comment is “&lt;!-- © TEI Consortium. Dual-licensed under CC-by and BSD2 licenses; see the file COPYING.txt for details. -->”.

[...]

PR #2526, opened 2024-02-09
update copyright notice

sydb:
Per #2514, updated the copyright notice to a shorter comment that fits on 1 line in the Source/Specs/\*.xml, Source/Guidelines/en/\*.xml, and Source/Guidelines/fr/HD-Header.xml files.

For a detailed discussion of what I did, see post on the issue itself.


ebeshero (approved):
Looks good for removing the extra line breaks and notation, short of reducing the slightly repetitive wording itself.

trishaoconnor (commented):
The changes look good to me. The updated copyright is more concise and it looks tidier having the notice on one line.

sydb:
@ebeshero — Feel free to change to or suggest different wording. Now that it is a completely consistent string all on one line, it is trivial to change. (I would use Perl, but sed, awk, or any good text editor like oXygen, BBEdit, Emacs, vim, or Sublime, could do this easily.)

ebeshero:
I double-checked the British English spelling of verb (license) and noun (licence), and it's all fine. Glad you put back the XML declaration (and found more such to missing lines to address).

PR #2532, opened 2024-03-07
Updated prose and added deprecation information for ´superEntry´ element

trishaoconnor:
Added deprecation information to the `<superEntry>` element spec page and updated the prose and examples related to the `<superEntry>` element accordingly.


torstenroeder (approved):
We reviewed this together and think it is great now.

Issue #2371, opened 2022-11-16
Multiple `<attDef>`s with the same `@ident` in the same `<attList>` should be invalid


martindholmes:
In the ATOP work, we came across an in-vivo ODD file which has this content:

~~~xml
<attList>
  <attDef ident="level" mode="change">
    <valList mode="add" type="closed">
      <valItem ident="j"/>
      <valItem ident="m"/>
    </valList>
  </attDef>
  <attDef ident="type" mode="change">
    <valList mode="add" type="closed">
      <valItem ident="italic"/>
    </valList>
  </attDef>
  <attDef ident="xml:id" mode="delete"/>
  <attDef ident="n" mode="delete"/>
  <attDef ident="facs" mode="delete"/>
  <attDef ident="key" mode="delete"/>
  <attDef ident="type" mode="change">
    <valList mode="add" type="closed">
      <valItem ident="italic"/>
    </valList>
  </attDef>
</attList>
~~~~~

The repetition of the `@type` attDef is clearly an error, but it's not invalid against tei_odds or tei_all. I propose introducing a Schematron rule to prevent this. In cases where multiple definitions are not identical, then a significant burden would be placed on a) maintainers of the ODD specification to specify exactly how they bear on each other and how they are expected to be combined during processing, and b) maintainers of processing tools. Forcing ODD writers to consolidate their modifications to a single attribute into a single `<attDef>` seems far preferable. A simple Schematron rule would constrain this.

lb42:
I am not sure I agree with this, though it's an obvious optimisation that any implementer might choose to make, and I have no quarrel with your doing so in ATOP. But for the record, the reason it's like that is that I at least thought you might well want to take several bites at the definition of an attribute in an ODD chaining context. For example, you might define a base set of @type  values in your "mother" ODD, and then add different sets of extra values in ODDs which are chained to it (or do I mean from it). Yes, of course you risk shooting yourself in the foot by making contradictory assertions for the same attribute, but that happens elsewhere too. If you're clever enough to chain your ODDs you're probably clever enough to notice such errors.

sydb:
@lb42 —
Your argument seems quite reasonable to support a `…/elementSpec[@ident eq 'duck']/attList/attDef[@ident eq 'quack'][@mode eq 'change']` in both the ODDs for customization A and for customization B, which might well be chained in one direction or the other. But (I think) @martindholmes is arguing against having 2 cases of that XPath _in the same `<schemaSpec>`_.

[...]

PR #2534, opened 2024-03-14
no duplicate `<attDef>`s allowed

sydb:
Address #2371 with a Schematron constraint that flags as an error any 2+ `<attDef>`s that both share an ancestor `<attList>` and have the same `@ident`  (regardless of the `@mode`) unless they are both (all) children of an `<attList>` that has an `@org` attribute value of "choice".

The file detest.odd has been updated to test this reasonably well.

If tests pass, I personally do not think any deprecation period is necessary. (What would it deprecate, anyway?) The error this new constraint catches is an error.

Reviewers: If you actually look at the code differences the most important bit (by far) to look at is the new lines 24–76 (in green) of attList.xml.


martindholmes:
@sydb Does this take account of namespaces? You could correctly have two atts with the same `@ident` but in different namespaces, no? Forgive me if I've missed it; I'm reading on my phone.

sydb:
No, it does not. For some reason I guess I thought we had decided to look at _only_ `@ident` (not +`@ns`), but that would be silly.
Will try to fix v. soon.

sydb:
Fixed, and added more tests to detest.

sydb:
One possible concern is that the comments I put in the Schematron show up in the _Guidelines_. Some folks will like that, others won’t.

martindholmes:
@sydb I know you will think this is trivial, but I don't think we should use YouTube URLs as spurious namespaces in tests. I don't think it's polite to commandeer someone else's domain for this sort of purpose, and I would also be concerned that any controversy or conflict arising in relation to one of these URLs in future might draw the TEI into something it has no connection with. I would prefer to use something like http://www.tei-c.org.ns/test/eg1, 2, 3 and so on. If you have no strong preference for keeping these YouTube URLs, I'll happily add a fix to the PR for this.

sydb:
1. Rebased current dev into this branch
2. Addressed concern about namespace URIs, above
3. Tested, passes all tests on my local system (except NVDL and XeLaTeX do not work on my system, so are untested). Only error is the 3 `<superEntry>` warnings that have nothing to do with this PR.

ebeshero:
@martindholmes Our VF2F subgroup agreed, and our suggestion was just to make up a portmanteau URL, like`https://teitube.org` or some such. Re-reading now, your ns suggestion is probably more befitting to the task.

martindholmes (commented):
Apologies for being slow on this. I keep looking at the Schematron rule and realising that I don't quite understand it, and I need to go away and write some specific tests. I don't get what the notanamespace thing is doing.

sydb:
I think the reason $notanamespace exists is because I wanted to compare `@ns eq $ad/@ns`. I was either worried that it was not legal to perform the comparison if one or both did not exist, or I was worried that if `@ns` is "ab" and `@ident` is "br", then just comparing `@ns||@ident` would erroneously succeed when `$ad/@ns` was missing and `$ad/@ident` is "abbr".

Either way, though, on reading it now, since the comparisons are done separately (not combined as in para above), we might be able to get away without the $notanamespace. 

But I tried removing the $notanamespace bit and just comparing `@ns eq $ad/@ns`, and the tests fail because in "no_duplicate_attrs_2_valid" the attribute definition for `@fun` is incorrectly flagged as a duplicate. (It is not a duplicate because it is in an alternation, i.e. its parent `<attList>` has `org="choice"`.)

If I were actually writing XSLT I would probably do this in multiple steps. First take $defs (the sequence of all `<attDef>`  descendants) and generate a sequence that, for each one, is the namespace & ident in Q-notation. Then a) all comparisons are just simple string comparisons and b) it is easy to generate Q-notation output. But I am not sure that is reasonable to do in Schematron alone.

ebeshero:
Reviewing this aging PR, I'm wondering what remains to do here once we update the branch? I hope we resolve this for the next release!

sydb:
OK. We really should not have let this fester for 11 months. (I take much of the blame for that, BTW.) Made it quite difficult to bring it up to date with current dev branch and to figure out what the BLEEP is being done. Luckily, I left reasonably detailed comments.

They are, however, the problem. That is, I _think_ this branch is ready to go, except for the issue of the comments. It still requires review and testing, of course, this is complicated stuff.

But the comments require a decision by Council.

This is some pretty complicated Schematron code. (I suspect, but do not know, that it would be somewhat simpler using the upcoming next version of Schematron, but I do not think it would be simple enough that anything in this discussion about comments would be different.) The good news is the comments therein make it somewhat easier to understand. The bad news is those comments get reproduced in the HTML output of the _Guidelines_. Some folks do not think the comments should be there.

Possible ways to address this concern (in a numbered list just so we can refer to them, no order of preference implied, as I have not made up my mind):

1. Leave things as they are — users should see the comments on the tagdoc pages. (We might want to make them prettier, though.)
2. Continue to use the comments in the source, change Stylesheets so that comments within a `<constraint>` are not reproduced on the tagdoc pages.
3. Remove the comments from the source. In which case, we probably want to add a big block comment **outside** the `<constraint>` that tries to explain what is going on in there. This makes both writing and using the comments somewhat harder, but does solve the problem.
4. Change the Stylesheets so that comments within a `<constraint>` are not reproduced on the tagdoc pages, and then start using the `<sch:p>` element to document things inside a `<constraint>` when we do want the documentation to show up in the tagdoc page. NOTE: OXygen says the `<sch:p>` element is valid; our schema says it is not. So we would have to investigate and fix that, first.
5. Invent a mechanism to allow the ODD author to say whether any particular comment should show up in the output or not. (E.g., a `<tei:comment>` element with an `@includeInDocumentation` attribute.)

sydb:
Council decides this PR is blocked awaiting https://github.com/TEIC/Stylesheets/issues/746.

ebeshero:
Just updating the comments here: No longer blocked because we resolved the method of excluding comments from the Guidelines output. So we're now testing whether this constraint works to target duplicate `<attDef>` elements with the same `@ident` in the same `ancestor::attList[not(@choice)]` and not in different namespaces. Reminding myself of that, I'll try to test the PR now.

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+            <attList>
+              <attDef ident="bad" ns="https://www.teitube.com/watch?v=enuOArEfqGo"/>
+              <attDef ident="good" ns="https://www.teitube.com/watch?v=6YdkQWN59HY"/>
+              <attDef ident="better" ns="https://www.teitube.com/watch?v=qSzYchDIk7E"/>
+              <attDef ident="best" ns="https://www.teitube.com/watch?v=klsEniiVg74"/>
+              <attDef ident="bad" ns="https://www.teitube.com/watch?v=Qikp08PPBf0"/>
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-15">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
```
ebeshero:
```suggestion
              attribute definitions, i.e. <name>no_duplicate_attrs</name>.</p>
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+              <p>This <gi>attList</gi> should be valid because
+              although there are two <att>bad</att> attributes defined
+              as siblings, they have different namespaces.</p>
+              <p>Note that because this element is not a member of any
+              model class nor in any other element’s content model, it
+              will probably not show up in an output schema, which is
+              a good thing, becase defining an attribute twice like
```
ebeshero:
```suggestion
              a good thing, because defining an attribute twice like
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+            <attList>
+              <attDef ident="bad"/>
+              <attDef ident="good"/>
+              <attDef ident="better"/>
+              <attDef ident="best"/>
+              <attDef ident="bad"/>
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-13">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
```
ebeshero:
```suggestion
              attribute definitions, i.e. <name>no_duplicate_attrs</name>.</p>
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+                <attDef ident="fun"/>
+                <attDef ident="fun"/>
+              </attList>
+              <attDef ident="til"/>
+              <attDef ident="her"/>
+              <attDef ident="daddy"/>
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-13">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
```
ebeshero:
```suggestion
              attribute definitions, i.e. <name>no_duplicate_attrs</name>.</p>
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+              <attDef ident="Inside"/>
+              <attList org="choice">
+                <attDef ident="my"/>
+                <attDef ident="your"/>
+              </attList>
+              <attDef ident="mind"/>
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-13">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
```
ebeshero:
```suggestion
              attribute definitions, i.e. <name>no_duplicate_attrs</name>.</p>
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+              alternation with one another (nor are the two
+              <att>your</att>s).</p>
+              <p>Note that the <att>Your</att> vs <att>your</att> and
+              <att>The</att> vs <att>the</att> are not problems at all
+              — XML is completely case sensitive, those are different
+              attributes.</p>
+              <p>Note that because this element is not a member of any
+              model class nor in any other element’s content model, it
+              will probably not show up in an output schema, which is
+              a good thing, becase defining an attribute twice like
```
ebeshero:
```suggestion
              a good thing, because defining an attribute twice like
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-13">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
+              <p>This <gi>attList</gi> should be invalid because there
+              are two <att>bad</att> attributes defined as siblings.</p>
+              <p>Note that because this element is not a member of any
+              model class nor in any other element’s content model, it
+              will probably not show up in an output schema, which is
+              a good thing, becase defining an attribute twice like
```
ebeshero:
```suggestion
              a good thing, because defining an attribute twice like
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+              attributes defined that are not in alternation, they are
+              in different namespaces.</p>
+              <p>Note that the <att>Your</att> vs <att>your</att> and
+              <att>The</att> vs <att>the</att> are not problems at all
+              — XML is completely case sensitive, those are different
+              attributes.</p>
+              <p>Note that because this element is not a member of any
+              model class nor in any other element’s content model, it
+              will probably not show up in an output schema, which is
+              a good thing, becase defining an attribute twice like
```
ebeshero:
```suggestion
              a good thing, because defining an attribute twice like
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+              <attDef ident="Inside"/>
+              <attList org="choice">
+                <attDef ident="my" ns="https://www.teitube.com/watch?v=FKHMRGMkQHU"/>
+                <attDef ident="your" ns="https://www.teitube.com/watch?v=TQGQgiKDBbU"/>
+              </attList>
+              <attDef ident="mind"/>
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-15">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
```
ebeshero:
```suggestion
              attribute definitions, i.e. <name>no_duplicate_attrs</name>.</p>
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+                  <attDef ident="the"/>
+                  <attDef ident="sky"/>
+                </attList>
+                <attDef ident="is"/>
+              </attList>
+              <attDef ident="gray"/>
+            </attList>
+            <remarks xml:lang="en" versionDate="2024-03-13">
+              <p>Here we are testing the Schematron that tests for duplicate
+              atttribute defintions, i.e. <name>no_duplicate_attrs</name>.</p>
```
ebeshero:
```suggestion
              attribute definitions, i.e. <name>no_duplicate_attrs</name>.</p>
```

P5/Test/detest.odd
```diff
@@ -548,6 +559,253 @@
...
+              <p>This <gi>attList</gi> should be invalid due to the
+              two definitions of the <att>the</att> attribute, which
+              are not in alternation.</p>
+              <p>Note that because this element is not a member of any
+              model class nor in any other element’s content model, it
+              will probably not show up in an output schema, which is
+              a good thing, becase defining an attribute twice like
```
ebeshero:
```suggestion
              a good thing, because defining an attribute twice like
```

P5/Source/Specs/attList.xml
```diff
@@ -21,6 +21,66 @@
...
+        -->
+        <sch:let name="nsidents"
+          value="for $ad in $defs return
+                   if ( $ad[
+                          parent::tei:attList[ @org eq 'choice']
+                          and
+                          preceding-sibling::tei:attDef[
+                            @ident eq $ad/@ident
+                            and
+                            ( @ns, $notanamespace )[1] eq ( $ad/@ns, $notanamespace )[1]
```
ebeshero, line 62: @sydb This part seems a little abstruse, but it's ultimately readable when we see that `$notanamespace` is just a bit of string. I guess it saves you from having to write an extra set of conditions for namespaced attribute or not, and somehow it makes sense for coping with the situation of the same attribute name with and without a namespace. And I can definitely verify that it works on a test ODD I just wrote.

ebeshero (approved):
@sydb I've reviewed the code, made a bunch of tiny prose edits on the detest file in GitHub's suggest mode that we can just review and merge in via the web interface. I can verify that this passes tests, doesn't break builds, and when I run exemplars and work with a generated TEI-All schema, it works to validate an ODD with a bunch of different test cases for attributes of the same `@ident` value with and without namespaces. Looks good to me.

P5/Source/Specs/attList.xml
ebeshero: I'm just committing my suggestions since they're pretty tiny and just prose edits.

PR #2538, opened 2024-03-17
Add `interleave` element (and deprecate `sequence/@preserveOrder`) 

### LINKED ISSUE #2154: Create an interleave element
Currently `@preserveOrder` on `<sequence>` is an optional attribute and the spec states: "if true, indicates that the order in which component elements of a sequence appear in a document must correspond to the order in which they are given in the content model."  However, the default behaviour for `<sequence>` is that order is preserved unless `@preserveOrder` is false, but the current description makes it seem that the opposite is true (i.e. that order doesn't matter by default); plus, ​setting to `false` hasn't worked since at least 2017 per https://github.com/TEIC/Stylesheets/issues/241.  

Since `@preserveOrder` is implicitly true, I think that it should gain a default value of true so that it states explicitly what's already happening; I would also suggest that the description be changed to reflect that "true" is default and also outline what happens when its value is "false".

joeytakeda:
Resolves #2154 by adding the `interleave` element and deprecating `sequence/@preserveOrder`

At the moment, I'm leaving this as draft as per @hcayless' [observation](https://github.com/TEIC/TEI/issues/2154#issuecomment-1427146673) that we should figure out how best to validate the defined content model to prevent overlap. IMO, we should probably warn about this in the prose and in the remarks for the element, but true validation will probably need to be left up to the ODD processor.


joeytakeda:
Moving this to be ready for review, since I think the question of validity is probably something that needs to be handled by the processor (but have assigned mostly ATOP folks for review to see what they think about that)

sydb:
Just to make the concern initially raised by @hcayless (on the ticket) echoed by @joeytakeda (here) explicit …

In RELAX NG it is not legal to have the same element occur in an interleave more than once. And because interleave “reaches in” to the patterns listed, this is hard to check for (other than by running a RELAX NG processor like `jing` … I just discovered that neither `trang` nor `nxml-mode` object).

So, for example, the following is **not allowed** in RELAX NG:
~~~
start = element blood { ( a & b & o & ab ) }

a = element a { empty }
b = element b { empty }
o = element o { empty }
ab = ( a | b )
~~~
And it does not matter whether the `ab` pattern is defined as an alternation (`<rng:choice>` or `|`), a sequence (`<rng:group>` or `,`), or an interleave (`<rng:interleave>` or `&`). In all cases the reference to the pattern `ab` throws an error. (The message is “the element "b" can occur in more than one operand of "interleave"”, which  I, for one, find confusing.)

P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -634,18 +634,20 @@ to mark any technical term, thus:
...
+        <specDesc key="interleave"/>
       </specList>
-      These two wrapper elements are also members of <ident
+      These three wrapper elements are also members of <ident
       type="class">att.repeatable</ident>. References listed as
       children of <gi>sequence</gi> must appear in the order and
-      cardinality specified. Only one of the references listed as
-      children of <gi>alternate</gi> may appear, although the
+      cardinality specified whereas references listed as children
+      of <gi>interleave</gi> may appear in any order. Only one of the
```
sydb, line 635: Can-of-worms department.

(Note to @joeytakeda — most of the problem here is not from your improvement, but has been here all along and is just foregrounded by your change.)

This sentence does not mention that cardinality of the stuff inside `<interleave>`. But on thinking of how to re-word it, I realize that these sentences discuss the cardinality of _references_, when what we are talking about is the cardinality of _the thing referenced_. I am not sure what, if anything, we should do about this.

P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -698,6 +701,56 @@ to mark any technical term, thus:
           </alternate>
         </content>
       </egXML>
+      In the above example, the alternation of element pairs must be
+      encoded in the order specified (i.e. <gi>sic</gi> and then
```
sydb: The “i.e.” should be “e.g.” or, better yet, “for example”.

P5/Source/Specs/interleave.xml
```diff
@@ -0,0 +1,39 @@
...
+          <elementRef key="persName"/>
+          <elementRef key="note" minOccurs="0"/>
+          <alternate minOccurs="0">
+              <elementRef key="placeName"/>
+              <elementRef key="location"/>
+          </alternate>
+        </interleave>
+      </content>
+    </egXML>
+    <p>This example describes a content model that consists of a mandatory <gi>persName</gi> element, an optional note element, and, optionally, one and only one of <gi>placeName</gi> or <gi>location</gi> in any order.</p>
```
sydb: “note” should be `<gi>note</gi>`.

P5/Source/Specs/sequence.xml
```diff
@@ -19,7 +19,8 @@
     </constraint>
   </constraintSpec>
   <attList>
-    <attDef ident="preserveOrder">
+    <attDef ident="preserveOrder" validUntil="2025-03-15">
```
sydb: Given that a) `<interleave>` did not make it into the recent release, and b) use of `@preserveOrder` still works in the Stylesheets, I think the `@validUntil` date should be moved up to late 2025 or early 2026.

sydb (approved):
I would prefer the issues I note be addressed before merge, but it is not a big deal. (And I may get to the easier ones myself, soon.)
That said, we do need a discussion on the fact that use of `<interleave>` in an ODD means that the RELAX NG generated by either the Stylesheets or by ATOP will not be transformable into DTD language.

P5/Source/Specs/interleave.xml
sydb: Fixed in b5f31ff5b.

P5/Source/Guidelines/en/TD-DocumentationElements.xml
sydb: Fixed in b5f31ff5b.

tuurma:
F2F discussion in Buenos Aires: We should provide a `<remark>` in the `<interleave>` element spec to caution people working with elements belonging to multiple classes, or when using `classRef`.

raffazizzi:
Subgroup @sydb @joeytakeda @raffazizzi suggest also creating a Stylesheet ticket to check validity of rng before continuing build in order to catch validation errors caused by `interleave/classRef` (and maybe more!)

joeytakeda:
Stylesheets ticket created here: https://github.com/TEIC/Stylesheets/issues/719  — for discussion at the next council meeting is whether the approval of this PR (e.g. the additional of interleave) is contingent on the validation process.

martindholmes (approved):
@sydb and I discussed this at length today, and we would like the Council to check that what we have said about the risks of using interleave when the output target is XSD is actually detailed enough to be helpful. We could not find any authoritative reference to cite which explains the problem in detail, although many articles assume readers already understand it.

PR #2540, opened 2024-03-19
add missing `<sch:rule>` elements

sydb:
Title says it all.

### DIFF
--- a/P5/Test/testmeta2010.odd
+++ b/P5/Test/testmeta2010.odd
@@ -8,8 +8,8 @@
             <title>Testing new ODD features added spring 2010</title>
          </titleStmt>
          <publicationStmt>
-	   <p> </p>
-	 </publicationStmt>
+           <p> </p>
+         </publicationStmt>
          <sourceDesc>
             <p>authored from scratch</p>
          </sourceDesc>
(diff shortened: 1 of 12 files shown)


P5/Test/testmeta2010.odd
```diff
@@ -20,176 +20,177 @@
...
-		 <valItem ident="good"/>
-		 <valItem ident="bad"/>
-	       </valList>
-	     </attDef>
-	   </attList>
-	 </classSpec>
+         <elementSpec ident="list" mode="change">
+           <attList>
+             <attDef ident="type" mode="change">
+               <constraintSpec mode="add" ident="a1" scheme="schematron">
```
martinascholger: European subgroup wonders why the constraintSpec is within the attList/attDef and not a child of elementSpec

martinascholger (approved):

martinascholger:
@sydb and @martinascholger will resolve the conflicts and merge it

P5/Test/testmeta2010.odd
sydb: I have answer, but not a _good_ answer.
The difference was that declaring the constraint specification within the attribute definition (as opposed to as a direct child of the element specification) meant that it was fired only for `<list>` elements that actually had a `@type` attribute, regardless of that attribute’s value. That has now been duplicated by using a `@context` of `"tei:list/@type"`.
What I cannot explain is _why_ anyone thought it was import to fire this test only when a `<list>` has a `@type`. Seems nonsensical to me. (Maybe the author wanted to test that it indeed only fired when there was a `@type`?)

PR #2545, opened 2024-04-11
Cosmetic changes: unify order of attributes

rettinghaus:


sydb (approved):
`make clean validate html-web test exemplars` works as expected on my system.

PR #2546, opened 2024-04-11
Add German translations

rettinghaus:
This PR provides some German translations.

### DIFF
--- a/P5/Source/Specs/acquisition.xml
+++ b/P5/Source/Specs/acquisition.xml
@@ -4,6 +4,7 @@
   <gloss xml:lang="en" versionDate="2007-06-12">acquisition</gloss>
   <gloss xml:lang="es" versionDate="2022-06-16">adquisición</gloss>
   <gloss versionDate="2007-06-12" xml:lang="fr">acquisition</gloss>
+  <gloss versionDate="2024-04-11" xml:lang="de">Akquisition</gloss>
   <desc versionDate="2019-01-17" xml:lang="en" xml:id="acquisn.desc">contains any descriptive or other information
 concerning the process by which a manuscript or manuscript part or other object entered the holding
 institution.</desc>
--- a/P5/Source/Specs/acquisition.xml
+++ b/P5/Source/Specs/acquisition.xml
@@ -4,7 +4,7 @@
   <gloss xml:lang="en" versionDate="2007-06-12">acquisition</gloss>
   <gloss xml:lang="es" versionDate="2022-06-16">adquisición</gloss>
   <gloss versionDate="2007-06-12" xml:lang="fr">acquisition</gloss>
-  <gloss versionDate="2024-04-11" xml:lang="de">Akquisition</gloss>
+  <gloss versionDate="2024-04-11" xml:lang="de">Akquise</gloss>
   <desc versionDate="2019-01-17" xml:lang="en" xml:id="acquisn.desc">contains any descriptive or other information
 concerning the process by which a manuscript or manuscript part or other object entered the holding
 institution.</desc>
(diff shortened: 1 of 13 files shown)


P5/Source/Specs/acquisition.xml
```diff
@@ -4,6 +4,7 @@
   <gloss xml:lang="en" versionDate="2007-06-12">acquisition</gloss>
   <gloss xml:lang="es" versionDate="2022-06-16">adquisición</gloss>
   <gloss versionDate="2007-06-12" xml:lang="fr">acquisition</gloss>
+  <gloss versionDate="2024-04-11" xml:lang="de">Akquisition</gloss>
```
sabineseifert: In the context of manuscripts/objects entering a collection, I only know this as "Akquise" (although the word "Akquisition" indeed exists in German).

P5/Source/Specs/provenance.xml
```diff
@@ -3,6 +3,7 @@
 <?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?><elementSpec xmlns="http://www.tei-c.org/ns/1.0" xmlns:sch="http://purl.oclc.org/dsdl/schematron" module="msdescription" xml:id="PROVENANCE" ident="provenance">
   <gloss xml:lang="en" versionDate="2007-06-12">provenance</gloss>
   <gloss versionDate="2007-06-12" xml:lang="fr">provenance</gloss>
+  <gloss versionDate="2023-08-08" xml:lang="de">Provenienz</gloss>
   <desc versionDate="2019-01-17" xml:lang="en" xml:id="prov.desc">contains any descriptive or other information
 concerning a single identifiable episode during the history of a manuscript, manuscript part, or other object after its creation but before its acquisition.</desc>
```
sabineseifert, line 8: The translation of the gloss is correct but I wonder a bit about the definition: Why is the phrase "concerning a single identifiable episode" in there? Provenance means the whole history of a manuscript/object (that, yes, consists of several single identifiable episodes). I am also not sure about the emphasis on manuscripts as they are one example of so many objects that can have a provenance.
Should we do something about this definition?

sabineseifert (approved):

P5/Source/Specs/acquisition.xml
rettinghaus: Yes, you're right. "Akquise" is the  better term here.

P5/Source/Specs/provenance.xml
lb42: The tei element provenance by definition describes a single significant episode in the life of an object. The element for describing several such episodes is history.

sabineseifert: In German "Provenienz" means the whole history of one object. Therefore, to be correct, one should translate the TEI element `history` with the gloss "Provenienz" (which might be confusing) and the TEI element `provenance` with --? I am not sure.

lb42: tei:history has three components : information about the origin of the item, information about its acquisition by  the place where it now resides, and information about anything that happened between the two. Is "Provenienz" an appropriate German term for all three of these?

rettinghaus: "Provenienz" is in German often used as term for the collection of all provenances, but still it only means a "single identifiable episode" in the object's history.

schassan: I would agree: "Provenienz" could mean both the whole history of an object as well as a single identifiable episode. In German we would distinguish between "the provenance" and "a provenance". To complicate it even further, in manuscript cataloguing, "Provenienz I" represents the origin, whereas "Provenienz II" represents the single identifiable episodes.

sabineseifert: Suggestion: Perhaps we can translate `history` with 'Provenienzgeschichte' and `provenance` with 'Provenienz'?

rettinghaus: Maybe, but that wouldn't affect this PR, would it?

sabineseifert: This is true as I am now ok with translating `provenance` with "Provenienz" and already approved all other translations. :-)

Perhaps @martinascholger or @schassan will also have a look at them?

peterstadler:
This looks like an improvement to me and all conversations have been resolved. Will merge

Issue #2543, opened 2024-03-27
Need Schematron to constrain attribute co-occurrence on `<elementRef>`


martindholmes:
Arising from ATOP work by @sydb, @HelenaSabel and myself:

`<elementRef>` does two distinct jobs: it can effectively import an elementSpec from another source using something like this:

~~~~~
<elementRef key="p" source="tei:1.2.1"/>
~~~~~

or it can reference an `<elementSpec>` somewhere in the ODD or imported from another source, to use it in a content model.

When `<elementRef>` is a child of `<schemaSpec>` or `<specGrp>`, it is doing the former job, and therefore it should only be allowed to have the `@source` attribute and not the `@minOccurs` and `@maxOccurs` attributes.

Conversely, when it's a child of a content model element such as `<content>` or `<sequence>`, it should be allowed to have `@minOccurs` and `@maxOccurs`, but not `@source`. 

We believe these constraints should be expressed in Schematron.

sydb:
So to be precise, we are thinking that `(schemaSpec|specGrp)/elementRef` should not be allowed to have `@minOcccurs` or `@maxOccurs`, and that `content//elementRef` should not be allowed to have `@source`.

I do not think a deprecation period is in order, as these constructions strike me as nonsensical. That said, I have not actually tested to see if the current Stylesheets produce anything useful when this occurs.

sydb:
Thus something like
~~~xml
  <sch:pattern>
    <sch:rule context="tei:elementRef[ parent::schemaSpec | parent::specGrp ]">
      <sch:report test="@minOccurs | @maxOccurs">An element reference is not repeatable when part of a schema specification (and thus this &lt;elementRef> should not have @minOccurs or @maxOccurs). </sch:report>
    </sch:rule>
    <sch:rule context="tei:content//tei:elementRef">
      <sch:report test="@source">An element reference within a content model must refer to a locally defined element specification (and thus this &lt;elementRef> should not have @source).</sch:report>
    </sch:rule>
  </sch:pattern>
~~~

martindholmes:
I would do those as assertions, not reports. I think these would be errors.

sydb:
[...]

PR #2552, opened 2024-05-02
constrain attributes of elementRef

sydb:
Added constraint to limit the use of attributes of the elementRef element per #2543.


sydb:
Added tests (in detest.odd), and lo and behold those tests caught an error! (I had forgotten the “tei:” prefix in one of the `@test` attributes.)
Fixed in [bd1570a](https://github.com/TEIC/TEI/pull/2552/commits/bd1570a027c3dca145d7b04dbf2573e6f50f39e4), at least it passed the tests on my system.

raffazizzi (approved):

PR #2555, opened 2024-05-16
Adding xenoData to standOff as per #2436

### LINKED ISSUE #2436: standOff should be allowed to contain xenoData
I think that `<xenoData>` should be allowed to appear inside `<standOff>`. My reasoning for this is two-fold:

1. `<standOff>` is "a container element for linked data, contextual information, and stand-off annotations" and many LOD projects are using `<xenoData>` for some of their processing workflows and don't create proper TEI structures for this. However, they sometimes mix and match and have some annotation in `<xenoData>` and some in `<standOff>`. So it would be useful for processing to be able to extract just the `<standOff>` containing both of those in a single container element. This is an appeal to convenience.   

2. Some of those using `<xenoData>` are storing non-TEI standoff data, which is pointing locally into the document, e.g. with web annotations data model. While in an ideal world they would store this using `<annotation>` inside `<standOff>`, it is unlikely they'll change their processing. So in storing this information, it would be good if it could appear in `<standOff>` but it will remain `<xenoData>`, as that is where information like this belongs in the TEI abstract model. This is an appeal to semantics. 

Proposal: 

- Have `<xenoData>` claim membership in model.standOffPart by adding `<memberOf key="model.standOffPart"/>` just after:  https://github.com/TEIC/TEI/blob/2075569ce78d975f4d160696c49b7d07083524c7/P5/Source/Specs/xenoData.xml#L19
- Adding a sentence such as "`The <gi>xenoData</gi> element is also available inside the <standOff> element for those using it for stand-off markup.`" at the end of the paragraph at  https://github.com/TEIC/TEI/blob/2075569ce78d975f4d160696c49b7d07083524c7/P5/Source/Guidelines/en/HD-Header.xml#LL2528C30-L2528C30

jamescummings:
Added xenoData to standOff as per #2436  with changes in xenoData spec and brief mention where xenoData is discussed in the HD chapter.

### DIFF
--- a/P5/Source/Guidelines/en/HD-Header.xml
+++ b/P5/Source/Guidelines/en/HD-Header.xml
@@ -2519,7 +2519,11 @@ grouping makes it easy to avoid accidentally flagging non-TEI data as errors
 during validation of the file against a TEI schema. The
 <gi>xenoData</gi> element, which may appear in the TEI header after
 the <gi>fileDesc</gi> but before the optional <gi>revisionDesc</gi>,
-is provided for this purpose.
+is provided for this purpose. Where the use of <gi>xenoData</gi> is primarily
+to store linked data, contextual information, or stand-off annotations 
+that refer to the contents of the TEI file itself,
+then it is also possible to place <gi>xenoData</gi> inside the <gi>standOff</gi> 
+element. 
 <specList>
   <specDesc key="xenoData"/>
 </specList></p>
--- a/P5/Source/Guidelines/en/HD-Header.xml
+++ b/P5/Source/Guidelines/en/HD-Header.xml
@@ -2521,9 +2521,9 @@ during validation of the file against a TEI schema. The
 the <gi>fileDesc</gi> but before the optional <gi>revisionDesc</gi>,
 is provided for this purpose. Where the use of <gi>xenoData</gi> is primarily
 to store linked data, contextual information, or stand-off annotations 
-that refer to the contents of the TEI file itself,
-then it is also possible to place <gi>xenoData</gi> inside the <gi>standOff</gi> 
-element. 
+that refer to the contents of the TEI file itself, the
+<gi>xenoData</gi> element may be placed inside the <gi>standOff</gi> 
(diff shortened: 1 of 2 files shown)


sydb (commented):
Passed all tests on my system (which means no NVDL or XeLaTeX tests), and it “works” in that `<xenoData>` is an allowable child of `<standOff>`. So I approve this PR.

That said, though, I wonder if we couldn’t do a bit better on the wording of the prose in HD, and if we shouldn’t mention something about it in 16.10 “The `<standOff>` Container” (#SASOstdf)? I will see if I can make a detailed suggestion for the former, below.

P.S. I just noticed a typo in #SASOstdf and corrected it directly in "dev" branch, then merged dev into this branch.

P5/Source/Guidelines/en/HD-Header.xml
```diff
@@ -2519,7 +2519,11 @@ grouping makes it easy to avoid accidentally flagging non-TEI data as errors
 during validation of the file against a TEI schema. The
 <gi>xenoData</gi> element, which may appear in the TEI header after
 the <gi>fileDesc</gi> but before the optional <gi>revisionDesc</gi>,
-is provided for this purpose.
+is provided for this purpose. Where the use of <gi>xenoData</gi> is primarily
+to store linked data, contextual information, or stand-off annotations
```
sydb, line 2523: Perhaps
~~~xml
Where the use of <gi>xenoData</gi> is primarily
to store linked data, contextual information, or stand-off annotations 
that refer to the contents of the TEI file itself, the
<gi>xenoData</gi> element may be placed inside the <gi>standOff</gi> 
element instead.
~~~
?

jamescummings: Yes, the use of the modal verb makes that clearer, etc. Go ahead with your proposed wording.

sydb: Done at e1673c3cf.

sydb (approved):

Issue #2544, opened 2024-03-28
allow `<quote>` in `<bibl>`


cthomasdta:
Since no-one objected to [my mail to TEI-L from Fri, 22 Mar 2024](https://listserv.brown.edu/cgi-bin/wa?A2=TEI-L;b7b66dfc.2403), I open this issue, asking to allow `<quote>` in `<bibl>`.

Background, as explained in the mail, is
> while encoding Goethe's letters for https://goethe-biographica.de/ we are putting information about sth that was sent together with the letter in `//teiHeader/physDesc/accMat/bibl[@type="attachment"]`. For example:
>
> ```XML
> <physDesc>
>      <accMat>
>             <bibl type="attachment">S.s Bemerkungen über den <quote>Umfang alter und neuer Architektur.</quote></bibl>
>       </accMat>
> </physDesc>
> ```
>
> This is not valid, since `<quote>` is not allowed in `<bibl>`, cf. https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-bibl.html.
I can't see any reason why not (plus there is `<q>` which is  allowed in `<bibl>`).

One could even argue, that there is no reason why `<quote>` should be verboten in any structure, but for now, we only ask to allow `<quote>` in `<bibl>`.

jamescummings:
Hi @cthomasdta, 

I'm not disputing the need for `<quote>` inside `<bibl>` and I can imagine some scenarios where this happens. In this instance, I just want to check is 'Umfang alter und neuer Architektur' not the title of a publication? If so, wouldn't `<title>` be more appropriate? 

I guess one avenue of objection might be that `<bibl>` (used for bibliographic citations) should be used for the bibliographic object and whether the quoted material really is part of that. One could wonder if the text isn't in quotation marks of some sort, is it really the allusion one thinks? Or, is it really part of the title? And if so, `<title>` already allows `<quote>` inside it.  

However, I think some of those objections have fairly obvious holes. 

[...]

PR #2557, opened 2024-05-17
Add `<quote>` to model.biblPart per #2544

sydb:
Title says it all; see #2544 for discussion.
Since it was the European subgroup at last SVF2F meeting that OKed this, I think this needs approval from at least one NA reviewer.


joeytakeda (approved):
👍 Tests pass, html output looks good, schemas validate

Issue #2541, opened 2024-03-19
silly Schematron in simplePrint


sydb:
I am making this issue just because if I fix this in dev right now we will probably end up with a conflict with #2540, as it involves the same lines of code.

## tei_simplePrint.odd, "choiceSize"
This constraint specification was made redundant by #633, and should just be deleted. (But if it were to stay that `>` sequence operator (spelled `&gt;` in the file) really should be a standard `gt` operator.)

## tei_simplePrint.odd, "choiceContent"
The `@test` in this constraint specification has two clauses. Unless I am missing something, the first clause is entirely superfluous and should just be deleted (along with the following `and` and the parens around the 2nd clause, of course).

sydb:
I have no idea where I got the “633”, above, because this was done by @jamescummings in 69858ffd9c11c7cde2340b652eb930d06dc511bf ; see issue #1690.

The fixes needed that I describe above were committed in d16c7979061dc0083a4f118a3bef229e8c8708b6.

This ticket could be closed if we open another one to delete the entire "choiceSize" constraint, which is pointless.

PR #2558, opened 2024-05-17
Remove extraneous "choiceSize" constraint from simplePrint

sydb:
Title says it all. This PR fixes the only remaining problem noted on issue #2541. So once this is merged that ticket should be closed.

Note that this PR is pretty trivial. Personally I think if the tests & builds work anyone can review and same person could merge.


jamescummings (commented):
All good, so hit merge.

Issue #2497, opened 2023-10-23
abstract model violation — NOT!


sydb:
[The following problem noticed by WWP encoder extraordinaire Grace O’Mara.]

The following (it seems to me) is a perfectly reasonable encoding:
~~~xml
<lg type="poem" subtype="stanzaic">
  <head>Ode II.</head>
  <head type="sub">The Mermaid.</head>
  <epigraph>
    <quote source="b:IT00863">
      <p>When at laſt they retired to reſt, <persName>Ajut</persName> went down to the beach, where
      <lb/>finding a fiſhing-boat, ſhe entered it without heſitation, and, telling thoſe … </p>
      <p>The fate of theſe lovers gave occaſion to various fictions and conjectures.
      …
      <lb/>her lover in the deſerts of the ſea.</p>
    </quote>
    <bibl><title ref="b:IT00863">Rambler</title>, N<g ref="#sup-o"/> 187.</bibl>
  </epigraph>
  <l>Blow on, ye death-fraught whirlwinds! blow,</l>
  <l>Around the rocks, and rifted caves;</l>
~~~

But the "abstractModel-structure-p-in-l-or-lg" constraint fires on each of those `<p>` elements complaining that “Lines may not contain higher-level structural elements”. That happens because those `<p>` elements **are** descendants of `<lg>`, but are **not** a descendant of `<floatingText>`, a child of `<figure>`, nor a child of `<note>`.[1]

Possible solutions, in my preferred order:

1. Change the test so it is only checking that `<p>`s that are a descendant of `<l>`, not of `<lg>`.
3. Add another clause to the test to check that the `<p>` is also not a descendant of `<epigraph>`.
2. Add another clause to the test to check that the `<p>` is also not a child of `<quote>`.
3. Require the `<epigraph>` be inside a `<floatingText>`.
4. Tell Grace that finally, at the end of her stellar 5-year career encoding for the WWP, she has found something that TEI simply cannot represent properly.

I like (1) the best, because I am not sure _why_ we are testing `<lg>` here. While I have a gnawing suspicion that it is at least partly my fault that we _are_ testing `<lg>`, I am having trouble figuring out what element a paragraph could occur in which should be an abstract model violation when it occurs in an `<lg>` but not outside. That is, I am (once again) complaining that
~~~xml
        <lg>
          <head><app><rdg><p/></rdg></app></head>
          <l/>
        </lg>
~~~
is an abstract model violation, but
~~~xml
[...]

PR #2559, opened 2024-05-19
fix an abstract model violation msg

sydb:
This is intended to address #2497. As a side effect, it simplifies the Schematron, too.

* Updated Schematron as per issue
 * Added a test to detest.xml
 * Updated expected results to match new test file


torstenroeder (commented):
The changes look right to me. What about the checks? Do they need to pass before approval?

ebeshero:
@torstenroeder The tests need to pass. The branch seriously needs to be updated from dev, and then we can test it again.

sydb:
Branch updated. Passed tests on my local system (not in docker, which I do not have running properly on this laptop).

ebeshero (approved):
I've now tested this thoroughly, including building the tei_all and testing it on an epigraph. All good.

Issue #2560, opened 2024-05-19
examples of `<geo>` with comma


peterstadler:
There are several examples of <geo> featuring a comma to separate values, see https://tei-c.org/release/doc/tei-p5-doc/en/html/examples-geo.html.
Yet, without any `@decls` attribute, "the assumption is that the content of each geo element will be a pair of numbers separated by whitespace, to be interpreted as latitude followed by longitude according to the World Geodetic System." (https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-geo.html)

To make these examples consistent and more clear to readers I propose to remove any comma within the content of `<geo>`.

An additional editorial task might be to check whether latitude and longitude are in the right order.

PR #2561, opened 2024-05-19
Remove extraneous comma from `<geo>`

sydb:
Removed extraneous commas from within examples of `<geo>` element, thus addressing main concern of #2560.

For each `<geo>` that had a comma, I checked to make sure it seemed to be in LAT LONG order (instead of LONG LAT). I did _not_ check those `<geo>`s that did _not_ have a comma.

**Note**
The following command was very useful in checking LAT LONG vs LONG LAT:
~~~bash
xmlstarlet sel -N x=http://www.tei-c.org/ns/Examples -t -m "//x:geo[ contains(.,'0')]" -n -o "---------" -f -o ":" -n -v "normalize-space(.)" -n Source/Specs/*.xml Source/Guidelines/en/*.xml | perl -pe 's;^([0-9.-]+),? ([0-9.-]+)$;normal= https://www.openstreetmap.org/?mlat=$1&mlon=$2&zoom=9\nreverse= https://www.openstreetmap.org/?mlat=$2&mlon=$1&zoom=9;'
~~~~

### DIFF
--- a/P5/Source/Guidelines/en/ND-NamesDates.xml
+++ b/P5/Source/Guidelines/en/ND-NamesDates.xml
@@ -879,7 +879,7 @@ target="#CONA"/>).  The tag -->The
       <egXML xmlns="http://www.tei-c.org/ns/Examples"><place xml:id="Rome">
                <placeName>Rome</placeName>
                <location>
-                  <geo>41.891775, 12.486137</geo>
+                  <geo>41.891775 12.486137</geo>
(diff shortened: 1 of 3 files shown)


peterstadler (approved):
Wow, that was super fast! Looks good to me :)

trishaoconnor (approved):
This looks good to me. I approve merging these changes to dev.

PR #2562, opened 2024-06-05
unify attribute sequence in specs

rettinghaus:
Most of the elements in the specs list their attributes alphabetically: `@versionDate` `@xml:lang`. 
This PR unifies all instances into that structure for better readability.


sydb (approved):
In one sense kinda pointless, as over time elements with attributes in different order will be introduced. But this PR does no harm, and at least temporarily makes these files a bit easier to read. So I am in favor. All tests pass, so as far as I am concerned this can just be merged.

rettinghaus:
rebased to fix conflicts

PR #2563, opened 2024-06-05
ci: update used actions

rettinghaus:


peterstadler (approved):
Thanks for the PR. Very much appreciated!

PR #2564, opened 2024-06-05
Add German gloss

rettinghaus:
Adding a German `gloss` to accompany the already existing German `desc`.


peterstadler (approved):
Looks good to me (as a German native speaker)

PR #2566, opened 2024-06-24
HTML and CSS to fit the look-and-feel of the new TEI site

hcayless:


ebeshero (commented):
Seems like the main visible change is to move navigation to the footer bar...which looks good with the new TEI logo! Home link seems to stay with index of the Guidelines--is that intended?

Sec32fun32 (approved):

PR #2568, opened 2024-07-04
Encode quotation marks

rettinghaus:
This PR properly encodes quoted words in descriptions.


ebeshero:
Thanks, @rettinghaus! This is simple and timely.

PR #2569, opened 2024-07-04
remove whitespace at end of descriptions

rettinghaus:
This PR removes some unneeded line breaks and superfluous whitespace ate the end of descriptions.


sydb (approved):
This PR basically nukes the extraneous whitespace before end-`<desc>` tags. While it misses some edge cases (like when there is a comment after the extraneous whitespace but before the end tag), and does not do other useful cleanup, it correctly does not mess with the end tag when the start tag was on a line of its own.

Thus the end result is better than what we have now. It passes all tests on my local system. There is no change to programming code or schema. Thus I think this PR should be merged immediately.

PR #2573, opened 2024-07-09
section of CMC `<post>` tagdoc mis-encoded

### LINKED ISSUE #2571: CMC attr description contains what should be remarks
The `<post>` element’s `@replyTo` attribute is defined as
~~~xml
<attDef ident="replyTo" mode="add">
    <desc xml:lang="en" versionDate="2021-09-07">indicates to which previous post the
        current post replies or refers to. This attribute should be used to encode
        <q>technical</q> reply information that is due to a formal reply action (such as
        activating a <q>reply</q> button in the client software) and that is formally
        represented in the source, e.g. in the <q>references</q> field of a Usenet message header
        or in the subject line of a forum post. It should not be used for inferred,
        or interpreted reply relations based only on ambiguous signs such as linguistic discourse markers or the indentation relations in Wiki talk pages.
        <!--See Lüngen/Herzberg (2019)--></desc>
    <datatype minOccurs="1" maxOccurs="unbounded">
        <dataRef key="teidata.pointer"/>
    </datatype>
</attDef>
~~~
That should (IMO) be
~~~xml
<attDef ident="replyTo" mode="add">
  <desc xml:lang="en" versionDate="2021-09-07">indicates to which previous post the current post replies or refers.</desc>
  <datatype minOccurs="1" maxOccurs="unbounded">
    <dataRef key="teidata.pointer"/>
  </datatype>
  <remarks xml:lang="en" versionDate="2021-09-07">
    This attribute should be used to encode <q>technical</q> reply
    information that is due to a formal reply action (such as
    activating a <q>reply</q> button in the client software) and that
    is formally represented in the source, e.g. in the
    <q>references</q> field of a Usenet message header or in the
    subject line of a forum post. It should not be used for inferred
    or interpreted reply relations based only on ambiguous signs such
    as linguistic discourse markers or indentation.
    <!--See Lüngen/Herzberg (2019)-->
  </remarks>
</attDef>
~~~
Although an argument could be made that the `@versionDate` should be updated for both the `<desc>` and `<remarks>`.

sydb:
This addresses #2571.

The only things changed are `<desc>`, `<remarks>`, and whitespace. Thus I think approval of 1 reviewer is sufficient.


ebeshero:
Adding @joeytakeda to the reviewer list since he and I were concerned about the definition of `@replyTo` in reviewing the CMC chapter...

P5/Source/Specs/post.xml
```diff
@@ -6,256 +6,258 @@ See the file COPYING.txt for details.
...
+        which is indicated by the elements contained within the <gi>post</gi>.</p>
+      </remarks>
+    </attDef>
+    <attDef ident="replyTo" mode="add">
+      <desc xml:lang="en" versionDate="2024-07-07">indicates to which previous post the current post replies or refers.</desc>
+      <datatype minOccurs="1" maxOccurs="unbounded">
+        <dataRef key="teidata.pointer"/>
+      </datatype>
+      <remarks xml:lang="en" versionDate="2024-07-07">
+	<p>This attribute should be used to encode <q>technical</q>
```
ebeshero, line 64: Noting that a portion is removed here. The description of `@replyTo` used to begin thus: "indicates to which post the current post replies or refers to". I suppose by removing it, we are moving some specific constraint on how people could apply this attribute, since it could now contain any "`<q>technical</a>` reply information ", and I'm not quite sure what that could mean. (Does it mean we indicate a button was pushed?) Before it seemed clearer that this was meant to be a pointer to a previous post to which this is a reply.  I think maybe we should put the indicator phrasing back. @sydb

ebeshero (changes requested):

P5/Source/Specs/post.xml
ebeshero: I do like the new phrasing of the next sentence, though re "<q>technical</q> reply information". But I don't think it stands on its own as a leading sentence in the description of `@replyTo`.

sydb: No, nothing (substantive) has been removed. Yes, the **description** used to begin “indicates to which post the current post replies or refers _to_” and now it **is** “indicates to which post the current post replies or refers” (the only difference is I dropped the word “to” from the end).
All the other stuff has been _moved_ (not _removed_) from the **description** to the **remarks**, where it belongs. But it also has not been changed substantively. It used to read “This attribute should … discourse markers or the indentation relations in Wiki talk pages.” and now instead reads “This attribute should … discourse markers or indentation.”.

peterstadler (approved):
This looks good to me, i.e. the PR fixes what the ticket complains about

P5/Source/Specs/post.xml
peterstadler: Agree with @syd here that it's just a moving of text parts without substantial changes

peterstadler:
This could go into the 4.8.1 ? Should the milestone for the ticket and PR be adjusted accordingly?

P5/Source/Specs/post.xml
ebeshero: Okay, no need to change it back, then. Sorry for missing the move to remarks.

ebeshero (approved):

PR #2576, opened 2024-07-16
compactify specGrps

rettinghaus:
There were tons of whitespace within `specGrp` elements in the Guidelines. I don't know where this came from, but it makes working with these files a nightmare. I removed the superfluous whitespace between the includes there to improve readability.

### DIFF
--- a/P5/Source/Guidelines/en/AB-About.xml
+++ b/P5/Source/Guidelines/en/AB-About.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- © TEI Consortium. Dual-licensed under CC-by and BSD2 licenses; see the file COPYING.txt for details. -->
-
-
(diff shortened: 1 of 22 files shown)


sydb (approved):
A quick glance says this change is only of (insignificant) whitespace, and thus should not effect the output _at all_.
All tests pass on my local machine. (All tests failed in my docker, but since it died on `make clean` I do not think that has _anything_ to do with this PR. If I do not solve that problem myself soon I will ask for help on Slack, probably.)
Thus I think this PR can be merged in with Council’s thanks without further ado. And the sooner the better, so we are less likely to have conflicts.
P.S. I believe, @rettinghaus, but am not sure, that a lot of that whitespace got inserted by an XSLT program that converted the previous instantiation of the TEI Guidelines source to use XInclude.

trishaoconnor:
Tests are passing and changes look good to me.

PR #2578, opened 2024-07-22
End descriptions with full stop

rettinghaus:
This PR unifies descriptions for models, elements, and attributes in order to end them with a full stop for English, French, and German (where I have sufficient knowledge to do so). 
I didn't touch the Italian descriptions (where it is even more inconsistent), or changed any wording. Found a small typo in a Spanish description and fixed it on the way as well.
The descriptions for attribute values weren't touched at all.


sydb:
Using the dev branch I checked the last character of the `<desc>` elements @rettinghaus is talking about (using the XPath `//(classSpec|elementSpec|attDef)/desc[lang('en') or lang('de') or lang('fr')]!normalize-space()!replace( ., '^.*(.)$', '$1')`), and found:
~~~
   2470 .
     98 e
     41 s
     40 t
     29 n
     24 d
     12 l
     11 y
     10 )
      8 I
      8 r
      6 a
      5 é
      4 j
      3 m
      3 u
      3 x
      2 i
      2 o
      1 6
      1 :
      1 b
      1 c
      1 h
      1 p
~~~
So it looks to me like 2,470 already end in dot, 303 should end in dot but don’t (this is the set of those that end in an alphabetic character), and 12 need looking into (those that end in close paren, digit, or colon).
I plan to take a look at those dozen weird ones in a few minutes. But in the meantime I think it is much better to have up to 12 false positives than 291 or more true negatives, so I am all in favor of merging this in _even if_ no one investigates the 12 weirdos.
That said, at the moment the automated build process is broken (because access to prior versions of P5 on the web is problematic) so in any case this should not be merged until the build process is working again and thus we can run tests properly.

sydb:
The dozen weird cases:
1. `att.global/@xml:lang`, "fr" = ends in ‘6’, deserves dot
2. `model.titlepagePart`, "fr" = ends in ‘)’, deserves dot (and needs space after comma, too)
3. `idno/@type`, "fr" = ends in ‘)’, deserves dot I think
4. `divGen/@type`, "fr" = ends in “etc.)”, so I am not sure what to do (in English, let alone French :-)
5. `docAuthor`, "de" = ends in ‘)’, deserves dot I think (note: "fr" needs space before dot deleted)
6. `mood`, "fr" = ends in ‘)’, deserves dot I think
7. `subc`, "en" = ends in “etc.)”, so I am not sure what to do
8. `subc`, "fr" = ends in “etc.)”, so I am not sure what to do
9. `lang`, "fr" = ends in ‘)’, deserves dot I think
10. `w`, "fr" = ends in ‘)’, deserves dot I think (note: "es" is missing a space and an ultimate dot)
11. `att.linguistic/@lemma`, "fr" = ends in ‘)’, deserves dot I think
12. `certainty/@type`, "en" = ends in ‘:’, probably needs to be entirely re-written

sydb:
We also need to decide (before merging) if the `@versionDate`s should be updated. My instinct is “no”, but I know very little about the translation process that relies on those attributes. I will ping (in Slack) a few folks who might know more.

rettinghaus:
Fixed some more overlooked cases and most of the ones pointed out by @sydb.

sydb:
1. The consensus was that the `@versionDate` should not be updated just for a final punctuation change. @martindholmes pointed out that (when "en" is updated) the corresponding translations should be checked to make sure they have the punctuation as well. That was only done for a subset of languages, here, but I do not think that should prevent us from accepting this PR.
2. I checked out rettinghaus:dots and checked the same XPath as previously. The new results are below. While it is mildly amusing that (sorted in frequency order) the first few after dot spell out “ends)”, it is also meaningless. The main point is that this branch is down to 64 of 2797 that do not end in dot (from 303 of 2773 — no, I do not understand what happened to 24 `<desc>`s), a _vast_ improvement, from 10.9% to 2.3%.

**below**
~~~
   2733 .
     16 e
      8 n
      6 d
      6 s
      4 )
      4 t
      3 l
      3 y
      3 é
      2 a
      2 h
      2 r
      1 :
      1 I
      1 c
      1 i
      1 o
~~~
So I am in favor of merging.

sydb (approved):
All that said, we (perhaps assignee) should ensure this passes the build before pushing a merge up. (That is, merge regginghaus:dots into TEIC:dev on a local, perhaps temporary, copy of the TEIC/TEI repo, and see if that copy builds. If it does, push push it up. If not, we would have to poke at the problem.)

PR #2579, opened 2024-07-26
fix description by restoring lost value

rettinghaus:
Small fix for Chinese description in `gap` where obviously one value got lost. Restored it from the similar Korean description.

### DIFF
--- a/P5/Source/Specs/gap.xml
+++ b/P5/Source/Specs/gap.xml
@@ -49,7 +49,7 @@
       <desc versionDate="2007-12-20" xml:lang="ko">누락의 이유를 제시한다. 그 값의 예들은 다음과 같다: <val>sampling</val>,
           <val>illegible</val>, <val>inaudible</val>, <val>irrelevant</val>, <val>cancelled</val>,
           <val>cancelled and illegible</val></desc>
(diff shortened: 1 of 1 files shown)


peterstadler (commented):
While this PR reintroduces a missing value (which was deleted in 05323e6), I think the more accurate way would be to remove the sample values from the `<desc>` completely as it is the case with other languages like English an Spanish. Here, the terms of the sample values are given in translation in the `<gloss>` to the `<valItem>`s: https://github.com/TEIC/TEI/blob/2e33676b16f5e25528ab386b74b0a0067a839ef1/P5/Source/Specs/gap.xml#L63-L110

rettinghaus:
Then this should be done by someone who actually speaks these languages.

sydb:
Note that _this_ PR is (as far as @trishaoconnor and I are concerned) is a GO for merge, but @peterstadler is right, we should find people who speak Korean and Taiwanese to correct the descriptions by removing the enumeration of values and whatever bit before them would no longer belong.

sydb:
Merging in this fix, and opening new ticket with @peterstadler’s (correct, IMHO) suggestion.

sydb:
See #2663.

PR #2581, opened 2024-08-05
Unify reason attribute

### LINKED ISSUE #2580: Harmonize reason attributes
`@reason` is available for the elements `gap`, `secl`, `supplied`, `surplus`, and `unclear`. 
Mostly they are based on `teidata.word`, however the first and the last one take the long way using `teidata.enumerated` instead. 
Is there a reason for this? I do not see any, so maybe it would be good to have all `@reason` attributes modeled in the same way.

rettinghaus:
This PR unifies all occurrences of the `reason` attribute, to use `teidata.word`.
Closes #2580

### DIFF
--- a/P5/Source/Specs/gap.xml
+++ b/P5/Source/Specs/gap.xml
@@ -58,7 +58,7 @@
       <desc versionDate="2007-01-21" xml:lang="it">indica il motivo dell'omisione</desc>
       <desc versionDate="2017-02-07" xml:lang="de">gibt den Grund für die Auslassung an.</desc>
       <datatype maxOccurs="unbounded">
-	<dataRef key="teidata.enumerated"/>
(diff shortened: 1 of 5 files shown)


sydb (changes requested):
While the changes here are perfect if Council decides to go with teidata.word for all 5, I am against that idea. (I would prefer to see teidata.enumerated with a vocabulary list for all 5, and maybe see `@reason` factored out into a class of its own.)
See [comments](https://github.com/TEIC/TEI/issues/2580#issuecomment-2331986499) on issue itself.

rettinghaus:
Changed the datatype in question to `teidata.enumerated`. I'm happy to move the attribute into a class of its own, if there is a definite decision on that.

sydb (approved):

Issue #2337, opened 2022-09-12
Need to standardize punctuation marks in Japanese translations


martindholmes:
There's a discussion on https://github.com/TEIC/TEI/pull/2294 about whether Japanese punctuation marks ( 、U+3001 for comma, 。U+3002 for period) or western period and comma should be used in the Japanese text of the Guidelines. At the moment it's a mixture. Whichever we settle on, we just need to standardize throughout the Guidelines Japanese text. This is an easy fix, which I'm happy to work away at once the decision is made.

knagasaki:
In the recent statement of the Japanese government below,  the "、U+3001 for comma" and 。"U+3002 for period" are recommended even in horizontal text.  So, if we don't need to mind consistency with the previous ones and it would be easy to change the existing marks into each one, it would be better to be done so. 
https://www.bunka.go.jp/seisaku/bunkashingikai/kokugo/hokoku/pdf/93651301_01.pdf

martindholmes:
Thanks @knagasaki ! I can make the changes , and we could even add a bit of Schematron to the Guidelines schema when all the conversions are done.

martindholmes:
Mixed with the regular ascii punctuation marks, we also have these:

U+FF0C FULLWIDTH COMMA
U+FF0E FULLWIDTH FULL STOP

I'll normalize these to U+3001 and U+3002 too.

martindholmes:
As of today, I've done all the normalizations I think are required. I'm going to keep this ticket open for the moment, because the Japanese translations are being updated steadily, and new instances are being introduced. This is something that can't easily be done with Schematron because there re contexts within the Japanese text where ascii commas and periods are actually fine (such as within numerals, or when discussing punctuation marks), so a quick pass before every new release is probably a good idea. I'll check again when the 4.7.0 freeze is in place.

martindholmes:
Found a couple more and fixed them in commit cd39bc5d4. I'm now going to set this to the next milestone.

martindholmes:
Pushing this to 4.10 since we're good for 4.9.

martindholmes:
Just flagging up that I did a trawl through the source today and fixed any instances found.

PR #2582, opened 2024-08-08
Added and revised Japanese translation

knagasaki:


ebeshero:
@martindholmes Would you take a look at this to review the translations? We also need to check with the conflict on the branch.

martindholmes:
@ebeshero I'm not really competent to review the translations as such, of course, but I do look at the spreadsheet where the work is happening sometimes to see if there's anything doesn't seem right to me. I think the att.pointing conflict arises from the other work to add final periods to desc elements, affecting the French translation which comes right after the Japanese; the other one seems to be a whitespace change, possibly from the same work.

I do want to make a point of running another check for issue #2337 before the next release, so once this is merged, I'll do that.

sydb:
Looks like all this needs is conflict resolution.

sydb:
I did the merge, but (perhaps because I messed up), it was committed to a branch on called knagasaki-knagasaki-patch-1 on this repo. Thus see PR #2701.

PR #2584, opened 2024-08-10
allow `<xenoData>` to have more than 1 child

### LINKED ISSUE #2583: content model of `<xenoData>` permits only one child
In tei_all.rnc from both the currently released TEI 4.8.0, and the 4.9.0a dev branch I built this morning, the `<xenoData>` element allows only one child element. This strikes me as a severe error, since it contradicts the _Guidelines_ which explicitly say “The xenoData element … may contain one or more elements from outside the TEI or data in some non-XML text format.”. 

The current content model for `<xenoData>` is
~~~xml
  <content>
    <alternate>
      <textNode/>
      <anyElement/>
    </alternate>
  </content>
~~~
I think that the `<anyElement>` should bear `maxOccurs="unbounded"` and either `minOccurs="0"` or `minOccurs="1"`.

That said, the `<xenoData>` element itself is repeatable, so someone may want to make an argument that the schema is correct, the _Guidelines_ should say it only allows 1 child — if you need both MODS records and Dublin Core metadata, use two `<xenoData>` elements. I do not think this is a reasonable argument (although in the MODS & DC case it would be good practice), because I may have a small pile of metadata elements that have nothing to do with some well defined metadata standard and that do not have a convenient wrapper.

So, personally, I think the only issue is whether the `<anyElement>` should have a minimum occurence of 0 or 1. But since the _other_ branch of its parent `<alternate>` allows the empty string to be the content of the `<xenoData>`, it does not make a whit of practical difference to the end user. Since those playing with schemas might prefer a 1 (since 0 `<anyElement>`s is the same as a zero length string, you might end up with ambiguous content if you use 0), I am in favor of a `@minOccurs` of 1.

sydb:
Fixes #2583 by adding `@minOccurs` and `@maxOccurs` to `<anyElement>` in content model of `<xenoData>`.

### DIFF
--- a/P5/Source/Specs/xenoData.xml
+++ b/P5/Source/Specs/xenoData.xml
@@ -16,7 +16,7 @@
   <content>
     <alternate>
       <textNode/>
-      <anyElement/>
+      <anyElement minOccurs="1" maxOccurs="unbounded"/>
     </alternate>
   </content>
   <exemplum xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" versionDate="2017-05-11" xml:lang="en">


martindholmes:
This looks straightforward to me. Should be merged.

ebeshero:
This passes tests and builds, and appears to be eminently sensible. I'm merging it as suggested.

PR #2589, opened 2024-09-09
repairing some errors in TD

sydb:
* more truthful discussion of `<sch:ns>`
 * fixed incorrect `@context` in an example
 * added missing Oxford comma
 * removed no longer valid reference to "isoschematron" as value for `@scheme` attribute
 * reversed logic: i.e., changed an `<sch:report test="not(TST)">` to an `<sch:assert test="TST">`.

Notes:

1. I did not generate an issue, just this PR
2. I do not think this needs 4 reviewers, I think 1 will do
3. I made this high priority because even though it is just a prose change a) what is there now is (in some cases) just plain wrong; and b) I would like to have this fixed in dev branch before I start writing the prose for `<constraintDecl>`, which will change some of the same portions.


P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -943,12 +947,13 @@ to mark any technical term, thus:
...
-        may require somewhat more effort when creating a customization
-        file. Such private schemes will generally be even more
-        problematic on implementation of the constraints themselves,
-        as it may require siginficant programming work. The TEI only
-        provides this capability for the suggested values.</p>
+        (currently only <val>schematron</val>) may require somewhat
+        more effort when creating a customization file. Such private
+        schemes will generally be even more problematic on
+        implementation of the constraints themselves, as it may
+        require siginficant programming work. The TEI only provides
```
HelenaSabel:
```suggestion
        require significant programming work. The TEI only provides
```

HelenaSabel (approved):
Tests fail after my commit when retrieving a P5 subset but this error is not related to the changes done (which I agree with @sydb should be merged relatively soon)

P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -858,35 +859,39 @@ to mark any technical term, thus:
...
       </p>
-      <p>The constraints in the preceding example all related to attributes in the empty namespace, and the
-        Schematron rules did not therefore need to define a TEI namespace prefix. The Schematron language <gi
-          scheme="Schematron">ns</gi> element should be used to do this when a constraint needs to refer to a
-        TEI element, as in the following example, which models the constraint that a TEI <gi>div</gi> must
-        contain either no subdivisions or at least two of them:
+      <p>Note that the <q>tei:</q> prefix needs to be bound to the TEI
+      namespace using the Schematron language <gi
+      scheme="Schematron">ns</gi> element, but typically this only
+      need be done once. The following two examples are written
```
sabineseifert: Shouldn't it be "this only needs be done once" - with 3rd person singular S on "need" or is it some special grammar rule I am not aware of?

P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -858,35 +859,39 @@ to mark any technical term, thus:
...
               <sch:rule context="tei:div">
                 <sch:report test="count( tei:div ) eq 1">if it contains any subdivisions, a
                 division must contain at least two of them</sch:report>
               </sch:rule>
             </constraint>
           </constraintSpec>
         </egXML>
-         Schematron rules are also useful where an application needs to enforce rules on attribute values, as
-        in the following examples which check that various types of <gi>title</gi> are provided:
+      The second demonstrates that Schematron rules are also useful
```
sabineseifert: I'd prefer: "The second example..."

sabineseifert (commented):
I just left 2 comments. It looks good to me but I must confess that I feel insecure to evaluate these changes..

P5/Source/Guidelines/en/TD-DocumentationElements.xml
sydb: Oooh, good question. I don’t think so. I think — but am not at all sure — that “need be done” is sort of an idiomatic way of saying either “needs to be done” or “needs to have been done”. In any case, I do not think I have ever heard or seen “needs be done”.

But in any case, perhaps the best thing to do is to avoid this usage altogether and write “… typically this only needs to be done once”.

P5/Source/Guidelines/en/TD-DocumentationElements.xml
sydb: Sounds reasonable to me.

P5/Source/Guidelines/en/TD-DocumentationElements.xml
trishaoconnor: This is a good point. I read it first thinking it was fine but after you've pointed this out, I'm not certain. I concur with writing "... typically this only needs to be done once".

P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -858,35 +859,39 @@ to mark any technical term, thus:
...
-        Schematron rules did not therefore need to define a TEI namespace prefix. The Schematron language <gi
-          scheme="Schematron">ns</gi> element should be used to do this when a constraint needs to refer to a
-        TEI element, as in the following example, which models the constraint that a TEI <gi>div</gi> must
-        contain either no subdivisions or at least two of them:
+      <p>Note that the <q>tei:</q> prefix needs to be bound to the TEI
+      namespace using the Schematron language <gi
+      scheme="Schematron">ns</gi> element, but typically this only
+      need be done once. The following two examples are written
+      presuming that the namespace binding for <q>tei:</q> has occured
+      elsewhere. The first models the constraint that a TEI
```
trishaoconnor: Maybe have it as "The first example models ..." for consistency, since we are adding the word "example" for the second example.

P5/Source/Guidelines/en/TD-DocumentationElements.xml
```diff
@@ -858,35 +859,39 @@ to mark any technical term, thus:
...
                 division must contain at least two of them</sch:report>
               </sch:rule>
             </constraint>
           </constraintSpec>
         </egXML>
-         Schematron rules are also useful where an application needs to enforce rules on attribute values, as
-        in the following examples which check that various types of <gi>title</gi> are provided:
+      The second demonstrates that Schematron rules are also useful
+      where an application needs to enforce rules on attribute values,
+      here checking that various types of <gi>title</gi> are provided
```
trishaoconnor: Maybe instead of "here" rewrite as "for instance," ?

trishaoconnor (commented):
I've added some very small comments that are very quick changes. 
It all looks good to me. It does not seem to be passing the tests on GitHub but since these are prose changes, I can't see why these shouldn't be merged into dev.

P5/Source/Guidelines/en/TD-DocumentationElements.xml
ebeshero: I'm correcting this now (while testing locally) to "this only needs to be done once."

P5/Source/Guidelines/en/TD-DocumentationElements.xml
ebeshero: Just added that "The second example..."

P5/Source/Guidelines/en/TD-DocumentationElements.xml
ebeshero: Done...

P5/Source/Guidelines/en/TD-DocumentationElements.xml
ebeshero: I'm modifying "here" to make it more specific as @trishaoconnor suggests, but also to train it on this first example, so the full sentence reads:
"The second example demonstrates that Schematron rules are also useful
      where an application needs to enforce rules on attribute values,
      in this case checking that various types of <gi>title</gi> are provided
      in the TEI header:"

ebeshero:
What I'm doing as an experiment is reverting @HelenaSabel 's push and then redoing her change at that point, and seeing if the build is still broken. I've also gone ahead and collected everyone's edits so far (thanks @sabineseifert and @trishaoconnor ). This is just a little experiment to see if the build might mend itself on my next push.... Fingers crossed this works! 🤞

ebeshero:
We have repaired the testing issues! It was all about repairing the `@calendar` deprecation warnings, and @sydb fixed it all--huzzah!!! 🥇

PR #2593, opened 2024-09-19
remove deprecation, it is now an error

sydb:
Went through all ~64 spec files that had their own declaration for `@calendar` (so that it could have a deprecation warning) and removed them.


ebeshero:
Delighted to merge this much-needed PR! Thank you @sydb !

Issue #2330, opened 2022-08-10
schemaSpec should provide a mechanism for specifing Schematron query language binding 


martindholmes:
Schematron schemas can specify a range of different values for the query language binding (typically xslt, xslt2, or xslt3, but many more are listed here: https://archive.xmlprague.cz/2022/files/presentations/schematron-qlb.pdf). The ATOP team believes that TEI ODD should have a mechanism for specifying this, perhaps through an attribute on the root schemaSpec element called schQueryLanguageBinding.

martindholmes:
The ATOP team is thinking of two distinct approaches to this:

1. (Simple): There is a single `@schQueryLanguageBinding` attribute available on `<schemaSpec>`, meaning that all Schematron rules must be compiled into a single Schematron output file and use the same query language binding.
2. (Adventurous): We create a `<constraintDecl>` element in the header, where various things can be defined and described, including the query language binding value, as well as namespace declarations etc. `<constraintSpec>` would then be added to `att.declaring`, and each `<constraintSpec>` could point up to a `<constraintDecl>`. For each distinct `<constraintDecl>` element, a distinct Schematron file would be created using the specified content, including the query language binding, and including all the constraints which point to it.

The second option would obviously provide much more flexibility, but would require non-trivial fixes to the existing stylesheets, and there may not be many real use-cases for it; generally, we would expect people to want to use the most advanced query language binding that their processor can support (why not?), and there's no particular reason to want to use an earlier language binding. There's also the question then of how you might override the binding for existing constraints in a downstream customization.

ebeshero:
I like the adventurous option in theory, but I’m struggling to imagine a practical use-case for generating multiple different Schematron files. I also wonder whether we run a risk of over-complicating schema validation: could multiple Schematron files potentially conflict with each other’s validation when Schematron rules from two different files are triggered in an overlapping context? I guess I wonder if the adventurous path leads to too much potential trouble.

lb42:
If you think of the generated schematron file as an output from the odd rather than as input to the validator, surely it makes sense to maximize flexibility for its format?

jamescummings:
The separation of concerns provided by the second option seems more in keeping with the TEI approach to such things.

sydb:
Also note that (per [#335](https://github.com/TEIC/Stylesheets/issues/335)) we should be documenting, somewhere, that one does not get a query binding if using Schematron embedded in RELAX NG.

sydb:
[...]

PR #2594, opened 2024-09-20
add `<constraintDecl>`

sydb:
Added `<constraintDecl>` element.


P5/Source/Specs/constraintDecl.xml
```diff
@@ -0,0 +1,96 @@
...
+    <memberOf key="model.encodingDescPart"/>
+    <!-- <constraintDecl> is part of content of <schemaSpec> directly, not via a class -->
+  </classes>
+  <content>
+    <sequence>
+      <alternate minOccurs="0" maxOccurs="unbounded">
+        <classRef key="model.identEquiv"/>
+        <classRef key="model.descLike"/>
+      </alternate>
+      <anyElement/>             <!-- typically <sch:ns> elements -->
```
raffazizzi: I would suggest making this optional, in case one only wanted to use `<constraintDecl/>` to declare a query language binding and nothing else.

P5/Source/Specs/constraintDecl.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- © TEI Consortium. Dual-licensed under CC-by and BSD2 licenses; see the file COPYING.txt for details. -->
+<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
+<elementSpec xmlns="http://www.tei-c.org/ns/1.0" xmlns:sch="http://purl.oclc.org/dsdl/schematron" module="tagdocs" ident="constraintDecl">
+  <gloss versionDate="2023-03-09" xml:lang="en">constraint declaration</gloss>
+  <desc versionDate="2023-03-09" xml:lang="en">contains declarations pertaining to formal constraints expressed elsewhere in <gi>constraintSpec</gi> elements</desc>
+  <classes>
+    <memberOf key="att.global"/>
+    <memberOf key="model.encodingDescPart"/>
```
raffazizzi, line 9: This won't do if we want to make this element only be used once, as https://github.com/TEIC/TEI/issues/2330 required.

P5/Source/Specs/encodingDesc.xml
```diff
@@ -15,44 +15,38 @@ source or sources from which it was derived.</desc>
...
-        <alternate minOccurs="1" maxOccurs="unbounded">
-          <classRef key="model.encodingDescPart"/>
-          <classRef key="model.pLike"/>
-        </alternate>
-      
-    
+    <alternate minOccurs="1" maxOccurs="unbounded">
+      <classRef key="model.encodingDescPart"/>
+      <classRef key="model.pLike"/>
+    </alternate>
```
raffazizzi, line 29: I would suggest interleaving a reference to an optional constraintDecl here.

```xml
<sequence preserveOrder="false">
      <alternate minOccurs="1" maxOccurs="unbounded">
       <classRef key="model.encodingDescPart"/>
        <classRef key="model.pLike"/>
      </alternate>
      <elementRef key="constraintDecl" minOccurs="0" maxOccurs="1"/>
    </sequence>
```

raffazizzi (changes requested):
Suggested different content models for contraintDecl and encodingDesc

P5/Source/Specs/constraintDecl.xml
sydb: I think you are quite right.
Good catch.
Hope to make change today.

P5/Source/Specs/constraintDecl.xml
sydb: We only want 1 `<constraintDecl>` per scheme (rather than allowing multiples and using decarable/decaring mechanism), but we still have to allow multiples in case there is more than one scheme. E.g.
~~~xml
<constraintDecl scheme="schematron" queryBinding="xslt3"/>
<constraintDecl scheme="prolog">
  $globalVarOne=FALSE;
  $globalVarTwo=17;
</constraintDecl>
~~~
(I have no idea if that is syntactically correct PROLOG; I kinda doubt it. But it does not matter for our purposes.)
That said, it might be a good idea to have a `<constraintSpec>` for `<constraintDecl>` that constrained use to only 1 per scheme. :smile:

P5/Source/Specs/constraintDecl.xml
sydb: Not just, BTW, in case user wants **no** `<sch:ns>`, but it would be appropriate to let poor user have 2 or more of them!

P5/Source/Specs/encodingDesc.xml
raffazizzi: I think this is no longer relevant if we're making constraintDecl a member of model.encodingDescPart

raffazizzi (approved):

ebeshero:
I resolved the merge conflict over TD (to reconcile the new material about constraintDecl with another update earlier in the chapter--may want to take a look at TD, but I think this is what we want.

ebeshero (changes requested):
Everything works here: all tests are passing, the Guidelines build and validate. Also,`<constraintDecl>` is functioning in a tester ODD I created from the CBML ODD with a new Schematron rule and an application of `<constraintDecl>`. 

There's one thing that's bothering me about the expected behavior of `<constraintDecl>` as exemplified in its element spec. 

We have this in the egXML:
```
   <constraintDecl scheme="schematron" queryBinding="xslt3">
        <sch:ns prefix="wwp" uri="http://www.wwp.northeastern.edu/ns/textbase"/>
      </constraintDecl>
```

But that does not work in my ODD to RNG transformation. The output RNG doesn't show an error, but when applied to a validate an XML file, validation fails due to an error over the unrecognized namespace (the one declared as shown in the egXML).

To make this work, I had to modify my tester ODD to set the `<sch:ns/>` element (with its `@prefix` and `@uri` inside the `<constraint>` of my `<constraintSpec>`--it had to be set in the immediate content of a schematron rule. I think this wasn't the intention /expected behavior of `constraintDecl` was it? 

I'm attaching my tester files, built on the CBML ODD with a simple Schematron test on cbml elements.
[out-PR2594.zip](https://github.com/user-attachments/files/17565642/out-PR2594.zip)

P5/Source/Specs/constraintDecl.xml
```diff
@@ -0,0 +1,96 @@
...
+        defined by the specification. Most processors only support one
+        or more of <val>xslt</val>, <val>xslt2</val>, and
+        <val>xslt3</val>.</p>
+      </remarks>
+    </attDef>
+  </attList>
+  <exemplum xml:lang="en">
+    <egXML xmlns="http://www.tei-c.org/ns/Examples">
+      <constraintDecl scheme="schematron" queryBinding="xslt3">
+        <sch:ns prefix="wwp" uri="http://www.wwp.northeastern.edu/ns/textbase"/>
```
ebeshero, line 89: This code does not function to deliver a specialized namespace to the constraintSpecs that evoke it, but we don't see this until we attempt to validate XML with the output RNG. I had to declare the `sch:ns` later on within the `<constraintSpec>` (immediately within a `<constraint>` element) for it to be recognized by Schematron and function for validation.

sydb:
I cannot tell if @ebeshero means we should not merge this PR until [Stylesheets#697](https://github.com/TEIC/Stylesheets/issues/697) has been addressed, or she was just not aware that it has not been addressed. Thus this PR now NEEDS DISCUSSION to figure that out. Personally I would like to see this handled pretty quickly, as several major files are changed and I would just as soon avoid a merge nightmare.

ebeshero:
@sydb It was probably this: "she was just not aware that it has not been addressed". Shall we address https://github.com/TEIC/Stylesheets/issues/697 then?

sydb:
Suppose so. I am not sure why @trishaoconnor (& I) put the PENDING label on it. Seems to me we are good to go for someone to do it and generate PR, no? I am not very familiar with the **odds/teiodds.xsl** code, but since (IIRC) there is no way to assert a query language binding when Schematron occurs inside RELAX NG, this may be as simple as setting up a template that matches `<constraintDecl scheme="schematron">` and copies its Schematron children into the output.

PR #2596, opened 2024-10-01
Created constraintDecl

raffazizzi:
This is an initial implementation to address #2330. I would like some input on a couple of things.

* I made `<anyElement/>` optional in the content of `constraintDecl`. I think the element can be empty if all that's needed is specifying the query language binding.
* I allowed `constraintDecl` in `encodingDesc`  to interleave with other elements. OK?
* `schemaSpec` has a `<sequence>` content model, so I couldn't interleave `constraintDecl`. Instead, I placed if after elements that typically go first (`equiv`, `desc`), but before spec / ref elements. Thoughts?
* I would appreciate feedback on the text I added to Chap 22 and whether it's in the right place.

### DIFF
--- a/P5/Source/Specs/encodingDesc.xml
+++ b/P5/Source/Specs/encodingDesc.xml
@@ -23,14 +23,13 @@ source or sources from which it was derived.</desc>
     <memberOf key="model.teiHeaderPart"/>
   </classes>
   <content>
-    
-      
-        <alternate minOccurs="1" maxOccurs="unbounded">
-          <classRef key="model.encodingDescPart"/>
-          <classRef key="model.pLike"/>
-        </alternate>
-      
-    
+    <sequence preserveOrder="false">
+      <alternate minOccurs="1" maxOccurs="unbounded">
+        <classRef key="model.encodingDescPart"/>
+        <classRef key="model.pLike"/>
+      </alternate>
+      <elementRef key="constraintDecl" minOccurs="0" maxOccurs="1"/>
+    </sequence>
   </content>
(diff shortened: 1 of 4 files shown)


P5/Source/Specs/encodingDesc.xml
```diff
@@ -23,14 +23,13 @@ source or sources from which it was derived.</desc>
...
-    
-      
-        <alternate minOccurs="1" maxOccurs="unbounded">
-          <classRef key="model.encodingDescPart"/>
-          <classRef key="model.pLike"/>
-        </alternate>
-      
-    
+    <sequence preserveOrder="false">
+      <alternate minOccurs="1" maxOccurs="unbounded">
```
raffazizzi, line 27: Note that this seems to result in relaxng that does not represent the ODD's model, but it may be a separate issue.

Expected:
```xml
<interleave>
  <choice>
    <oneOrMore>
      <ref name="model.encodingDescPart"/>  
    </oneOrMore>
    <oneOrMore>
      <ref name="model.pLike"/>
    </oneOrMore>
  </choice>
  <optional>
    <ref name="constraintDecl"/>
  </optional>
</interleave>
```

Actual:
```xml
<interleave>
   <oneOrMore>
      <choice>
         <ref name="model.encodingDescPart"/>
         <ref name="model.pLike"/>
      </choice>
   </oneOrMore>
   <optional>
      <ref name="constraintDecl"/>
   </optional>
</interleave>
```

raffazizzi:
Closing as duplicate of https://github.com/TEIC/TEI/pull/2594

PR #2598, opened 2024-10-03
Emergency fix: return removed mode=add due to missing value lists

sydb:
Put mode="add" back on valList elements per #2597.

### What I did
- Looked up when change that removed the mode=add occurred. (It was 2024-01-26).
- Checked out dev branch as of the day before.
- Copied P5/Source/Specs/ directory to a temporary location A/.
- Checked out dev branch as of the day after.
- Compared all files in P5/Source/Specs/ tree to the corresponding file in A/.
- Using a little Emacs macro, repeatedly
  - searched the `diff` output from above for `<valList .*mode="a`
  - from that spot searched backwards for filename
  - opened that file
  - searched for `<valList`
- Checked that it was the right `<valList>`, and if not found right one.
- Inserted `mode="add"`

Did this to 45 files.
I have tested the result in that it builds on my local laptop, passing the tests.
I have not tested in that I have not looked to see if it worked, i.e. to see if it fixes the problem. Too tired right now. G’night.

### DIFF
--- a/P5/Source/Specs/abbr.xml
+++ b/P5/Source/Specs/abbr.xml
@@ -45,7 +45,7 @@
                 l'abbreviazione secondo una tipologia funzionale.</desc>
       <desc versionDate="2016-11-25" xml:lang="de">erlaubt es, die Abkürzung nach einer geeigneten Typologie zu klassifizieren.</desc>
       <datatype><dataRef key="teidata.enumerated"/></datatype>
-      <valList type="open">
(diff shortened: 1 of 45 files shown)


tuurma:
F2F: Stylesheets meeting on Oct 22 will try to tackle this, aiming to have a fix for tentatively planned TEI release mid-November

HelenaSabel:
Unnecessary now that the Stylesheets have been fixed

PR #2599, opened 2024-10-06
version number, take 2

sydb:
For the moment this branch is a partiallly working version of version sublcasses that is NOT ready for actual use, yet — in large part untested, and prose not written. This version is simply for Council to see if they like it, but (I think) it does address the problem for which original ticket was open.


sydb:
All tests pass on my local system (Padfoot, not in the Docker).

P5/Source/Guidelines/en/BIB-Bibliography.xml
martindholmes, line 1: The second biblStruct should be titled Calendar Versioning, not Semantic Versioning 2.0.0. Copy/paste error I think.

P5/Source/Specs/teidata.version.xml
martindholmes, line 1: I think this needs a new remarks element explaining that its content can be one of four different formats of version, and perhaps pointing to the definitions of the dataSpecs. It's going to be quite challenging to figure out which specific dataSpec you're not quite aligning with when you get error messages typing a version string.

P5/Source/Guidelines/en/BIB-Bibliography.xml
sydb: Hah! I just noticed that a little while ago and thought “I should fix that, as it has nothing to do with what I want Council paying attention to, but no one but @HelenaSabel is going to notice …”! :smile:

P5/Source/Specs/teidata.version.xml
sydb: Agreed, although I have no intention of putting in that effort until Council approves the general approach.

P5/Source/Guidelines/en/BIB-Bibliography.xml
sydb: Fixed in 6da041d47.

martindholmes, line 1: I'd never use `<biblStruct>`s for new bibliography entries. @sydb if you're OK with it, I volunteer to convert these to full `<bibl>`s.

P5/Source/Specs/TEI.xml
martindholmes, line 1: @sydb The `<desc>` for Italian has acquired a new `@versionDate` set to 2026, but the original content from 2007 remains unchanged. Misleading, surely? Especially since the English `@versionDate` is newer than 2007 but older than 2026. I think this must be an error. The Italian does not match the English desc, but it would not show up as needing an update since its `@versionDate` is later than that on the English `<desc>`. Also, should the date for English really be 2018?

P5/Source/Specs/teidata.version.Unicode.xml
martindholmes, line 1: @sydb The French `<desc>` is completely wrong here, so I think it should be fully updated before going further. I suggest:

définit la gamme des valeurs d'attribut exprimant un numéro de version Unicode.

P5/Source/Specs/teidata.version.xml
martindholmes: @sydb The French `<desc>` needs updating here too. Should be perhaps:

définit la gamme des valeurs d'attribut  exprimant un numéro de version TEI ou Unicode.

P5/Source/Guidelines/en/BIB-Bibliography.xml
```diff
@@ -3658,6 +3658,26 @@
...
+        <!-- This probably needs to be fleshed out more —Syd, 2024-05-21 -->
+        <monogr>
+          <title>Semantic Versioning 2.0.0</title>
+          <ptr target="https://semver.org/#semantic-versioning-200"/>
+          <imprint>
+            <date when="2023-12-19"/>
+          </imprint>
+        </monogr>
+      </biblStruct>
+      <biblStruct>
```
ebeshero: @sydb @martindholmes I don't think I care whether these are biblStruct or bibl, but we do need each entry to include an `@xml:id`.

P5/Source/Specs/att.styleDef.xml
```diff
@@ -45,23 +40,25 @@
...
     </attDef>
     <attDef ident="schemeVersion" usage="opt">
       <desc versionDate="2013-04-12" xml:lang="en">supplies a version number for the style language provided in <att>scheme</att>.</desc>
-      <datatype><dataRef key="teidata.versionNumber"/></datatype>
+      <datatype><dataRef key="teidata.version.Unicode"/></datatype>
+      <!--
+          All version numbers of XSLFO (1.0, & 1.1) and of CSS (1, 2,
+          2.1, & 3) are expressible using teidata.version.Unicode. If
+          other possible values are added to @scheme, the datatype
+          used here may need to be revised. —Syd, 2025-01-13
```
ebeshero: I wonder about this, @sydb ... In case of alternative style languages (what about Sass?), should we just permit teidata.version?

P5/Source/Guidelines/en/BIB-Bibliography.xml
ebeshero: If you're pushing some updates to this branch, @sydb or @martindholmes, would you add an xml:id for the Calver entry in BIB?

ebeshero (changes requested):
@sydb Small changes requested (see line comments).

P5/Source/Guidelines/en/BIB-Bibliography.xml
sydb: Go for it!

P5/Source/Specs/teidata.version.xml
sydb: > I think this needs a new remarks element explaining that its content can be one of four different formats of version, and perhaps pointing to the definitions of the dataSpecs.

OK, although that does mean if & when we add another teidata.version.WHATEVER we need to update the remarks to say “five” instead of “four”.

> It's going to be quite challenging to figure out which specific dataSpec you're not quite aligning with when you get error messages typing a version string.

True, but having better remarks is not really going to help.

It is worth noting that the **only** place we use teidata.version itself is for the `@version` of the `<application>` element. I think there is a reasonable argument to be made (that I thought you were going to make) that `application/@version` should be declared as teidata.text or (at most restrictive) teidata.word, as there is no predicting what the version numbering scheme for some random application is.

P5/Source/Specs/TEI.xml
sydb: Uhhh … I am not so sure. My reading of git blame (which I am not very tood at reading) is that the 2026 Italian starts with “la” and ends with dot, and the 2007 did not. My hope is that @Raff will just sweep in and fix it. :smile: 

As for the English description, yes, it was changed between 3.2.0:
> specifies the major version number of the TEI Guidelines against which this document is valid.

and 3.3.0:
> specifies the version number of the TEI Guidelines against which this document is valid.

P5/Source/Specs/teidata.version.Unicode.xml
sydb: I updated it to your suggestion.

Another problem is the `@versionDate` of the French remarks. (Which also need an `@ident`, but lots of things do. I will probably do that all at once later.)

P5/Source/Specs/att.styleDef.xml
sydb: I suppose so. "css" and 'xslfo" are the only two languages in our controlled vocabulary, but someone could use "other", of course. I do not really understand the versioning of Sass, but someone could use one of the have-to-be-pre-processed-into-CSS systems (like `less` or `stylus`) too, I guess. So it might make sense to use teidata.version, or per the 
[end of above comment](#discussion_r3575254061) teidata.word or teidata.text.

P5/Source/Guidelines/en/BIB-Bibliography.xml
sydb: Good catch, thanks; done.

P5/Source/Specs/att.styleDef.xml
martindholmes: I think we should only recognize languages maintained by international bodies such as the W3C. Sass in particular is increasingly irrelevant as CSS itself moves forward (variables, nesting, etc.).

ebeshero: @martindholmes I'm not really in agreement with this position about which languages we formally recognize by implication of datatype formats. But overall, the real question is whether `teidata.version` is simply sufficient here. Would you agree that it is?

ebeshero: For @sydb , I don't think after all the effort of introducing these datatypes that we really want `teidata.word` or `teidata.text` here! Just `teidata.version`, and that provides some nice options.

ebeshero (approved):
This looks good now! I pushed one last change for the CalVer date to show in BIB.

ebeshero (approved):
After the changes made during the F2F in Vancouver, I think this is ready to merge.

ebeshero:
@sydb @martinascholger @martindholmes I removed the "draft" status from this PR and approved it after the modifications we made today during the F2F.  I presume that's okay? I believe it is ready to merge to dev and should be part of the next release.

PR #2600, opened 2024-10-06
WIP: remove mobi from the generated outputs; close #2574

### LINKED ISSUE #2574: Do we need to generate Kindle mobi files?
The build process currently generates a special EPUB version and then processes it with the kindlegen tool to produce a .mobi file. However, Kindle now supports EPUB natively, and .mobi seems to be on the way out:

https://kdp.amazon.com/en_US/help/topic/G200634390

Since kindlegen itself is not open-source (one reason it's not installed on my build server), and .mobi isn't necessary, I suggest we completely remove this step from our build process, and offer only the PDF and EPUB formats for offline reading. 

The relevant Makefile lines are here:

https://github.com/TEIC/TEI/blob/dev/P5/Makefile#L124

tuurma:


martindholmes (commented):
@tuurma Does the Guidelines build process need additional changes to remove the icon from the top of the pages?

tuurma:
Thanks @martindholmes, I made a more thorough scan for mobi references, will need to test but we have a lunch break atm

martindholmes:
Found a few more bits and pieces that might need to be removed at Council FTF 2025-05-18.

PR #2610, opened 2024-10-30
add xml:lang to constraint specifications

sydb:
Added `@xml:lang="en"` to the 88 `<constraintSpec>`s that did not already have `@xml:lang`. (It turns out there were three that already had it, all of which are in "fr".)

### DIFF
--- a/P5/Exemplars/isofs.odd
+++ b/P5/Exemplars/isofs.odd
@@ -38,7 +38,7 @@
 	 <moduleRef key="tei"/>
 	 <moduleRef key="iso-fs"/>
 	 <elementSpec ident="fLib" mode="change">
-	   <constraintSpec ident="c1" scheme="schematron" mode="add">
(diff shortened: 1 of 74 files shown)


sydb:
Things I do not understand:

- Why is detest.odd validated against p5odds (instead of tei_odds, or _maybe_ tei_customization)?
- Same question for tei_enrich.odd and tei_simplePrint.odd?
- Or for that matter, tei_customization? For cryin’ out loud, I wrote that one, I know full well it is was not meant to be valid against p5odds!
- Why are the constraintSpec-sans-scheme-attr tests now being run twice?

tuurma (approved):
I posted a comment on Slack with asking for clarification on the long-term handling of different languages in constraints, but approving as I believe this is obviously a step in the right direction.

trishaoconnor:
Changes look good and passing all tests.

PR #2609, opened 2024-10-29
Fix build by passing $lang to schematron extraction

sydb:
The fix for Stylesheets [#702](https://github.com/TEIC/Stylesheets/issues/702) (i.e., [PR 704](https://github.com/TEIC/Stylesheets/pull/704)) caused the TEI build to break, because the natural language being processed was not being handed to extract-isosch.xsl, which it now needs to know. This fixes said problem. It passed all tests in Docker on my system.

PR #2611, opened 2024-10-31
heed `@role` of `<sch:assert>` and `<sch:report>`

sydb:
Turns out we were only testing for "nonfatal" vs everything else. However (since oXygen gives user the choice of "info", "information", "warn", "warning", "error", or "fatal") we have several cases of other values. Changed code to check for any of the union of the values we actually have and the values oXygen has in its “sample values include” list.

I did not set a milestone for this because it is _not_ for the 4.9.0 or some future release, it is for tomorrow’s 4.8.1 release.

### DIFF
--- a/P5/Utilities/validator.xsl
+++ b/P5/Utilities/validator.xsl
@@ -46,7 +46,13 @@
   <xsl:template match="svrl:failed-assert|svrl:successful-report">
     <xsl:variable name="outGI">
       <xsl:choose>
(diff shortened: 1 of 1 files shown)


tuurma (approved):

HelenaSabel (approved):
Note: tests fail here because the changes from PR https://github.com/TEIC/TEI/pull/2609 are not incorporated in this branch

PR #2614, opened 2024-11-11
Remove dead link

rettinghaus:
Same problem as with TEIC/Stylesheets/issues/703.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,5 @@ The [TEI](https://www.tei-c.org) is an international and interdisciplinary stand
 This repository contains the sources for building the [TEI P5 Guidelines](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/index.html) and
 [schemas](https://www.tei-c.org/guidelines/customization/). If you want to work on the Guidelines, you should look in the [P5 directory](https://github.com/TEIC/TEI/tree/dev/P5).
 
-Some notes on working with Git and on this GitHub repository may be found in the following documents:
+Some notes on working with Git and on this GitHub repository may be found in the following document:
 * [Git README](https://github.com/TEIC/TEI/blob/master/Documents/Git-README.md)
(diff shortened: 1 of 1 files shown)


sydb (commented):
Although this particular change is fine (and could be merged, I think), it would be better to solve the “where did how to use GitHub go” problem. And if we don’t solve that problem, it would probably be better to re-word entire sentence. (After all, the only document pointed to could be just a link within the sentence, and is not a general-purpose guide but obviously aimed directly at new Council members. And could be named something better than README, no?)

GusRiva:
Building on Syd's comment: I found the page on the WaybackMachine:

https://web.archive.org/web/20181115011630/https://www.tei-c.org/guidelines/p5/using-the-tei-github-repository/#expand 

Unfortunately, I am not able to find this content anywhere in the TEI GitHub repos. Where could it be? Do we want to keep this content and republish it?

hcayless:
I think this page is obsolete and the correct course is to remove the link rather than try to restore the page. Some of the notes on repository structure could usefully be imported into the GitHub Readme file. "Git README" link may be obsolete too. I wrote it about 9 years ago to help Council get off Subversion and into working with Git. Those days are long gone.

ebeshero:
As we all agree that we should remove the dead link, I'm just merging this now. Thanks @rettinghaus !

rettinghaus:
Found two more occurances of this: `using-the-tei-github-repository`

One in the [index](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/index.html): 
[P5/Source/Guidelines/index.xml](https://github.com/TEIC/TEI/blob/4e6b08ef5986bd63ac87ec87cad28e86ed1c35a2/P5/Source/Guidelines/index.xml#L25)

And another one here: 
[P5/Utilities/guidelines.xsl.model](https://github.com/TEIC/TEI/blob/4e6b08ef5986bd63ac87ec87cad28e86ed1c35a2/P5/Utilities/guidelines.xsl.model#L600)

PR #2615, opened 2024-11-13
Removal of calendar from provided customization ODDs

raffazizzi:
Additional work for completion of https://github.com/TEIC/TEI/issues/2045

### DIFF
--- a/P5/Exemplars/tei_jtei.odd
+++ b/P5/Exemplars/tei_jtei.odd
@@ -1948,14 +1948,12 @@
             <attList>
               <attDef ident="notAfter" mode="delete"/>
               <attDef ident="role" mode="delete"/>
-              <attDef ident="calendar" mode="delete"/> <!-- remove 2024-11-11 when @calender removed from <affliation> -->
             </attList>
           </elementSpec>
           <elementSpec ident="application" mode="change">
             <attList>
               <attDef ident="type" mode="delete"/>
               <attDef ident="notAfter" mode="delete"/>
-              <attDef ident="calendar" mode="delete"/> <!-- remove 2024-11-11 when @calender removed from <application> -->
             </attList>
           </elementSpec>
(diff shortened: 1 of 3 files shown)


sydb:
This PR is just waiting for someone (probably @GusRiva, but anyone, really) to review it and in particular to make sure tei_jtei, tei_lite, and tei_tite build properly and have `@calendar` where they are supposed to. IMHO, once that quick review & test are done, if there are no problems this could just be merged without further discussion.

GusRiva:
I am may be understanding something wrong, but I think there might be a problem with att.calendarSystem in tei_lite.odd and tei_jtei
The `<date>` element is part of tei_lite but does not have att.calendarSystem listed as tei_all does. 
I see the line:

`<classSpec ident="att.calendarSystem" module="tei" type="atts" mode="delete"/>` 

So I assume this is intentional? I couldn't find this in the discussion of the issue.

GusRiva:
I traced back the removal of the calendarSystem from tei_lite and tei_jtei to this commit:

7006f1d108815fb2f5c3165c6607e0d1a3fef8b1 

That same day the individual attributes were removed manually: 5bcc343df5c8e3bf268c72cf23522a6e495f25d5 

And the day before this related commit happened:

4e4bac5f91759e1af9b20f85fd6809ac7175e186

The way I see it: either 
1- we want calendar in `<date>, <time>, <origDate>, and <docDate>` in lite and jtei in which case we should remove the line:

`<classSpec ident="att.calendarSystem" module="tei" type="atts" mode="delete"/> `

2- We don't want it, in which case we can just merge.

Edit: calendar was always removed in lite, jtei and tite, so we want to remove it still now as part of the new class

GusRiva (approved):
Looks good, the deprecated parts where calendar was att.datable are out and calendar is being removed properly as part of att.calendarSystem

PR #2616, opened 2024-11-14
towards defining and describing the dependency att class

bansp:
This PR is meant to illustrate the modifications by means of GitHub diff tools.


bansp:
Whoops, that was never meant to go against the TEIC/dev! :-)
Sorry. Late night tinkering.

Issue #2617, opened 2024-11-16
TEI Modules missing in table


martinascholger:
The TEI modules 9 [Computer-mediated Communication](https://tei-c.org/release/doc/tei-p5-doc/en/html/CMC.html), 21 [Non-hierarchical Structures](https://tei-c.org/release/doc/tei-p5-doc/en/html/NH.html), and 24 [Using the TEI](https://tei-c.org/release/doc/tei-p5-doc/en/html/USE.html) are missing in the table provided in subchapter https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ST.html#STMA

sydb:
See PR #2618.

martinascholger:
This was implemented with PR #2618 -- so I'm closing the issue.

PR #2618, opened 2024-11-16
table of modules — re-generated based on new utility

sydb:
Note that new table reverses the position of the “Where defined” and “FPI” columns. (I think _very_ few readers care about the FPI column at all.)
Intended to address #2617. Yes, it would be better if the build process automagically generated and inserted the table. But that would take a lot longer, and we do not change modules frequently. (I am not saying we should not do so, only that I do not personally want to do so right now.)


ebeshero:
This XSLT is nifty, @sydb ! I think maybe we should find an easy way (if there is one) to incorporate it in the build process, but I agree that it is not urgently necessary.

trishaoconnor:
This seems ready to go, pending someone to build to review the output and either merge or discuss any issues with Council.

GusRiva:
I like that now the table is sorted in the order of the guideline chapters and not alphabetically by module name.

I think there is potentially still something to correct. In the original issue it was mentioned that 3 modules were missing:  

> 9 [Computer-mediated Communication](https://tei-c.org/release/doc/tei-p5-doc/en/html/CMC.html), 21 [Non-hierarchical Structures](https://tei-c.org/release/doc/tei-p5-doc/en/html/NH.html), and 24 [Using the TEI](https://tei-c.org/release/doc/tei-p5-doc/en/html/USE.html) 

In the new table we now have 9, but 21 and 24 are still missing.  Is this correct? I guess these chapters don't correspond to any particular module and are not included in the table?

lb42:
21 and 24 do not define a module

GusRiva (approved):

PR #2619, opened 2024-11-25
Fix gloss

rettinghaus:
This is a minor clean-up PR removing white space before the closing tags of `gloss` entries.

grafik


trishaoconnor:
Thank you for your work correcting this.

Issue #2623, opened 2024-11-30
search-and-replace a tad too greedy in att.datcat.xml (or uncovering a deeper, scary truth of a wicked opening tag)


bansp:
While trying to investigate the source of a recent breakage in the dev source, I have come across something that probably does not end my search, but is worth noting in case I've overlooked other such cases, and may result in a fix in the culprit search regex, perhaps:

In commit https://github.com/TEIC/TEI/commit/b19701c763dba63aea39fe949ed3ee0ee3fe35d9 (pointing a finger at @sydb , hahah ;-)), there's at least one change too many, resulting in:

```xml
<egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="DATCAT-egXML-lh"> xmlns="http://www.tei-c.org/ns/Examples" xml:id="DATCAT-egXML-xk"&gt;
```

I'm going to post a PR for that presently, But it was by pure luck that I've spotted that one -- perhaps other analogous fixes are still needed. Note that Syd's s-a-r may actually be completely innocent, and simply received totally unexpected input -- because something is very wrong about that opening tag above.

ebeshero:
@bansp We did find (today!) the source of the broken dev build introduced last week, and we merged one of @sydb 's PRs on Stylesheets which fixed the build. (It was to do with handling the egXML namespace in Schematron constraints.) But this discovery of yours does call for a closer look to see how and when it was introduced...in case of more like it!

ebeshero:
Noting that this is now associated with PR https://github.com/TEIC/TEI/pull/2624

PR #2624, opened 2024-11-30
Fix the opening tag (#2623)

bansp:
(I just hope it's not a more widespread issue -- I got completely no warning from the editor, it was all green, even when I just replaced the entity with `>`)


sydb:
Too obvious to bother with review, it is an obvious corrigible error.
However, I am going to check for others …

PR #2625, opened 2024-11-30
add the missing entities for angle brackets

bansp:
This one is very loosely connected to #2623 in that I just continue a very shallow inspection of that commit. 
This seems to be a case where the regex did not catch the brackets to be converted to entities (or I miss something important, and hence the request for @sydb to inspect the PR)


bansp:
Ah no, it's nonsense, apologies for this.

PR #2627, opened 2024-12-03
re-write of 1st para on namespaces

sydb:
Install @lb42’s suggested re-wording for 1st para of section `#SGname`.


martindholmes (approved):
Nice elegant prose, but whenever I read this sort of thing I flinch on behalf of the second-language readers of the Guidelines who might gladly forgo this fulsome eloquence in favour of shorter, simpler sentences.

Issue #2516, opened 2023-12-21
Allow `@scope` in `//profileDesc/langUsage/language`


cthomasdta:
Following the thread via TEI Mailing starting here List https://listserv.brown.edu/cgi-bin/wa?A2=TEI-L;dea9ab8d.2311, we propose to allow `@scope` in `//profileDesc/langUsage/language` to annotate the `sole`/`major`/`minor` language used in a document.

The existing `@usage` attribute is not sufficient here, especially since %-Values are hard to determine with the exactitude the attribute use seems to be suggesting, and can hardly be determined with some degree of reliability at all before the text is transcribed. But it would be no problem, and also be efficient because we use an existing attribute and its proposed values, if the `@scope` attribute was allowed here, with its values `"sole"`, `"major"` and `"minor"`.

To quote from my initial message via TEI-L:
> Currently, `@scope` in this function is available in `<handNote>` and "specifies how widely this hand is used in the manuscript" (https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.handFeatures.html#tei_att.scope). We would suggest that `@scope` be made available in `<language>`, as well, to specify "how widely this language is used in the manuscript [or document in general]".

ebeshero:
Thanks @cthomasdta! The proposal made sense to me when you posted it on the TEI-L. Considering how to implement, I think it doesn't make sense to add language to att.dimensions since that's designated for physical dimensions of a document. If we do this, I think we would add `@scope` directly to language. 

But I wonder if this is necessary? We have textLang, which seems to include simple verbal explanations of the languages in a document. 
https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-textLang.html

Comparing the use of textLang vs. `@scope` on language, I do think `@scope` provides a little more precision, so I endorse this proposal and think we should proceed.

sydb:
I think I generally agree with @ebeshero , except to say that we want a version of the `@scope` of att.handFeatures, not that of att.dimensions, anyway. My instinct is to factor out `@scope` from att.handFeatures to its own class (att.impreciseScope or some such), and then ask that both `<language>` and att.handFeatures claim membership in the new class.

ebeshero:
Completed in PR #2631

PR #2629, opened 2024-12-04
p5odds.odd bug fix

sydb:
The constraint on `<eg>` was looking for _all_ descendant text nodes, not just child text nodes. That means that if there is any phrase-level encoding in an `<eg>`  then there is more than one text node that has (for example) `position()=last()`, and the XPath fails because only one item is allowed as the first argument to `matches()`.

Fix seems trivial: search only for child text nodes, not descendant text nodes.

Note: I am not sure the “TEI: ODD” label is correctly — it might be for the TEI language defined _by_ the Guidelines, not the TEI language defined _for_ the Guidelines. Sigh. Anyway, this is **high** priority because it is holding up work on #2516.

### DIFF
--- a/P5/p5odds.odd
+++ b/P5/p5odds.odd
@@ -184,8 +184,8 @@
             </desc>
             <constraint>
               <sch:rule context="tei:eg">
-                <sch:report test="matches( .//text()[last()], '&#x0A;\s*$')">trailing newline not allowed</sch:report>
-                <sch:report test="matches( .//text()[1],      '^\s*&#x0A;')">leading newline not allowed</sch:report>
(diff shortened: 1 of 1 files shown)


sydb:
Correction to OP: @martindholmes suggests that we, instead of attempting to get the right text() node to test, just test the entire string value of the `<eg>`. I think he is right, so have switched test to use `string(.)` instead. (Strictly speaking the call to `string()` is not necessary, as `matches()` is expecting a string as its 1st argument and thus will automagically convert, but calling the function makes what we are doing explicit.)

martindholmes (approved):
This looks good.

PR #2630, opened 2024-12-04
p5odds: Allow "URN" as value of `@type` of `<idno>`

sydb:
Per Slack conversation with @raffazizzi , add "URN" as possible value to `@type` of `<idno>` in p5odds. While I was there I also:

- Changed "url" to "URL"
- Added "URI"
- Updated glosses & descriptions a bit

I have checked. Use of `<idno type="URN">` results in output that is not really pretty in the HTML _Guidelines_ (specifically, in the Bibliography), but not particularly ugly, either, and does not falsely imply that your web browser will open a URN. (Which use of `<ptr target="urn:duck:billed:platypus"/>` would.)

I have also checked, and there were no occurrences of `<idno type="url">` in P5, so changing this to "URL" should not cause problems.

### DIFF
--- a/P5/p5odds.odd
+++ b/P5/p5odds.odd
@@ -44,12 +44,12 @@
         <moduleRef url="Exemplars/mathml2-main.rng"/>
         <moduleRef url="Exemplars/relaxng.rng"/>
 
-	<elementSpec module="tagdocs" ident="constraintSpec" mode="change">
-	  <attList>
-	    <attDef mode="change" ident="xml:lang" usage="req"/>
-	  </attList>
-	</elementSpec>
-	
+        <elementSpec module="tagdocs" ident="constraintSpec" mode="change">
+          <attList>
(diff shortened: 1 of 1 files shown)


sydb:
Note: I am not sure the “TEI: ODD” label is correctly applied to this — it might be for the TEI language defined _by_ the Guidelines, not the TEI language defined _for_ the Guidelines. Sigh. Anyway, this is high priority because it is holding up work on https://github.com/TEIC/TEI/issues/2516.

joeytakeda (commented):
All tests pass and the changes seem reasonable to me.

martindholmes (approved):
Looks good. Upper case for URL brings it in line with the others in P5.

ebeshero:
I've tested and built guidelines and exemplars with it. All good, and the update is definitely useful in the cause of new and wider-ranging examples of `<idno>` for the Guidelines. As it's an urgent PR holding up other tickets, I'm merging it now.

Issue #2367, opened 2022-10-26
"typographic" (`lb`) vs. "topographic" (`line`)


gerritbruening:
The wording of the elements' definitions of `lb` and `line` slightly differ: 

> `<lb>` (line beginning) marks the beginning of a new (t**y**pographic) line ...

> `<line>` contains the transcription of a t**o**pographic line ...

Assuming that this is not intended, but historical, I would like to suggest the following:

1. Check whether "topographic" is appropriate in the given context.
2. If it is, use "topographic" in both definitions, since "typographic" refers to printed material only.

hcayless:
"Typographic" is clearly incorrect. "Topographic" must be metaphorical, and maybe there's a specialist use of the term to describe writing on a surface. But it's a metaphor that's never explained that I can see. I'd favor getting rid of it, or replacing it with something clearer.

lb42:
"Typographic" is only incorrect if you are talking about non print materials, surely. The point there is to distinguish a typographic line from e.g. a metrical line. As for "topographic" I don't see why you think it's metaphorical or unclear. The Guidelines say "Within a zone the transcription may be organized topographically in terms of lines of writing, using the line element, or in terms of further nested zones, or as a combination of the two" .  And a random online Oxford dictionary defines "topoographical" as "relating to the arrangement or accurate representation of the physical features of an area". Which seems exactly right here.  So maybe what  needs revision is the discussion of what `<lb/>` means.

martindholmes:
I agree with @lb42; this distinction seems intentional, apt, and accurate.

hcayless:
1) Yes, but `<lb/>` is a perfectly usable element in encoding non-print materials, so calling it "typographic" doesn't add anything. And might make people think it isn't appropriate for non-typographic texts.
2) "Topographical" features of a text implies the text is like a map (there's our metaphor) and the markup is like (e.g.) contour lines. And that's a reasonable analogy, but I'd prefer we either explain it more clearly or stop using it.

trishaoconnor:
Action on @trishaoconnor to rephrase the desc of <lb/>  according to @gerritbruening proposal: "...marks the beginning of a topographic line". 

@sabineseifert also notes that Council needs to update the outdated prose in the third line of note and the example in `<lb>`

>  "This element is intended to be used for marking actual line breaks"

 change to 

> "This element is intended to be used for marking the line beginning"

[...]

PR #2633, opened 2024-12-10
Updated specs for `<line>` and `<lb>`

trishaoconnor:
Updated the description of `<lb>` according to @gerritbruening proposal to 

> ...marks the beginning of a topographic line in some edition or version of a text

Also updated the outdated prose in remarks and the examples for `<lb>`, according to @sabineseifert's request (see [#2367](https://github.com/TEIC/TEI/issues/2367) ), to change mentions of 'line breaks' to 'line beginnings'

Also addressed the related comment for the <line> element Spec page: to add a sentence of clarification for the first example on the `<line>` spec page, since the note specifically states to **not** use `<line>` for encoding a poem yet the first example provided is of a poem being encoded using `<line>`.

The clarification that I added for `<line>` is adapted from the clarification provided in `<lb>`:

> This example shows typographical line breaks within metrical lines, where they occur at different places in different editions

The proposed update for the <line> example is copied below:

```
<exemplum xml:lang="en">
    <p>This example shows topographical lines as a means of preserving the visual appearance of a poem:</p>
      <surface>
        <zone>
          <line>Poem</line>
          <line>As in Visions of — at</line>
          <line>night —</line>
          <line>All sorts of fancies running through</line>
          <line>the head</line>
        </zone>
      </surface>
    </egXML>
  </exemplum>
```


P5/Source/Specs/lb.xml
```diff
@@ -32,7 +32,7 @@
   </classes>
   <content><empty/></content>
   <exemplum xml:lang="en">
-    <p>This example shows typographical line breaks within metrical lines, where they occur at different
+    <p>This example shows topographical line beginning within metrical lines, where they occur at different
```
sydb: At the very least, I think “beginning” should be plural.
But I am not really fond of using it as a noun like that at all. How about “This example shows the encoding of the beginning of each topographic line as it occurs in both the 1664 and 1667 editions” or some such?

P5/Source/Specs/lb.xml
```diff
@@ -89,15 +89,15 @@ soi-<lb break="no"/>xante &amp; dixhuytiesme livre.
       starts. The <att>n</att> attribute, if used, indicates the number or other value associated
       with the text between this point and the next <gi>lb</gi> element, typically the sequence
       number of the line within the page, or other appropriate unit. This element is intended to be
-      used for marking actual line breaks on a manuscript or printed page, at the point where they
+      used for marking line beginnings on a manuscript or printed page, at the point where they
```
sydb: As previous comment, perhaps “… used for marking the beginning of each line on a …”?

P5/Source/Specs/lb.xml
```diff
@@ -89,15 +89,15 @@ soi-<lb break="no"/>xante &amp; dixhuytiesme livre.
...
       occur; it should not be used to tag structural units such as lines of verse (for which the
         <gi>l</gi> element is available) except in circumstances where structural units cannot
       otherwise be marked. </p>
     <p>The <att>type</att> attribute may be used to characterize the
-    line break in any respect. The more specialized attributes
+    line beginning in any respect. The more specialized attributes
     <att>break</att>, <att>ed</att>, or <att>edRef</att> should be preferred when the
     intent is to indicate whether or not the line
-    break is word-breaking, or to note the source from which it
+    beginning is word-breaking, or to note the source from which it
```
sydb: Seems to me this one could just stay “break”, or perhaps “… whether or not the new topographic line implies a word break, or to note …”?

sydb (commented):
Overall a definite improvement, so I think it _could_ be merged as it is. That said, I think we can do a bit better while we’re at it. See individual word tweak suggestions. They may need to be improved upon, too, but I think are another step in the right direction.

P5/Source/Specs/lb.xml
trishaoconnor: Agreed! It definitely should have been plural, that is entirely a slip on my part. I also agree with your rewording of the sentence.

P5/Source/Specs/lb.xml
trishaoconnor: Thank you. Yes, I agree.

P5/Source/Specs/lb.xml
trishaoconnor: I prefer the latter, personally, but can be swayed to revert to 'break' if that is the consensus. Thank you for your feedback.

gerritbruening: I find "line break" more natural in this specific context.

P5/Source/Specs/lb.xml
```diff
@@ -89,15 +88,15 @@ soi-<lb break="no"/>xante &amp; dixhuytiesme livre.
...
       occur; it should not be used to tag structural units such as lines of verse (for which the
         <gi>l</gi> element is available) except in circumstances where structural units cannot
       otherwise be marked. </p>
     <p>The <att>type</att> attribute may be used to characterize the
     line beginning in any respect. The more specialized attributes
     <att>break</att>, <att>ed</att>, or <att>edRef</att> should be preferred when the
-    intent is to indicate whether or not the line
-    beginning is word-breaking, or to note the source from which it
+      intent is to indicate whether or not the beginning of a new line
+     is word-breaking, or to note the source from which it
```
trishaoconnor, line 99: @gerritbruening @sydb @sabineseifert I updated the phrasing of this section last night, but I can revert back to 'line break' if you prefer?

sabineseifert (approved):

ebeshero:
@sydb and/or @gerritbruening Would you have a look at this PR and see if you agree it's ready to go? I think it's a definite improvement.  We're getting ready for a release at the end of January and I think we should include this.

P5/Source/Specs/lb.xml
sydb: Oops. I led you down a treacherous path. We now have disagreement in number:

> the beginning of each topographic line …, where they occur

(Because “each” is singular but “they occur” is plural.)

So either something like “This example shows the encoding of the beginning of each new topographic line within a metrical line, indicating where it occurred in both the 1667 and 1674 editions:” or “This example shows the encoding of the beginning of all topographic lines that occur within metrical lines, indicating where they occur in both the 1667 and 1674 editions:” or some such.

P5/Source/Specs/lb.xml
sydb: I do not have strong feelings either way at the moment.

sydb:
Except for the disagreement in number flagged above, I think this could be merged.

gerritbruening:
> I think this could be merged.

I agree, thank you all for considering #2367!

trishaoconnor:
Thank you @sydb for alerting me to the discrepancy. I'll update section right away and get this ready for merge. 
Thank you all @gerritbruening @sydb @sabineseifert and @ebeshero your reviews.

trishaoconnor:
Thank you for you @sydb @gerritbruening for your patience. I have added those changes now.
If you are happy with the updates, then @torstenroeder can merge once all the checks pass again.

PR #2635, opened 2024-12-14
Resolve 2026: Restore missing part of example in guidelines chapter 23

### LINKED ISSUE #2060: Prose in "Element and Attribute Descriptions" (TDphraseEA) refers to non-existent <ptr> in example
https://github.com/TEIC/TEI/blob/9e6d7fb88717edf3befe52f83aceb315ec8298a7/P5/Source/Guidelines/en/TD-DocumentationElements.xml#L156

The prose below the example starts with "When formatting the <gi>ptr</gi> element in this example..." but the example does not contain a &lt;ptr> element.

GusRiva:
Addresses issue: #2060 

Restores a required part of an example that was then mentioned in the next paragraph. The missing text had been deleted when updating the example in a previous commit.


ebeshero (approved):
This edit looks good to restore the missing `<ptr>` element from the example.

sydb:
The HTML looks fine (which is to say the important bit, the `<ptr>`, has been restored and the example looks no worse than our other examples) and it passed tests in my Docker. Thus merging.

PR #2638, opened 2024-12-16
Make rotate attribute of type teidata.numeric

### LINKED ISSUE #2637: zone can only be rotated by full degrees
Currently `zone@rotate` is of type [teidata.count](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-teidata.count.html). 

If the zone is rotated counterclockwise by 1 degree, you have to enter "359". (Somewhat annoying, but doable.)
But what about a roation of 3.5 or 10.2 degrees? 

Shouldn't it also be of type `teidata.numeric`?

rettinghaus:
This PR changes `zone@rotate` into `teidata.numeric` in order to allow negative and non-integer values.
Closes #2637

### DIFF
--- a/P5/Source/Specs/zone.xml
+++ b/P5/Source/Specs/zone.xml
@@ -41,7 +41,7 @@ element.</desc>
         Als Bezugspunkt gilt dabei die natürliche Ausrichtung des <gi>surface</gi>-Elternelements, die entweder
         im <gi>msDesc</gi>-Element beschrieben ist oder durch die Koordinaten des <gi>surface</gi>-Elements selbst.
         Die Drehung wird in Bogengrad angegeben.</desc>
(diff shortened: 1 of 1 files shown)


ebeshero:
@rettinghaus Thanks for this! Council members raise a good question of why we might have limited ourselves to positive integers only in the first place, so we want to discuss this first before deciding whether to merge!

martindholmes:
There isn't a single example of the use of `@rotate` in the Guidelines, nor is it mentioned anywhere in the prose. That's no reason to reject this PR, but it would be good to see some examples. This might be an easy one to start with:

https://graves.uvic.ca/gr01_0472_004.html

martindholmes:
If the example above would be useful, the rotation of the bullfight ticket from its proper orientation is -87.92 degrees.

Issue #2551, opened 2024-04-19
Allow `@place` in `<div>` 


cthomasdta:
_Originally posted by @cthomasdta in https://github.com/TEIC/TEI/issues/2542#issuecomment-2022474764_ [Code-example below slightly modified to clarify my point]

[@lb42 asked:]

> […] Why not use seg ?

In my view, a question of priorities, what we want to express: `<seg>` is very generic, whereas the most important info imo is that this _is_ a `<div>`, more specifically a `<div type="writingSession" n="3">`, _exactly like_ the two other `<div type="writingSession">` that the letter consists of. 

So we have a clear succession of the same elements for the respective text containers (and not `div`-`div`-*`seg`* -- as if the last one was something else than the two blocks before that):

```XML
<div type="letter">
   <div type="writingSession" n="1"><p>[…]</p></div>
   <div type="writingSession" n="2"><p>[…]</p></div>
   <div type="writingSession" n="3" place="top" rend="upsideDown"><p>[…]</p></div>
</div>
```

The (again, imo) a little less important info, but still worth noting preferably _as an attribute of this **`div`**_, is that the third writing session was squeezed in on the top of the page (and written upside-down). Since there is something exactly fitting this annotation demand, namely `@place` (and maybe `@rend`), I would like to be able to use it there. [Btw I would argue the same for `<head>`, but that again would be another issue :)]

[…]

sydb:
I have to admit, “writing session” does not seem like a viable categorization of logical textual divisions to me. (I concede I may be convinced otherwise.) It seems to me that writing sessions and logical textual divisions are orthogonal to one another. I may very well have written chapters 1, 2, and the first 3½ sentences of chapter 3 in one writing session, and the last 97½ sentences of chapter 3 in the next, no? Thus my instinct is that writing sessions are better left to revision campaigns or `<sourceDoc>` encoding.

cthomasdta:
> I have to admit, “writing session” does not seem like a viable categorization of logical textual divisions to me. (I concede I may be convinced otherwise.) 

Hi Syd, I'll have to think about that a little longer to maybe convince you – or to concede that you are right.

But for the time being, can we agree that this concern does not relate to the initial request, i.e. allow `@place` in `<div>`? This still seems reasonable to me and I would like to focus on this first. Thanks again!

sabineseifert:
European subgroup at VF2F April 27:

[...]

PR #2643, opened 2024-12-23
Create new att.source and include att.written and att.placement in it

### LINKED ISSUE #2550: Allow `@hand` in `<emph>`; Allow `@hand` in `<dateline>`
Dear all, I was checking if the issue was already raised and stumbled upon https://github.com/TEIC/TEI/issues/23#issuecomment-2061723249

> 3. make the `@hand` attribute available on `<app>`, `<hi>`, and `<emph>`

This seems not to have happened, at least in [`<emph>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-emph.html), `@hand` is not allowed. 

So I ask for `@hand` to be allowed in `<emph>`. We want to use it the same way as `@hand` in `<hi>`, i.e. when another scribe is highlighting text (by underlining it for example), either as a more "typographical" sign (`<hi>`) or as a stronger emphasis, hence we need `<emph>` also.

---

In the same direction I want to ask for `@hand` to be allowed in `<dateline>`, following up on a discussion started by Denise Jurst-Görlach here https://listserv.brown.edu/cgi-bin/wa?A2=TEI-L;d9f6e911.2403. There was no opposition on the list, on the contrary. So this also, as the previous request here, should not be problematic.

sabineseifert:
and also added note to description of att.placement concerning its use.

An 'good' example gently guiding the use of `place` in e.g. `div` is still missing.


ebeshero (changes requested):
@sabineseifert About this, I was just re-reading our minutes from Buenos Aires. It seems we were going to move the classes `att.written` and `att.placement` into a new class called `att.source`. And we left some guidance about parallel cases to help with this. Take a look (you can find the discussion in [the F2F minutes](https://tei-c.org/activities/council/meetings/tei-technical-council-f2f-meeting-in-buenos-aires-6-8-october-2024/) by searching for `@place` on the posted minutes on the website.)

sabineseifert:
This PR was tied to #2551 but is now tied to #2550.

What I did:

- created new class `att.source` -- Please review carefully what might be missing/wrong here. (E.g. is there an `xml:id` needed in `<classSpec>`, like `xml:id="class-attr-source"`?)
- added `att.written` and `att.placement` to new `att.source` (both as discussed at F2F in [October 2024](https://tei-c.org/activities/council/meetings/tei-technical-council-f2f-meeting-in-buenos-aires-6-8-october-2024/))
- added a remark to `att.placement` as discussed in Council meeting [2024-09-03](https://tei-c.org/activities/council/meetings/tei-technical-council-teleconference-2024-09-03/) (in relation to #2551)

What I did not:
- not yet added `att.source` to `att.global.rendition` (see #2551)

sydb:
@sabineseifert — I presume this should currently be considered a DRAFT pull request, yes?

Although you have _created_ att.source, you have not added it to the _Guidelines_. It needs to be included in the list at #DSTECAT in Source/Guidelines/en/ST-Infrastructure.xml. (Starts on line ~864.) Adding it is trivially easy. Adding it _in the right place_ can sometimes be tricky.

As for `@xml:id` on `<classSpec>`, we use it about ½ the time. Will look in a moment at how often we reference it, but I suspect it is very rarely.

Could someone remind me why we are calling this new class “att.source” — seems confusingly close to att.global.source. Maybe att.renditionLike, att.styleLike, att.sourceDetails, att.specificRendition, or some such?

sydb:
Of the just over 100 `@xml:id` attributes we have on the over 200 `<classSpec>`s in the _Guidelines_ source, exactly 0 of them are referenced.

sabineseifert:
Good idea to mark this as a draft PR which I just did!

I looked at the ST-Infrastructure.xml and must confess that I wouldn't know where to add it. But before, we need to settle the question of the name anyway: `att.source.rendition` and `att.source` were proposed at the F2F in [October 2024](https://tei-c.org/activities/council/meetings/tei-technical-council-f2f-meeting-in-buenos-aires-6-8-october-2024/)). And: the parallel case `model.phrase` and `model.limitedPhrase` might help with naming and defining the new class. But I think we did not finish discussing the name. 

And concerning the 0 references to the over 100 `@xml:id`s: We probably should not keep on going adding these but we perhaps should not delete the already existing ones?

Thanks, @sydb, for looking into this!

sydb:
I am ducking the name issue for the moment, mostly because I do not really know what to call this thing.

But as for where in ST the `<xi:include href="../../Specs/att.WHATEVER.xml"/>` goes, it should be inserted somewhere between lines 864 and 929. As for where within that list, the only requirement (IIRC) is that an attribute class which has a `<memberOf key="DUCK">` has to be _after_ the `<xi:include href="../../Specs/att.DUCK.xml"/>`. Thus those includes that refer to classes do NOT have a `<memberOf>` can go anywhere, and we generally list them alphabetically.

About ½ of our attribute classes are a member of some other class. But of course, an average by accident roughly ½ of those should be in the correct order without intervention. I am working on re-organizing that list and commenting it better, but will not get to check anything in for hours, I am sure.

As for adding `@xml:id` to `<classSpec>` — I do not think it matters.

sydb:
I am going to propose a different solution.

I note that the majority of elements that are in att.placement are already in att.written. Thus I am wondering (aloud, first to @ebeshero then here) if we should perhaps just add att.placement to att.written. The result would be only 4 elements that currently do not have `@hand` gaining it: `<metaMark>`, `<notatedMusic>`, `<rt>`, and `<witDetail>`.

This might not work, because some elements that are in att.placement are already in att.placement, so we might have some conflict problems. But it is worth investigating.

Here are the elements that are either directly or indirectly in att.placement and att.written:

| placement | written |
| --------- | ------- |
|              | ab	|
| add          | add	|
| addSpan      | addSpan	|
|              | closer	|
|              | damage	|
|              | damageSpan	|
|              | dateline	|
|              | del	|
|              | delSpan	|
|              | div	|
|              | emph	|
| figure       | figure	|
| fw           | fw	|
| head         | head	|
|              | hi	|
| label        | label	|
| metamark     |	|
|              | lem	|
|              | line	|
|              | mod	|
| notatedMusic |	|
| note         | note	|
| noteGrp      | noteGrp	|
| rt           |	|
|              | opener	|
|              | p	|
|              | path	|
|              | postscript	|
|              | rdg	|
|              | rdgGrp	|
|              | redo	|
|              | restore	|
|              | retrace	|
|              | rt	|
|              | salute	|
|              | seg	|
|              | signed	|
| stage        | stage	|
|              | subst	|
|              | substJoin	|
|              | text	|
| trailer      | trailer	|
|              | undo	|
| witDetail    |	|
|              | zone	|

ebeshero:
@sydb Reviewing the classes `att.written` and `att.placement` just now, I noticed that `att.written` is a member of `att.transcriptional`, and wonder if it might be easier to treat `att.transcriptional` as a class that could also include `att.placement`? 

https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.transcriptional.html

sydb:
As sort of expected, the error “error: duplicate attribute "hand"” occurs almost immediately on trying the above proposal. So the question becomes, how much class-jiggering would be needed to avoid that problem, and is it worth it?

ebeshero:
@sydb I wonder if the common ancestor idea is still a good one: `att.written` is a member of `att.transcriptional`. What about also including `att.placement` in `att.transcriptional`? Doesn't that effectively do what we initially proposed in Buenos Aires, and prevent some of the conflict, too? It also seems like the appropriate class--the description seems more applicable to `att.placement` than the description of `att.written`.

https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.transcriptional.html

martindholmes:
I really don't like the name att.source. It's going to be completely confusing; most of these attributes have nothing to do with the source of anything, except that they're used in transcription, and transcription is usually of some kind of source.

ebeshero:
Work on the issues to be covered by this PR was moved to another PR here: https://github.com/TEIC/TEI/pull/2665

ebeshero:
Work on the issues to be covered by this PR was moved to another PR: https://github.com/TEIC/TEI/pull/2665
This branch wasn't passing tests and we started down a different path for a solution!

PR #2646, opened 2024-12-27
Added remarks to ptr and corrected sch:report

### LINKED ISSUE #2210: inconsistencies between `<ptr>` and `<ref>`
The specification for `<ptr>` has `<remarks>` that are not mirrored in `<ref>` (either they both should have the remarks, or neither should).
The `<sch:report>`s in the two elements are there for identical reasons, but their content is not identical. (It should be.)

sabineseifert:
Closes #2210

### DIFF
--- a/P5/Source/Specs/ptr.xml
+++ b/P5/Source/Specs/ptr.xml
@@ -41,6 +41,18 @@
       <ptr cRef="1.3.4"/>
     </egXML>
   </exemplum>
+  <remarks versionDate="2005-10-13" xml:lang="en">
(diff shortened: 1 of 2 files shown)


sydb:
Passed all tests in my docker. Looks good. (If you would like to see for yourself check out [Exemplars/](http://bauman.zapto.org/~syd/temp/4TEICouncil/issue_2210-ss_outputs/Exemplars/) and [Guidelines-web/](http://bauman.zapto.org/~syd/temp/4TEICouncil/issue_2210-ss_outputs/Guidelines-web/en/html/.))

PR #2647, opened 2025-01-01
improve editability of inclusion of attribute classes into ST

sydb:
On trying to explain attribute specification inclusion on #2643, it occurred to me that it would be more useful to future editors to have the explanation _in ST_ (the "tei" module) itself. While I was there I also re-arranged the order of attribute class inclusions, hopefully in a way that is easier to understand. No changes to prose or schema, just re-arranged the sequence of attribute specifications and added comments to explain it.

This necessitated one minor change in the tests, as the order of Schematron constraints (some of which are specified in the attribute classes) is now a bit different.

Note to potential reviewers: I think passing tests is probably sufficient (and they all pass in my Docker). But if you look at the files themselves, definitely hide whitespace differences.


P5/Source/Guidelines/en/ST-Infrastructure.xml
```diff
@@ -862,70 +862,85 @@ one, or vice versa, should be done with care. </p>
...
+                possible. Thus we specify the classes that are NOT
+                members of another class first, in set 1. Then, in set
+                2, we specify those classes that are a member of a
+                class in set 1. Then, in set 3, we specify those that
+                are a member of a class in set 2. (We have not needed
+                a set 4 yet, but might someday.)
+
+                Within each set classes are specified in alpahabetic
+                order.
+            -->
```
ebeshero, line 885: Thanks, @sydb ! This comment is super helpful "future proofing" for editors of this page. :-)

ebeshero (approved):
This passes tests and doesn't break the build for me locally. Also the big table in ST just looks a lot better and is organized more logically in the resulting build of the Guidelines, though I think that might be paired with another recent update in Stylesheets on my dev branch.

PR #2648, opened 2025-01-03
Normalizing Japanese punctuation per issue #2337.

martindholmes:
Not sure who to choose to review this -- I don't see any native Japanese speakers in the list of people -- but it's very basic stuff and I can explain it to anyone.


sydb (approved):
According to [Wikipedia](https://en.wikipedia.org/wiki/Comma#East_Asia) the IDEOGRAPHIC COMMA (U+3001) is used in Japanese where we use COMMA (U+002C) as a separator of items in a list. All of these appear to be that kind of case. 

Furthermore, in all 5 cases of `//*[@xml:lang eq 'ja'][contains( ., ',')]` the U+002C does not seem to be a list item separator.

(That said, I think an ` xml:lang="en"` should be added to #MEASUREMENT-egXML-cw. This need not be part of this PR, of course.)

PR #2654, opened 2025-01-14
allow nested facsimile elements

### LINKED ISSUE #2565: Allow nested <facsimile>
After some debate on the TEI-L (https://lists.psu.edu/cgi-bin/wa?A2=TEI-L;f4bdaac4.2406&S=), I would like to raise the topic of nested `<facsimile>`. 

The starting premise is that a `<facsimile>` element contains a set of images that represent a material artefact. In the description of a composite manuscript, we introduce `<msPart>` elements which represent a single material part of that primary source as an artefact. If this is the case, then it would be reasonable that `<facsimile>` will have nested facsimile parts corresponding to those material parts of the manuscript: 

`<facsimle>`
`    <facsimle> ... </facsimle>`
 `   <facsimle> ... </facsimle>`
`</facsimle>`

Of course, this is not valid TEI; `<facsimile>` elements are not supposed to nest. However, nesting indeed represents the relation "I am a part of" as Magdalena pointed out. Nesting is a primary structural device in XML. Hence, `<msPart>` elements are nested as children of `<msDesc>`, they do not follow as its siblings. 

In the last mail to the discussion on the TEI-L, I tried to explain why `<surfaceGrp>` is not a good solution  to the problem. 

I propose to reconsider allowing the `<facsimile>` as a possibly nested element. 
Related project: [Register of Older Slovenian Manuscripts](https://rrss.manuscripta.zrc-sazu.si/).

martinascholger:
This PR addresses #2565


tuurma (approved):

martindholmes:
After some discussion with @martinascholger, we decided that this approach is too loose, and allows lots of weird content mixes (facsimiles interleaved with surfaces, for instance) whose meaning is hard to fathom. The feature request is just for the ability to create a facsimile containing child facsimiles, so perhaps the content model should just be an alternate between the existing content model and one or more facsimiles.

tuurma:
@martinascholger @martindholmes I agree with the proposed alternate - either nested facsimiles or surfaces etc

sydb:
Just to be sure I understand, y’all are proposing 
~~~
facsimile = ( model.graphicLike | surfaceGrp | surface )+ | facsimile+
~~~
as opposed to the current definition (on this ms_2565_facsimile branch) of
~~~
facsimile = ( model.graphicLike | surfaceGrp | surface | facsimile )+
~~~
Right? If that’s the case, seems like a good idea to me, but I would want to hear from the likes of @MatijaOgrin, @jamescummings, or @schassan before merging into dev.

sydb:
(Just to be clear, the actual content model being discussed allows an optional `<front>` before, and an optional `<back>` after, all the other stuff. Thus what I really should have said [above](#issuecomment-2594335481) is that the intervention suggestion is to use
~~~
facsimile = ( front?, ( model.graphicLike | surfaceGrp | surface )+, back? ) | facsimile+
~~~
instead of
~~~
facsimile = front?, ( model.graphicLike | surfaceGrp | surface | facsimile )+, back?
~~~
which I still think is a good idea.) Given that we have gotten a thumbs-up from @martinascholger, @martindholmes, @tuurma, myself, and @MatijaOgrin,¹ I am inclined to just make the change and then ask @jamescummings for a final review.

Although I do wonder if there is any need for discussion or examples in the _Guidelines_.

**Note**
¹ Presuming the thumbs-up emoji was approval of the idea, not just approval of getting feedback, first.

martindholmes (commented):
Looks good to me. I agree that an example tying this to the parallel with msPart would be helpful.

martindholmes (approved):

sydb:
Egads! @martindholmes, my apologies! You just reviewed it without my having pushed the latest change. Pushed now, which is a bit premature, as we know it will fail to build until #2680 is merged (presumably into dev and then to this branch).

sydb:
I am not going to review this PR, both because I have actually committed changes that are being considered, but also because I still do not understand why this is a useful feature to have. (I have read the entire email referred to from the issue this addresses, and did not see anything that convinced me this was a helpful approach to encoding multiple facsimiles of manuscript parts. But then again, I do not deal with manuscript descriptions every day, so maybe I am missing something.)

ebeshero:
I have reviewed this, and built the Guidelines tei_all.rng and tested it to validate the nested `<facsimile>` as well as other options and it all works as written. So I am happy to merge this and for it to be part of the Guidelines as a new feature. But I would definitely prefer to have an example and some explanation of when it's appropriate and useful to self-nest the `<facsimile>` element in the Guidelines.

PR #2657, opened 2025-01-17
Popup fix 

### LINKED ISSUE #2656: Footnote and bibliography popups  show back to text links
Back to text links are not useful in a popup; moreover more than one can be shown at a time, which looks really messy. Examples:

For example in `att.canonical`:

Image

Another example in https://tei-c.org/release/doc/tei-p5-doc/en/html/SA.html#index-egXML-d54e129400

Image

raffazizzi:
Removed links back to text in footnote and bibliography popups. Also removed dependency on jQuery for this script.

This should fix #2656


sydb:
I am unable to review this well because I do not really speak Javascript. But I generated the _Guidelines_ in this branch, put them on a server,¹ and took a look. Indeed, the problem described in the issue is resolved. Seems to accentuate the problem of the pop-up appearing quite a distance away from the link, but that is not (I don’t think) a problem for this ticket. 
The result here is, I think, far better than what you get from dev, so I am in favor of merging this ASAR.

In case you want to see a comparison without generating yourself:

- [pre (i.e., dev branch) version of att.canonical](https://bauman.zapto.org/~syd/temp/4TEICouncil/issue_2656/Guidelines-web_dev/en/html/ref-att.canonical.html)
- [post (i.e., this branch) version of att.canonical](https://bauman.zapto.org/~syd/temp/4TEICouncil/issue_2656/Guidelines-web_2656/en/html/ref-att.canonical.html)

**Note**
¹ Because you never see the pop-ups if you just open the local file.

martindholmes:
Agreed. The fix works nicely, but we should raise another ticket for the popup location. I think this was all written back when screens were small and nothing was very far away from anything else. Easy to locate it to the left or right of the click event location, depending on the space available.

raffazizzi:
@martindholmes and @sydb, I made a few adjustments to the JS to make it more idiomatic and I improved the positioning of the popups.

More importantly, I _think_ I removed the dependency on bootstrap and jquery. I couldn't locate any place in the code that seems to use them. I could use some help double checking that I didn't break anything.

 I think the Stylesheets still use jQuery somewhere but also load their own code. JQuery is no longer necessary in 2025; as browser native APIs have caught up and are generally faster and better. It would be good to get rid of that dependency entirely, eventually.

martindholmes:
@raffazizzi @sydb When I wrote that popup thing, it definitely had no dependency on Bootstrap (which I don't think existed at the time) or JQuery (which I've always hated). I think Sebastian put JQuery in there from before that because it was used for some listing or table-of-contents functionality, but anything it would have been doing back then would be trivial to do in CSS nowadays. 

@joeytakeda and I were talking today about the need to make the Guidelines more phone-friendly, especially those tagdocs tables at the end; in some work on the staticSearch documentation (which is generated from the ODD file with the Stylesheets) I've been working on that, and it is doable.

ebeshero (approved):
This looks great! @sydb and I reviewed it this evening and determined that we should probably reconsider the position of the popup, but that we can save for a later ticket!

PR #2660, opened 2025-01-18
use `<constraintDecl>` in Guidelines and tei_customization, and add CMC to French Guidelines

sydb:
Three changes:
 * Use `<constraintDecl>` for Schematron namespace declarations in tei_customization
 * Use `<constraintDecl>` (instead of hack `<constraintSpecs>`) in the _Guidelines_ (both en & fr)
 * Added CMC to fr _Guidelines_ (kinda embarrassing that it wasn’t there!)
 
 **WARNING:**
 This PR will **not** work against the released or current dev Stylesheets, but rather against the sydb_fix_constraintDecl_oversight_01 branch of the Stylesheets; and thus it should **not** be merged until (TEIC/Stylesheets#727) is merged!

### DIFF
--- a/P5/Source/Specs/TEI.xml
+++ b/P5/Source/Specs/TEI.xml
@@ -42,24 +42,6 @@
       </alternate>
     </sequence>
   </content>
-  <constraintSpec ident="c1" scheme="schematron" xml:lang="en">
-    <constraint>
-      <sch:ns prefix="tei" uri="http://www.tei-c.org/ns/1.0"/>
-      <sch:ns prefix="xs" uri="http://www.w3.org/2001/XMLSchema"/>
-    </constraint>
-  </constraintSpec>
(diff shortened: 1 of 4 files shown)


martindholmes:
Looks good to me, but why do we still need the old Schematron namespace?

sydb:
Good question. Not at all sure that we do. (I just copied the namespace decls that we had.)

martindholmes:
It occurs to me to wonder what the Stylesheets will do with a chain in which some ODDs use the old declare-it-wherever-you-want approach to ns declaration and the new constraintDecl mechanism. @sydb Do you know?

martindholmes:
ATOP group tested this today, and everything works as intended, so we're good to go.

martindholmes (approved):
Good to go.

PR #2662, opened 2025-01-20
Implement @hcayless solution #1 for issue @2246 — do not test `@spanTo` across files

sydb:
Since no one has discussed this further in over 2 years, I just went ahead and changed the Schematron (as proposed on the issue) to only test that the target element follows the current element iff the `@spanTo` starts with ‘#’. I used (what I think) is an easier-to-understand XPath, though.

This is not an ideal solution, of course, because (IMHO) we should allow a user to put whitespace in the attr value and still get the test (i.e., we should use normalize-space() in the code). Also, doesn’t `.#duck` also point to the element with ID "duck"? For that matter, doesn’t `../me.xml#duck` point to it if the current file happens to be called me.xml?


martindholmes (approved):
Good to go as far as I can see.

ebeshero (approved):
I just pulled in this branch, ran tests and built exemplars so I could see the built RelaxNG. I tested that new version against a local TEI file with a `@spanTo` that points into external file rather than to a following anchor, and all works as expected. The new version successfully loosens the constraint so we can do this without a validation error. And the new version of the Schematron *also* works to validate a pointer to a following anchor inside the file. So all's good!

ebeshero:
With two positive reviews and a refridge period beginning, I'm just going ahead and merging this now to dev.

hcayless (approved):
@sydb It's not perfect, but I'll point out that the previous rule also assumed a plain fragment identifier. Approved.

PR #2664, opened 2025-01-21
temporary patch alternative to #2643

sydb:
In case we do not either handle the class jiggering or new class naming that would be needed for #2643, this PR just adds `<div>` to att.placement in an effort to solve OP request of #2551.

Issue #2542, opened 2024-03-22
Allow @place in postscript


cthomasdta:
This was originally discussed with the TEI SIG Correspondence, cf. my initial mail https://listserv.brown.edu/cgi-bin/wa?A2=TEI-CORRESP-SIG;d1a80614.2401. You can read a little more context there, but to keep things very short for the purpose of this issue I will simply quote these few lines: 

> following the proposals in the **Encoding Correspondence Manual** on postscripts, Ch. "Postscripts of Letters", https://encoding-correspondence.bbaw.de/v1/postscripts.html, we would raise an issue with the TEI to allow `@place` in `<postscript>`.
> 
> This is also proposed as one possibility in the chapter:
>> Furthermore, the `<postscript>` element should be supplied with an attribute `@place` to indicate the place where the postscript is entered on the page. 

So, the `<postscript>` element should become a member of [class `att.placement`](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.placement.html).

JanelleJenstad:
This request has my full support. We need `@place` available to fully describe the mise-en-page of many features of the text. I hope Council will consider other candidates for `att.placement` while they are considering adding `<postscript>` to `att.placement`.

cthomasdta:
> hope Council will consider other candidates for `att.placement`

yes, I would argue for `<div>` as well and could provide an example for a distinct `<div type="writingSession">` (not a postscript!) written upside-down at the top of the page. But maybe for now or at least for this issue we stay with the motion for adding `<postscript>` to `att.placement`.

lb42:
Making div into a free floating block of text does some violence to its existing semantics, such as they are, in my view. Why not use seg ?

cthomasdta:
> […] Why not use seg ?

In my view, a question of priorities, what we want to express: `<seg>` is very generic, whereas the most important info imo is that this _is_ a `<div>`, more specifically a `<div type="writingSession" n="3">`, _exactly like_ the two other `<div type="writingSession">` that the letter consists of. 

So we have a clear succession of the same elements for the respective text containers (and not `div`-`div`-*`seg`* -- as if the last one was something else than the two blocks before that):

```XML
<div type="letter">
   <div type="writingSession" n="1"><p>[…]</p></div>
   <div type="writingSession" n="2"><p>[…]</p></div>
   <div type="writingSession" n="3"><p>[…]</p></div>
</div>
```

[...]

Issue #2528, opened 2024-02-14
Add sp to att.placement


JanelleJenstad:
It would be useful to be able to encode the placement of speeches on the dramatic page in cases where a speech is beside another speech or in the marginal area. Please add the `<sp>` element to the `att.placement` attribute class. 

Use Cases
1. _Hamlet_, second quarto from 1604/5. In the Mousetrap scene (i.e., Hamlet's play) there are two speeches that are set to the right of other speeches. The placement is important for editors and for people studying performance because the placement suggests that Hamlet _interrupts_ or speaks over the Player Queen. We see similarly significant speech placements across the corpus of early modern drama and would like to be able to use the `@place` attribute and values from our place taxonomy to capture this significant information (not least because we want to be able to search for speeches that appear beside others):
![image](https://github.com/TEIC/TEI/assets/5943608/4b1bbdba-9ee1-404c-8e2a-7a9994f80126)
(Note that we will likely use the `<spGrp>` element to group these speeches. 

2. _As You Like It_, 1623 Folio. Here's an example where the compositor has crunched in a speech in the marginal space. It's not semantically significant, but if one is interested in compositorial practice and mise-en-page, one would want to be able to indicate placement. 

![image](https://github.com/TEIC/TEI/assets/5943608/ecd8e320-c038-4686-8b76-91ad0e2d8310)

Current encoding is this:
`<sp xml:id="emdHam_Q2_sp533" rendition="rnd:rightAdjacent">
<speaker>Ham.</speaker>
<ab>That’s
                        <lb/>wormwood
</ab>
</sp>`

Desired encoding is this:
`<sp xml:id="emdHam_Q2_sp533" place="plc-rightAdjacent">
<speaker>Ham.</speaker>
<ab>That’s
                        <lb/>wormwood
</ab>
</sp>`

sydb:
:+1:

lb42:
This sort of placement is quite uncommon in 19c material. What we have to contend with far more frequently is  things like this:

![parallel](https://github.com/TEIC/TEI/assets/660724/692aa02a-c45a-4e37-b9e9-a0ae2de8280d)

Would this change help with that? How would you recommend encoding it?

JanelleJenstad:
Congrats on the release, folks! Now that "Six Degrees" is out the door, I'm hoping you'll have a moment to deal with this ticket and linked ticket #2572. It's an easy fix.

JanelleJenstad:
[...]

PR #2665, opened 2025-01-22
added att.placement to att.transcriptional, modified specs for #2542, #2551, #2528, #2572

### LINKED ISSUE #2572: Add speaker to att.placement
Please add `speaker` to `att.placement`. It's common for manuscript and printed plays to put speech prefixes in a variety of places. The editors for the Linked Early Modern Drama Online project are requesting this change. See screenshot below from a single manuscript with speech prefixes in three different places. 

This ticket is similar to #2528 and could be dealt with at the same time.

[image](https://github.com/TEIC/TEI/assets/5943608/ce5f4eca-c76b-4ba7-9f1a-e3ba65d50b54)

ebeshero:
This PR may resolve 4 tickets calling for expanded use of `@place` on various elements. It currently addresses 2 tickets: #2542 and #2551. If this works, I'm also going to see if we can apply it to resolve #2572 and #2528. If everything all checks out, this may supersede these PRs:
* PR #2643 
* PR #2664

The method for applying att.placement involves adding it to att.transcriptional, which also allows these elements access to related attributes designed for related purposes: These attributes are all "specific to elements encoding authorial or scribal intervention in a text when transcribing manuscript or similar sources."

### DIFF
--- a/P5/Exemplars/tei_tite.odd
+++ b/P5/Exemplars/tei_tite.odd
@@ -1045,8 +1045,8 @@ some say, "No."
       <equiv filter="tite-acdc.xsl" mimeType="text/xsl" name="b"/>
       <desc  xml:lang="en" versionDate="2014-01-12">(bold) for capturing typographical feature: bold glyphs.</desc>
       <classes>
-       <memberOf key="model.hiLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.hiLike"/>
       </classes>
       <content>
        <macroRef key="macro.paraContent"/>
@@ -1056,8 +1056,8 @@ some say, "No."
       <equiv filter="tite-acdc.xsl" mimeType="text/xsl" name="i"/>
       <desc  xml:lang="en" versionDate="2014-01-12">(italics) for capturing typographical feature: italicized glyphs.</desc>
       <classes>
-       <memberOf key="model.hiLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.hiLike"/>
       </classes>
       <content>
        <macroRef key="macro.paraContent"/>
@@ -1067,8 +1067,8 @@ some say, "No."
       <equiv filter="tite-acdc.xsl" mimeType="text/xsl" name="ul"/>
       <desc  xml:lang="en" versionDate="2014-01-12">(underline) for capturing typographical feature: underlined glyphs.</desc>
       <classes>
-       <memberOf key="model.hiLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.hiLike"/>
       </classes>
       <content>
        <macroRef key="macro.paraContent"/>
@@ -1078,8 +1078,8 @@ some say, "No."
       <equiv filter="tite-acdc.xsl" mimeType="text/xsl" name="sub"/>
       <desc  xml:lang="en" versionDate="2014-01-12">(subscript) for capturing typographical feature: subscript glyphs.</desc>
       <classes>
-       <memberOf key="model.hiLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.hiLike"/>
       </classes>
       <content>
        <macroRef key="macro.paraContent"/>
@@ -1089,8 +1089,8 @@ some say, "No."
       <equiv filter="tite-acdc.xsl" mimeType="text/xsl" name="sup"/>
       <desc  xml:lang="en" versionDate="2014-01-12">(superscript) for capturing typographical feature: superscript glyphs.</desc>
       <classes>
-       <memberOf key="model.hiLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.hiLike"/>
       </classes>
       <content>
        <macroRef key="macro.paraContent"/>
@@ -1100,8 +1100,8 @@ some say, "No."
       <equiv filter="tite-acdc.xsl" mimeType="text/xsl" name="smcap"/>
       <desc  xml:lang="en" versionDate="2014-01-12">(smallcaps) for capturing typographical feature: glyphs in small capitals.</desc>
       <classes>
-       <memberOf key="model.hiLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.hiLike"/>
       </classes>
       <content>
        <macroRef key="macro.paraContent"/>
@@ -1114,8 +1114,8 @@ some say, "No."
       <desc  xml:lang="en" versionDate="2014-12-11">with the <att>cols</att> attribute is used to mark
        where a document changes columnar layout.</desc>
       <classes>
-       <memberOf key="model.milestoneLike"/>
        <memberOf key="att.global"/>
+       <memberOf key="model.milestoneLike"/>
       </classes>
       <content>
 	<empty/>
@@ -1145,9 +1145,9 @@ some say, "No."
         <att>rend</att> attribute and leave the element empy; if the ornament can be represented
        with characters, include these in the element.</desc>
       <classes>
+       <memberOf key="att.global"/>
        <memberOf key="model.inter"/>
        <memberOf key="model.titlepagePart"/>
-       <memberOf key="att.global"/>
       </classes>
       <content>
        <textNode/>
--- a/P5/Source/Specs/add.xml
+++ b/P5/Source/Specs/add.xml
@@ -32,7 +32,6 @@
     <memberOf key="att.global"/>
     <memberOf key="att.cmc"/>
     <memberOf key="att.dimensions"/>
-    <memberOf key="att.placement"/>
     <memberOf key="att.transcriptional"/>
     <memberOf key="att.typed"/>
     <memberOf key="model.linePart"/>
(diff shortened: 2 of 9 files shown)


ebeshero:
To prepare this, I reviewed which elements were in att.placement, and which were in att.transcriptional to see where we had overlap. As you see from the chart, there were only three elements that needed to be modified in order to add `att.placement` to `att.transcriptional`, and I simply removed `att.placement` from their specs to avoid attribute duplication. Then, I added the requested elements from our tickets to att.transcriptional so they have a good set of transcription-related attributes now.

| members of att.placement    | members of att.transcriptional |
| -------- | ------- |
| ./add.xml  |  ./add.xml |   
./addSpan.xml | ./addSpan.xml  |       
|     | ./del.xml    |
|     | ./delSpan.xml |
| ./figure.xml | |   
| ./fw.xml  | | 
|     | ./handShift.xml |  
 | ./head.xml | |   
| ./label.xml  | |  
| ./metamark.xml | |
| |   ./mod.xml |
| ./notatedMusic.xml | |  
| ./note.xml  | |  
| ./noteGrp.xml | |
| |   ./redo.xml | 
| | ./restore.xml |
| | ./retrace.xml |
| ./rt.xml | ./rt.xml |
| ./stage.xml | |
| |   ./subst.xml |
| |   ./substJoin.xml | 
| ./trailer.xml | |
| | ./undo.xml  |  
| ./witDetail.xml | |

ebeshero:
After some intensive discussion with @sydb , I've "walked this back" a little, so that we are simply doing the following in this PR:

* This adds `att.placement` to `att.transcriptional` (where it may be of use for any element subscribing to `att. transcriptional`). 
* However, we are only adding `att.written` and `att.placement` to the elements discussed in the tickets #2542, #2551, #2528, and #2572, in keeping with Council's discussion in fall 2024 meetings and Buenos Aires F2F, where we recognized the need to associate these attributes meaningfully. We are associating these attributes together a little differently by giving them both membership in att.transcriptional, but also recognizing that not all attributes in att.transcriptional should necessarily apply to these elements.

sydb (changes requested):
`@place` is now available on the following (new ones in **bold**):
* `<add>`,
* `<addSpan>`,
* **`<del>`**,
* **`<delSpan>`**,
* **`<div>`**,
* `<figure>`,
* `<fw>`,
* `<head>`,
* `<label>`,
* `<metamark>`,
* **`<mod>`**,
* `<notatedMusic>`,
* `<note>`,
* `<noteGrp>`,
* **`<postscript>`**,
* **`<redo>`**,
* **`<restore>`**,
* **`<retrace>`**,
* `<rt>`,
* **`<sp>`**,
* **`<speaker>`**,
* `<stage>`,
* **`<subst>`**,
* **`<substJoin>`**,
* `<trailer>`,
* **`<undo>`**, and
* `<witDetail>`.

It certainly makes sense to have `@place` on `<div>`, `<postscript>`, `<sp>`, and `<speaker>` — I am all in favor, and I think those were the original requests, at that.
But I am suspicious about having `@place` on `<del>`, `<mod>` , `<redo>`, `<restore>`, and `<subst>` (and for that matter, `<addSpan>`, but that is not part of this ticket) — I am not sure it makes sense. Has anyone asked for these?
And I am pretty convinced it makes no sense to have `@place` on `<delSpan>`, `<retrace>`,  `<substJoin>`, and `<undo>`. (Three of those are empty, and the fourth is placed, by definition, wherever its parent element is placed, as it _must_ be in the same place as it is a retracing of the text that was already there!)

This ticket adds `@hand` to `<dateline>`, `<emph>`, `<sp>`, and `<speaker>`. None of this seems problematic. (I do not quite get `<emph>`, but it does not matter. It is written stuff, so it _could_ be written in a different hand, and `@hand` is an easier way to say that than `<handShift>`.)

ebeshero:
@sydb Of all the objections that you raise, this ticket does *not* unilaterally add `@hand` to `<emph>`. This was requested and already added in the now-closed #2550.

ebeshero:
Having worked with manuscript encoding that has quite a lot of deletions and complex modifications written in the margins that had a difficult time of describing the relative positioning of these deletions and modifications in the source, I think it extremely valuable to give these elements `@place`.  Notice that they are members of `att.transcriptional`. 

Note that `<addSpan>`, `<delSpan>`, `<retrace>` `<redo>` (as well as `<undo>`, which you did not list) , `<restore>`, and `<subst>` were already members of `att.transcriptional`. The only change this PR introduces is simply allowing `@place` as an optional encoding for these elements. If your objection is that a few of these elements are empty (like `<delSpan>` and `<addSpan>`, that objection does not make the application of `@place` any less meaningful than the application of `@hand` to these elements. The ability to mark closed elements that indicate a span of added or deleted text allows us to point to anchored targets across hierarchies, and this is widely used in mansucript projects. Just because the element is closed does not mean we do not need attributes on it to describe placement or hand that makes a modification. Adding `@place` to the collection of attributes that designate intervention about the source of a marked intervention should not be problematic and should be an improvement to the encoding we have available. 

Our conversation in Buenos Aires highlighted that we should anticipate calls for expanded capacities for elements. (Keep in mind that this *one PR* addresses (really!) 5 distinct tickets from different members of the TEI community, if we count the request for `@hand` on `<emph>`). If anything, we might find the application of transcriptional attributes like `@place` more controversial *outside* the encoding applied to manuscripts--the ones called for in the tickets that this PR resolves. 


While some the small changes introduced by this PR were not explicitly called for, they appear to be appropriate to be able to indicate the placement of transcribed modifications of all kinds. We do not simply volunteer our efforts to react to requests as they arise. We do try to think ahead to the best solution for those who consult the Guidelines. Adding `@place` (via `att.placement`) to `att.transcriptional` will make this attribute more widely available with the other transcriptional attributes and that should not be a point of minute objection.

sydb:
I cannot tell, @ebeshero, whether your argument is that `@place` _does_ make sense on `<delSpan>`, `<retrace>`, `<substJoin>`, `<undo>`, etc., or that it is irrelevant that it does not make sense on those few elements (because the resulting class system is cleaner, which is more important).

At the moment I am entirely unconvinced that `@place` makes sense on `<retrace>` (to pick the worst offender) and friends.

But if your argument is that having a simpler, clearer class system outweighs the lack of semantics of some attribute-element combinations, and that we have certainly done this sort of thing before, then I see it as a balancing act. My instinct is that we have done this sort of thing before — allowed elements to have attributes that do not make sense in order to placate the class system — and lived to regret it. But you (and others) may see that balance differently. 

So if you and others think that this PR results in a more coherent system and that advantage overrides the disadvantage of attribute semantic dissonance, so be it. My objection is noted.

torstenroeder (approved):
I would agree that a clear class system outweights semantic weirdness in some combinations. The semantics might be redundant e.g. for `<retrace>`, but it does not make them wrong.

torstenroeder (approved):

ebeshero:
@sydb To clarify my point, I am arguing that it most certainly makes semantic sense for `@place` to join `@hand` on `<delSpan>`, `<substJoin>`, `<undo>` and others that mark interventions on a manuscript. Anywhere that we want to indicate the hand of the inscriber, we should also be able to indicate the placement of the inscription in the document.  This was the point of making for a cleaner association of these classes through `att.transcriptional`. 

Yes, to be sure, @torstenroeder , it would be redundant to do so in the case of `<retrace>`, so we could modify that one in flight by removing `att.placement` from it. But for each of the others, we should be able to mark their placement and it would be "handy" (haha) to be able to apply the `@place` attribute.

torstenroeder:
@ebeshero I think I like the model to be consistent, which means some semantic redundancy here and there would be totally acceptable. Who knows when it might come in _handy_. I guess somebody could pull out some manuscript where a retracing has been done on a paper strip glued on the original document. (So please let us allow `@place` on `<retrace>`, or I will find a usecase and request it ;-) )

martindholmes (approved):
I understand @sydb 's position, because once attributes have been added to elements that shouldn't really have them, there's a long deprecation road to eliminate them; and the more attributes a schema allows, the more confusing it is for encoders choosing between them. However, on balance, I think this should be merged for this release, and then a ticket raised to consider the illogical combinations, to see if there are any for which no plausible use-case exists; then we could deprecate those.

joeytakeda:
I agree with @ebeshero 's point that "Just because the element is closed does not mean we do not need attributes on it to describe placement or hand that makes a modification." In my mind, whatever is on `<del>` makes sense to have on `<delSpan>` (since they should be semantically equivalent but the latter used in cases where the hierarchy disallows); one could of course easily reconstitute a different hierarchy in which `<del>`s are `<delSpan>`s and `<delSpan>`s `<del>`s.

The only one that I also find a bit strange is `substJoin` (which isn't a thing at all arguably, but an instruction/description of an event), but as noted above, it already has `@hand`, so there's already a lot of slippage anyway.

I also just want to flag that this is an ODD breaking change such that a customization which attempts to remove `att.placement` from (I believe) `<add>`, `<addSpan>`, and `<rt>` will no longer have any effect. 
 
But regardless, I agree with Martin -- I vote **approve** and that we should merge and then deprecate carefully afterwards where needed. We should probably also add some better guidance to the Guidelines as to what `@place` means (and what it means in relation to itself in the case of place within nested elements).

Issue #2155, opened 2021-06-09
Standardize on em dashes


JanelleJenstad:
We have been inconsistent in our use of em dashes and double hyphens. @martindholmes and I noted that "There are 89 instances of double-hyphens in the TEI source files, some of which look like they would have to be retained (because they're part of some formal system such as a keyword taxonomy, or they're exemplifying how to create a comment), but most of which could be changed to em dashes. There are 317 instances of actual em dashes, so they are the majority."

With @martindholmes walking me through the process, I made three edits on the dev branch. I will continue this work in my own branch and update the style guide.

sydb:
Not sure I see a need to be using a separate branch, here (except for practice).

JanelleJenstad:
I'm making a branch only because I need the practice.

trishaoconnor:
I'm resuming @JanelleJenstad and @martindholmes's work on this ticket with @sydb's guidance. I have saved the changes directly to dev.  

I have saved changes from double hyphens to em dashes to the following files within `Source/Specs/*xml`:

- [ ] att.cmc.xml
`auch von Beckerath gem. Literatur ? Grüße --</p><signed`

- [ ] att.indentation.xml
`...was unsicher bei der ganzen Sache und bin jetzt damit auch viel glücklicher!--</p>`

- [ ] post.xml
`glücklicher!--</p><signed rend="inline" generatedBy="template"><gap`

- [ ] classDecl.xml

```
<term>United States -- Politics and government —
<term>Bien et mal -- Enseignement coranique</term>
```

- [ ] hi.xml
`on eût écrit --<q>tu voiras</q>--. </p>`

- [ ] keywords.xml
```
<term>Littérature française -- 20ème siècle -- Histoire et critique</term>
<term>Littérature française -- Histoire et critique -- Théorie, etc.</term>
<term>Français (langue) -- Style -- Bases de données.</term>
```

- [ ] taxonomy.xml
`<catDesc>Littérature européenne -- 16e siècle</catDesc>`

- [ ] textClass.xml
```
<item>Littérature française -- 20ème siècle -- Histoire et critique</item>
<item>Littérature française -- Histoire et critique -- Théorie, etc.</item>
<item>Français (langue) -- Style -- Bases de données.</item>
```

- [ ] view.xml
```
face-- </view>
a brownstone--</view>
<view><name>馬克思</name>隨他女兒到窗邊。<hi>雨水</hi> 打在他臉上-- </view>
[...]

PR #2675, opened 2025-02-06
Reverted em-dashes to double hyphens

bleekere:
Following up on #2155 to revert a number of em-dashes back to double hyphens, especially in cases where the double hyphens are part of original text in `<egXML>`. Note that this PR does not cover *all* cases marked in #2155; we are still awaiting some responses.


martindholmes:
I don't understand why most of these changes are being made -- surely most of these cases SHOULD be em dashes?

PR #2677, opened 2025-02-10
Change whitespace in heading, fixes #2669.

### LINKED ISSUE #2669: Ugly links on examples pages
There's a slight ugliness on the pages where all examples for an element are collected together. The link to the element spec page itself, in the title of the page, extends to include the space on either side of the element as well as the colon preceding it.

sydb:
All I was doing was trying to find where the problem was, just to ascertain whether @martindholmes’ ticket #2669 should be moved to Stylesheets or assigned to someone here. It took _forever_ to find it, so once I found it I just fixed it so we would not have to go through search again.

### DIFF
--- a/P5/Utilities/guidelines.xsl.model
+++ b/P5/Utilities/guidelines.xsl.model
@@ -218,12 +218,15 @@
             <div class="main-content">
               <xsl:call-template name="startDivHook"/>
               <h3>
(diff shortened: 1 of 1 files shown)


martindholmes:
Tested and working -- should be good to merge. Thanks @sydb.

PR #2678, opened 2025-02-11
remove bold emphasis

rettinghaus:
This removes the `rend="bold"` from emphases in "Languages and Character Sets" of the Guidelines. In general it is considered bad practice to have single bold words in running text, also the double-emphasis (bold + italic) looks weird.

There are a few more cases in "Computer-mediated Communication", but used in list items.


lb42 (commented):
Actually shouldnt these emph elements be better as term elements?

rettinghaus:
@lb42 I agree for "abstract character" and "glyph".

sydb:
Encoding (with `<term>` as per @lb42) looks reasonable to me. Passed all tests on my local system. I looked at the affected parts, they look fine.

PR #2679, opened 2025-02-11
Remove dead links

### LINKED ISSUE #2672: Remove/replace broken links to "using-the-tei-github-repository"
[One down](https://github.com/TEIC/TEI/issues/2614), two more to go.

rettinghaus:
If I'm correct <P5/Source/Guidelines/index.xml> is not only outdated but unused, so I removed it.

closes #2672


martindholmes:
Today Council decided they would prefer to move the index.xml file to the Sources/Defunct folder, rather than delete it. @rettinghaus would you like to make that change?

rettinghaus:
done as requested

martindholmes:
Council instructions from March meeting suggest we should close the PR and move the index page to defunct manually, but the link fixes are also required, so I'm going to merge the PR instead.

PR #2680, opened 2025-02-11
remove contextless schematron deprecation

sydb:
Change deprecation warning for contextless `<sch:assert>` or `<sch:report>` to an actual error.  

Note that this is HIGH priority as our “complain about deprecations 60 days before they expire” rule means the _Guidelines_ build is broken.

(Note — I have just realized that a contextless `<sch:assert>` or `<sch:report>` could sneak in via `<constraintDecl>`. But that is a problem for a different ticket.)

### DIFF
--- a/P5/Source/Specs/constraintDecl.xml
+++ b/P5/Source/Specs/constraintDecl.xml
@@ -18,6 +18,8 @@
       <anyElement minOccurs="0" maxOccurs="unbounded"/>             <!-- typically <sch:ns> elements -->
     </sequence>
   </content>
+  <!-- NOTE: Also see constraint "context-required" in file constraintSpec.xml,
+       which may be fired on a <constraintDecl> as well. -->
   <constraintSpec scheme="schematron" ident="one-constraintDecl-per-scheme" xml:lang="en">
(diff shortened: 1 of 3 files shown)


martindholmes:
This looks fine to me. (Replying to Slack question:) I think `@role="warning"` on line 79 is fine because there will be a lot of old ODDs that need to be updated, but also because omitting `@context` won't necessarily immediately break anything; the RNG should still work, and if the Stylesheets have not been updated to ignore cases with no context (which I don't think they have), then the Schematron should still work in most cases.

martindholmes (approved):

sydb:
Do you (@martindholmes) think it is OK to remove the `role="warning"`?

martindholmes:
@sydb I think it would be kinder to keep it, at least for a while; this will affect a great many ODD files, don't you think? See my comment above.

sydb:
Well, yes, but people have been getting the deprecation warning for a year now. (This deprecation was committed in December 2023, so users have gotten it since the release immediately after that.)

martindholmes:
OK, in that case feel free to remove it and make it an error. I suppose a year should be long enough.

PR #2682, opened 2025-02-12
fix `@ident` of `<remarks>`, at least partially

sydb:
Changed the definition of `@ident` of `<remarks>` (from teidata.text to teidata.name); and changed its description (from “specifies the remark concerned.” to “supplies the identifier by which the remarks may be referenced.”).


martindholmes:
@sydb On line 32/24, the desc seems to have been changed, but the `@versionDate` attribute still shows 2020. Shouldn't it show the current date?

raffazizzi (approved):
Update versionDate as requested by @martindholmes

PR #2692, opened 2025-03-28
Move note content handling into the event

amclark42:
Rather than trying to determine which note goes with which anchor at page load, wait until a click has occurred and use the link's `@href` to determine the note.

This fixes a bug where **all** links' popups showed the last note on the page. For example, clicking [anchor 96 in chapter 23](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/TD.html#Note131_return) currently shows note 101. (And the same is true for anchors 97 through 101.)

Screenshot 2025-03-28 at 10 00 41 AM

On page load, `var dest` was set iteratively, concluding with the last anchor link. Because functions aren't run when they are declared, when the anonymous event ran and set `showPopupFootnote()` in motion, it would *always* set the second argument to the current value of `dest` — the last-processed note.

I have not tested this fix, but I'm 90% certain it should work.


amclark42:
Syd (@sydb) has informed me that, while this passes QA tests, the fix does *not* work. I think I know why: event bubbling means that the "thing the user clicked on" may have actually been the `<sup>` and not its container `<a>`, meaning there is no `@href` to get when you click what appears to be the link. Will try a fix shortly!

sydb (approved):
I tried building in the Docker:
- Passes tests.
- Loaded 3 chapters, and in each case it worked for me.
Thus approval.

However, I am not qualified to actually review the code itself. So if someone Javascript-savvy could take a look, that would be good.

P5/webnav/popupFootnotes.js
```diff
@@ -30,13 +30,24 @@ document.addEventListener("DOMContentLoaded", function() {
   for (var i=0; i<links.length; i++){
     if (links[i].hasAttribute("href") && links[i].getAttribute('href').substring(0, 5) == '#Note'){
       if (links[i].getAttribute('class') != 'link_return'){
-        var dest = links[i].getAttribute('href').substring(1);
+        /* When an anchor link is clicked, we use its @href value to determine which 
+         note's content should be loaded into the popup. */
         links[i].onclick = function (e) {
```
joeytakeda: I know this is from the original, but if we're changing this, then why not move more toward ES6

```js
const links = document.querySelectorAll('a[href ^= "#Note"]:not('.link_return')');
links.forEach(link => {
  link.addEventListener('click', (e) => {
     e.preventDefault();
   // And then the rest as @amclark42 proposed
  })
}
```

joeytakeda (changes requested):
Given that we're updating, I think it's worth taking the opportunity to build on @amclark42 's work and update the surrounding bits of code too.

amclark42:
@joeytakeda I think updating the code is a good idea. I went for a minimum of changes because I had no way to test my changes against the site — the more dramatic my changes, the less confident I would have been in the fix. Syd helpfully tested the code on his local setup.

I've now downloaded a copy of [Chapter 23](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/TD.html)'s HTML, and hooked it up to assets in my copy of the Guidelines repo. So I *can* make more substantial changes, because I can test them.

But rereading your comment, it sounds like someone else is being asked to make the change. Can you confirm who is making the requested changes?

joeytakeda:
Hi @amclark42 ! Thanks very much for this and apologies for the vagueness —I hadn’t meant anyone specifically, so if you are so inclined, I’d say please feel free to make the changes.

amclark42:
@joeytakeda Awesome, thank you for clarifying!

amclark42:
@joeytakeda @sydb I've just pushed up new commits which update and streamline most of the JS. I did not update `clearContent()` or the bibliography-related functions, because...

I noticed that the footnote popups were doing a disservice to people using screen readers. If you tabbed to a link and hit enter, a popup would appear (good), but the popup would not be immediately in focus (bad). So anyone using a screen reader would have to read through the entire narrative (including the footnotes) before they noticed the popup and its content.

I've made tweaks such that when a link is clicked, focus goes to the close button (yes, it's now actually a button, with an ARIA label to boot). When the close button is clicked or the escape key is pressed, focus returns to the link which triggered the popup.

This is not a robust solution, accessibility-wise — ideally, we'd indicate that a link opens a popup, and constrain interaction to the bounds of the popup while it's open — but I think it's an improvement.

I think I will leave the rest of the JS to the Council.

joeytakeda:
Thanks very much for this @amclark42 . I've tested this now and it looks great. 

Agreed that there is still more to do here (ideally, I think that the Guidelines would use the [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element, which has [better handling for focus](https://www.scottohara.me/blog/2023/01/26/use-the-dialog-element.html) out of the box), but I think this should be merged in as both a necessary fix and great improvement to what we had before.

joeytakeda (approved):
As noted, all looks great and should be merged

sydb:
I agree with @joeytakeda , this is much better and should be merged.

That said, in my test in both Firefox (136.0.3) and Chromium (134.0.6998.165) on Ubuntu (24.04.2) the correct note popped up, but did not seem (to me) to be in focus. Pressing ENTER did not close the pop-up; pressing TAB moved the selected item from the footnote anchor to the next link in the body of the text. Pressing ESC (even after moving selection with TAB) did close the pop-up.

@amclark42 — let me know if you want to see this in a screen-share. (But if you want to step away having fixed an egregious bug, feel free.)

amclark42:
> That said, in my test in both Firefox (136.0.3) and Chromium (134.0.6998.165) on Ubuntu (24.04.2) the correct note popped up, but did not seem (to me) to be in focus. Pressing ENTER did not close the pop-up; pressing TAB moved the selected item from the footnote anchor to the next link in the body of the text. Pressing ESC (even after moving selection with TAB) did close the pop-up.
> 
> @amclark42 — let me know if you want to see this in a screen-share. (But if you want to step away having fixed an egregious bug, feel free.)

Ah, @sydb, try hard-refreshing and let me know if that fixes it.

sydb:
@amclark42 — No, hard refresh did not change anything in Firefox. (I did not bother hard refresh in Chromium, as I had launched it for the sole purpose of testing this. :-)

@amclark42, @joeytakeda, @martindholmes, and anyone else interested — I have also noticed that if the footnote anchor is near the top of my web browser’s window, the top of the pop-up (and thus the close button) is out of the window. (The same problem does not hold true when the anchor is near the bottom of the window. If the anchor is actually not visible off either the top or bottom of the window, then the pop-up is not visible, either; not too surprising, I guess.)
This is true both using this fix-popup-note-js branch and as currently published on the web. Let me know if it deserves another ticket of its own.

martindholmes:
@sydb In code I've written before to handle popups which need to be near the click location but stay onscreen, I've had to calculate on the basis of the popup size whether it's best placed above or below the click point, and to its left or right, in order to be visible on the page. I think with the original popup code from years ago, I just settled on a fixed position for the popup, because it seemed less disruptive -- the reader always knows where it's going to be, and it's less likely to obscure any of the text they're looking at.

sydb:
@amclark42 figured out that I was inadvertently using an older version of er code. (For some reason I had _two_ remotes pointing to the amchark42 fork.) Once proper code was pulled, focus is in the proper place. Yay!

Still think this can be merged. (Yes, there are other improvements to be made, but no reason not to fix this egregious bug right away, with thanks to @amclark42!)

Issue #2683, opened 2025-02-15
Encoding of ordered lists


rettinghaus:
There seems to be an inconsitency with the encoding of ordered lists. Most are using `@rend="numbered"`, a few `@rend="ordered"`.

What I get from this [dated discussion](https://github.com/TEIC/TEI/issues/1099) it should be ordered, but couldn't figure out, what the preferred encoding within the Guidelines should be nowadays. 

The Stylesheets use `list type="ordered"` for converting to TEI.

hcayless:
My own contention is that all lists are ordered and cannot be otherwise, so to label them as such is silly. But opinions may vary.

lb42:
Hughs  contention would be more persuasive if tei offered any way of explicitly encoding unordered aggregates of items. But it doesnt....

hcayless:
An item in TEI is defined as a component of a list, so that might be a reason why it doesn't. If you need one, we could create an itemGrp, I suppose. "Unordered" as in HTML ul tags or old-style TEI list `@type`s means "not numbered". They still have an order. They can't help it.

jamescummings:
My recollection is that our previous discussion concluded that it wasn't really about whether the items were **in** an order (as Hugh says they obviously are) but whether that order **mattered** to the encoder or whether they only put them in that order because one must to encode them as a list.

lb42:
So long As rhe guidelinese are expressed in something like xml, its obviously true that the order of any series of child elements is determined. We do however already provide an @org attribute  for encoders wishing to override this behaviour. See its definitions on att.divLike and vColl . Maybe adding list to the att.divLike class would  be a good idea?

cthomasdta:
Quick question to understand the direction of the discussion, Klaus @rettinghaus 

> There seems to be an inconsitency with the encoding of ordered lists. Most are using `@rend="numbered"`, a few `@rend="ordered"`.

Does your initial observation include an objection against using `@rend="numbered"`, e.g. for lists that are explicitly numbered (instead of e.g. bulleted) in the manuscript transcribed? 

Or are you and is everybody else fine with (most) people using `@rend="numbered"` (and `@rend="bulleted"` etc.), and are only wondering if the value `"ordered"` is reasonable and makes sense for `<list>`s?

rettinghaus:
I'm a little surprised at the direction the discussion on this simple question has taken. I apologise if I have not made myself clear. 
[...]

PR #2693, opened 2025-03-29
improve encoding of `<list>`s

sydb:
Per #2683 I have taken the first (and by count largest) step in harmonizing all `list/(@type|@rend)`s. Thus the current counts are

| count | `@type` | `@rend` |
| ----- | ------- | ------- |
|   106 |  |  bulleted |
|    59 |gloss  |  |
|    42 |  |  |
|    39 |  |  numbered |
|     7 |  |  simple |
|     1 |  |  specList |

What I did:
* type=simple → rend=simple
* type=ordered → rend=numbered
* rend=ordered → rend=numbered

I have thoroughly convinced myself there is no untowared effect of these changes, at least to the HTML. (I did not check PDF output.)

### how I convinced myself

1. Checked out dev branch into two different directories (called A and B here).
2. Built Guidelines-web/ in each directory (using `make clean html-web`) and checked for errors (there were none). 
3. Developed a comparison command for later use (see below).
4. In directory B made the changes noted above.
5. In directory B re-built Guidelines-web/ and checked for errors (there were none).
6. Issued somewhat complex comparison command explained below. It demonstrated that (not surprisingly) AB, TC, and TD were the only files effected. It further told me exactly where there were changes in those files.
7. For each of those 3 chapters I loaded the A version and the B version in tabs next to each other; in each tab I found the spot(s) that was changed; I then changed rapidly back and forth between the two tabs, looking for any changes (which would appear as movement). There wasn’t any in any of the ~½ dozen cases.

### below
The comparison cmd I used was:
~~~bash
$ for f in B/P5/Guidelines-web/en/html/*.html ; do b=$(basename $f) ; echo "---------$b:" ; n="A/P5/Guidelines-web/en/html/${b}" ; diff -C0 -bBw <(egrep -v 'March 2025' $f | perl -pe 's,index-[a-zA-Z]+-[0-9a-f]+",PROTECTED",g;') <(egrep -v 'March 2025' $n | perl -pe 's,index-[a-zA-Z]+-[0-9a-f]+",PROTECTED",g;') ; done
~~~
That deserves some explanation:
* For each HTML file in the B copy (called `$f` or `${f}`),
* generate its base name (i.e., strip the path off), call the result `$b` (or `${b}` for “base”),
* generate the name of the corresponding file in the A copy (called `$n` or `${n}` for “new”),
* issue a `diff` command (with my favorite `-C0 -bBw` switches) that uses 2 process substitutions as input files.
* A _process substitution_ (in this case) takes the form `<( CMD with args)`. The CMD is executed (with args), and the result is written to a temporary file, and the name of that temporary file is returned, and thus passed to `diff` as an argument.
* These particular process substitutions read in the aforementioned `$f` and `$n` files, and execute:
  * `egrep -v 'March 2025'` — This means “search for all lines in the input that contain `March 2025` and return all the other ones”. I had noticed back in step 3 that all of the lines that are different because of the difference in the timestamp all had “March 2025” in them.
  * **`|`** the standard Unix pipe operator — take the output of cmd on L and feed it as input to the cmd on R
  * `perl -pe 'PERLESE'` — execute the Perl command PERLESE on each line of input and return whatever Perl returns.
  * In this case PERLESE is `s,index-[a-zA-Z]+-[0-9a-f]+",PROTECTED",g;` which is a search-and-replace command that finds strings matching the regular expression `index-[a-zA-Z]+-[0-9a-f]+"` and replaces each with the string `PROTECTED"`. I noticed back in step 3 that all of the changes to ID (and IDREF) values matched this pattern.


sydb:
See [comment on issue](https://github.com/TEIC/TEI/issues/2683#issuecomment-2764004490) for discussion of "specList" — but it boils down to there is no need to change it.

hcayless (approved):
Looks good to me!

rettinghaus (approved):

rettinghaus:
Thank you!

Issue #2642, opened 2024-12-19
Introduce a way to set stage directions prior to speech prefixes


jjokisch:
I have noticed repeatedly that stage directions do not just form structures with speech prefixes by following them but quite frequently precede them or simply have one word in the stage direction appear typographically marked as the speech prefix. There are countless examples really:

John Home's [_Douglas_](https://archive.org/details/case_y_134_188_vol_15/page/n41/mode/1up) (1757)
![grafik](https://github.com/user-attachments/assets/1d762cea-e1d8-4885-94fe-b80b28fdf10a)

Jean-Baptiste Legouvé's [_Attilie_](https://gallica.bnf.fr/ark:/12148/bpt6k6113280c/f81.item) (1775)
![grafik](https://github.com/user-attachments/assets/41ccee8d-f379-4ac3-9eba-503786f2ea54)

The problem with that being that `<speaker>` can only occur as the first child of `<sp>` and neither can `<speaker>` be nested in `<stage>` nor `<stage>` in `<speaker>`. Thus, while I can encode, for example, "GLENALVON solus" as `<sp><speaker>GLENALVON</speaker><stage>solus</stage>[...]</sp>` a construction like "Manet GLENALVON" or even "Manet GLENALVON solus" only allows for rather questionable encodings, i.e., (1) by disassembling the entire structure:
```
<stage>Manet</stage>
<sp>
<speaker>GLENALVON</speaker><stage>solus</stage>
</sp>
```
or (2) by giving up on the nuance between speech prefix and stage direction:
```
<sp>
<stage>Manet <persName>GLENALVON</persName> solus</stage>
</sp>
```

@DanilSko actually had made me aware that the DraCor group had also run into this issue before: https://github.com/dracor-org/udracor/issues/11

(2) seems the less favorable as it opens the door to all kinds of terminologically unsound arguments. After all, if "Manet GLENALVON solus" is merely a stage direction, why wouldn't "GLENALVON" not be one as well? And if it were a stage direction with a `<persName>` in it, why would we ever use `<speaker>` for a speech prefix and not consider all speech prefixes merely as names? 

[...]

PR #2698, opened 2025-04-10
Fix for issue #2642.

martindholmes:
I believe this addresses the issue, allowing members of `model.stageLike` in alternation with the existing `model.global` before the optional `<speaker>` element. I've also added an instance into one of the test files.

### DIFF
--- a/P5/Source/Specs/sp.xml
+++ b/P5/Source/Specs/sp.xml
@@ -32,11 +32,12 @@
   </classes>
   <content>
     <sequence>
-      <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
-      <sequence minOccurs="0">
-        <elementRef key="speaker"/>
+      <alternate minOccurs="0" maxOccurs="unbounded">
         <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
-      </sequence>
+        <classRef key="model.stageLike" minOccurs="0" maxOccurs="unbounded"/>
+      </alternate>
+      <elementRef key="speaker"/>
+      <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
       <sequence minOccurs="1" maxOccurs="unbounded">
         <alternate>
           <elementRef key="lg"/>
--- a/P5/Source/Specs/sp.xml
+++ b/P5/Source/Specs/sp.xml
@@ -32,10 +32,10 @@
   </classes>
   <content>
     <sequence>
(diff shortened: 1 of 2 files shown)


P5/Source/Specs/sp.xml
```diff
@@ -32,8 +32,11 @@
   </classes>
   <content>
     <sequence>
-      <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
-      <sequence minOccurs="0">
+      <sequence minOccurs="0" maxOccurs="unbounded">
```
sydb: This makes the content model
~~~
(
  ( model.global*, model.stageLike* )*,
  ( speaker, model.global* ),
  (
    ( lg | model.lLike | model.pLike | model.listLike | model.stageLike | model.attributable ),
    ( model.global* | q )
  )+
)
~~~

While this is not incorrect (any number of either model.global or model.stageLike is allowed before `<speaker>`), it strikes me as an odd way to do this. The more normal way (I think) would be to use
~~~
  ( model.global | model.stageLike )*,
~~~
as the 3rd line of the above. I.e.
~~~xml
  <content>
    <sequence>
      <alternate minOccurs="0" maxOccurs="unbounded">
        <classRef key="model.global" minOccurs="1" maxOccurs="1"/>
        <classRef key="model.stageLike" minOccurs="1" maxOccurs="1"/>
      </alternate>
      <sequence>
        <elementRef key="speaker"/>
        <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
      </sequence>
      <sequence minOccurs="1" maxOccurs="unbounded">
        <alternate>
          <elementRef key="lg"/>
          <classRef key="model.lLike"/>
          <classRef key="model.pLike"/>
          <classRef key="model.listLike"/>
          <classRef key="model.attributable"/>
        </alternate>
        <alternate>
          <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
          <elementRef key="q"/>
        </alternate>
      </sequence>
    </sequence>
  </content>
~~~

P.S. Can anyone tell me what that `<q>` is doing at the end? I know it has _nothing_ to do with this issue & PR, but it is making my head spin.

sydb (approved):
See detailed comment, which I think is a better content model. But this one works.

P5/Source/Specs/sp.xml
martindholmes: I can't tell you what the q is doing there, but let's not let that get in the way of solving this issue.

PR #2713, opened 2025-05-29
Translation from lampsacus/TEI

eduardo-aleman:
### Regarding Add Spanish translations for TEI Guidelines: AB-About.xml
Any updates on this request?


sydb:
@HelenaSabel and @lampsacus —
This PR breaks the build because there are 2 `<remarks>` in Spanish in abbr.xml.

@lampsacus —
I have taken the liberty of fixing the cosmetic problems introduced into the 1st line(s) of each of the 4 changed files. But my Spanish is not good enough to do much else.

Issue #2715, opened 2025-05-30
clean up: time to ditch I18N/ directory?


sydb:
Working on a recent [PR](https://github.com/TEIC/TEI/pull/2691), we (@bleekere, @trishaoconnor, and I) noticed that it does not seem like anything in the I18N/ directory is still used. We asked @hcayless, who agreed, although not definitively.

I took a look. The string “I18N” appears in only 7 files in the entire P5/ hierarchy:

- 4 of which are release notes (and are not about a file or directory, but about a project)
- 1 of which is a .jar file (as part of HTML DTD)
- 1 of which is build-if-newer.sh (a program that is completely, unusably, out-of-date — it presumes Subversion)

While I can see why we want to keep this stuff around for historical or sentimental purposes, it is in a version control system. So my suggestion is to remove I18N/ and build-if-newer.sh.

Someone might argue that they are not causing any harm, but I disagree. Every time we need to figure out how something works there is the possibility that files these files _might_ be involved, and it takes time figure out they are not. So while I admit they are not causing _much_ harm, I do not see _any_ benefit, so still think they should be nuked.

ebeshero:
Completed with #2716.

PR #2716, opened 2025-06-01
delete some cruft 

sydb:
Per #2715, remove files that (I hope) are unused.

-  build-if-newer.sh, a clever program that has not worked since we left Subversion for git.
-  The entire I18N/ directory, which as far as I can tell is unused.

Note: I think this requires approval of _both_ reviewers before being merged.


hcayless (approved):
I think this is fine, and a good idea. We can always resurrect these if it becomes clear they were actually useful.

martinascholger (commented):
Agreed, I think we don't need this anymore

martinascholger (approved):

PR #2720, opened 2025-06-16
add missing periods in the AI chapter (18.4.2)

bansp:
The prose quotes/mentions two sentences (source and translation) without periods, but it's the entire (source) sentence that gets annotated in the immediately following example, _together with the period_. 

It looked as if the visual editor were going to allow me to push directly to the dev branch, and I panicked a bit and created the PR manually. I hope that's OK.


sydb (approved):
Given that the period appears in the example, this seems like a no-brainer to me.

PR #2722, opened 2025-06-18
Adding missing `<listRef>`s for #2652

trishaoconnor:
This PR addresses the missing `<listRef>`s highlighted in #2652. I have started work on this issue and added `<listRef>`s to the Spec pages of the following individual elements: `<listRef>`, `<paramList>`, and `<outputRendition>`. These were the most straightforward instances which clearly required a `<listRef>` and where there was a specific section in the Guidelines to which we could point.

I have not yet started to address the remaining specifications which also do not have a `<listRef>` (see [table](https://github.com/TEIC/TEI/issues/2652#issue-2782672438)) as these other instances require further discussion at Council, but I hope that this PR could serve to store these future updates as well.

### DIFF
--- a/P5/Source/Specs/listRef.xml
+++ b/P5/Source/Specs/listRef.xml
@@ -69,5 +69,6 @@
   </remarks>
   <listRef>
     <ptr target="#TDcrystalsCEdc"/>
+    <ptr target="#IM-unified">
   </listRef>
 </elementSpec>
--- a/P5/Source/Specs/listRef.xml
+++ b/P5/Source/Specs/listRef.xml
@@ -69,6 +69,6 @@
   </remarks>
   <listRef>
     <ptr target="#TDcrystalsCEdc"/>
-    <ptr target="#IM-unified">
+    <ptr target="#IM-unified"/>
   </listRef>
 </elementSpec>
--- a/P5/Source/Specs/listRef.xml
+++ b/P5/Source/Specs/listRef.xml
@@ -67,4 +67,7 @@
     </list>
     </p>
   </remarks>
+  <listRef>
+    <ptr target="#TDcrystalsCEdc"/>
+  </listRef>
 </elementSpec>
--- a/P5/Source/Specs/outputRendition.xml
+++ b/P5/Source/Specs/outputRendition.xml
@@ -66,7 +66,6 @@
   </remarks>
    
   <listRef>
-    <ptr target="#TDPMPM"/>
     <ptr target="#TDPMOR"/>
   </listRef>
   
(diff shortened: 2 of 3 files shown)


P5/Source/Specs/outputRendition.xml
```diff
@@ -64,10 +64,11 @@
...
    
   </remarks>
-  <!-- 
+   
   <listRef>
-    <ptr target="#WhereWeDiscussIt" type="div2"/>
-    <ptr target="#WhereWeDiscussIt" type="div2"/>
+    <ptr target="#TDPMPM"/>
+    <ptr target="#TDPMOR"/>
+    <ptr target="#TDPMMC"/>
```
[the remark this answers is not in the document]
ebeshero: Scratch that! Just delete the last one pointing to '#TDPMMC'. The alternative target I found was inside the elementSpec for outputRendition so that would be redundant!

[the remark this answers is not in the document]
ebeshero: I see why #TDPMMC didn't turn up in my search: this was a use of `<outputRendition>` inside an `<egXML>`, and because the example is explicitly about how to process this for rendering it in output, I think it's a good reference point. Never mind my objection, and we should keep all three `<ptr>` elements in your `<listRef>` for '<outputRendition>'.

P5/Source/Specs/listRef.xml
```diff
@@ -67,4 +67,7 @@
     </list>
     </p>
   </remarks>
+  <listRef>
+    <ptr target="#TDcrystalsCEdc"/>
```
ebeshero, line 71: To this I would also add:
`<ptr target="#IM-unified">` because a significant mention of `<listRef>`appears there in context with how elements can be repeated without being grouped.

ebeshero (changes requested):
@trishaoconnor I reviewed this after writing an XQuery script to see if we can programmatically locate mentions of our specs that are missing `<listRef>`. Your PR helped me modify my code, so thank you for this work! :-) And I just request one small change to add one more `<ptr>` to the listRef spec.

P5/Source/Specs/listRef.xml
trishaoconnor: I've added `<ptr target="#IM-unified">` to `<listRef>` now. Thank you very much for finding this reference for me.

P5/Source/Specs/outputRendition.xml
trishaoconnor: I've removed `<ptr target="#TDPMMC"/>` as it only has an example in which `<outputRendition>` occurs but no actual discussion of the element that would be useful for a reader

ebeshero (approved):
Approved after review and discussion in Stylesheets meeting today! Thanks @trishaoconnor !

Issue #2723, opened 2025-06-20
Puzzling errors in bib links in published Guidelines


martindholmes:
On this page:

https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.datable.iso.html

If you click on the link to XML Schema Part 2: Datatypes Second Edition, you get an "Undetermined" popup (at least, I do). However, the link itself looks fine (`ref[@target='#XSD2']`), and there are other links to the same target in other documents (such as att.datable.w3c.xml) that work just fine. The HTML output looks OK too, so perhaps it's something to do with the JavaScript that retrieves the data for the popup.

ebeshero:
@martindholmes Aha! I thought this looked familiar, and indeed, it is. Take a look at the bibliography mouseover value vs. the popup window contents as raised in https://github.com/TEIC/TEI/issues/2708 . I think we'll solve these two tickets together.

ebeshero:
@HelenaSabel Do you think we can add this to Stylesheets agenda? Hoping it's a quick fix.

martindholmes:
The footnote code is supposed to first check to see if the thing pointed at is available in the page (as a footnote would be), and if so then show it; otherwise if it's a BIB.html reference, it should retrieve BIB.html and get the thing with the correct id. For some reason that's not happening in some cases, even though the links look identical to contexts where it does work. It's a puzzler.

The JS is a bit more complicated than it needs to be; it creates an iframe and tries to load the BIB document into it. I can see many reasons why that's a bad idea, especially nowadays. I don't remember my original code working that way, but maybe it did. If so, it should be changed. I'll try to dig into it this weekend.

martindholmes:
This JS really does need a rewrite; I think everything can be done much more straightforwardly than it was all those years ago. I'll undertake to do that. It's not really a Stylesheets thing; I think it's just a JS bug.

martindholmes:
Planned approach (so I don't forget it):

onload, fetch the bibliography page and attach it to the body as a shadow dom. Only then replace all the links to bibl items with calls to the popup function, which can now grab its content from the shadow dom. No iframe, and once the page is retrieved once it'll be in the browser cache so everything should be fast.

martindholmes:
[...]

PR #2727, opened 2025-06-29
First pass at issue #2723.

martindholmes:
This is a partial rewrite of the JS that handles turning bibliography items into popups. Instead of the old approach of attaching the BIB.html file as an invisible iframe, which is surely frowned-upon these days, it retrieves the bibl lists from BIB.html and attaches them as to a shadow DOM to an invisible div added to the body, and then pops them up from there. The original links are left unchanged until the shadow DOM has been created, and then instead of using an onclick attribute, the action is attached properly using `addEventListener()`. The `@href` is removed. There's one tweak to the guidelines.css file to ensure they retain their colour, underline, and cursor after the `@href` is gone.

You can test the fix for issue #2723 in a Jenkins build of this branch:
https://teijenkins.hcmc.uvic.ca/job/TEIP5-Documentation-dev-issue-2723/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/ref-att.datable.iso.html
(pointing to the page which is mentioned on the issue).

### DIFF
--- a/P5/webnav/popupFootnotes.js
+++ b/P5/webnav/popupFootnotes.js
@@ -34,7 +34,6 @@ document.addEventListener("DOMContentLoaded", function() {
   // Add event listeners to show popup notes when each link is clicked.
   links.forEach(link => { link.addEventListener('click', showPopupFootnote) });
   
-  //addBiblFrame();
 });
 
 async function getBiblSource(){
--- a/P5/webnav/popupFootnotes.js
+++ b/P5/webnav/popupFootnotes.js
@@ -153,7 +153,6 @@ function showPopupBibl(el, biblId){
 //We have to be cautious here; some browsers block access to the iframe 
 //document contents from another document, especially when running locally.
   try{
-    console.log(biblId);
     bibl = divBiblWithShadow.shadowRoot.getElementById(biblId);
   }
   catch(e){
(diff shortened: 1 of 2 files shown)


P5/webnav/popupFootnotes.js
```diff
@@ -104,39 +132,29 @@ function setupBiblPopups () {
...
-
 //This function shows a bibl popup. It differs slightly from the function for note
 //popups, so it is distinct.
 function showPopupBibl(el, biblId){
   var bibl = null;
 //We have to be cautious here; some browsers block access to the iframe 
 //document contents from another document, especially when running locally.
   try{
-    bibl = biblFrame.contentDocument.getElementById(biblId);
+    console.log(biblId);
```
raffazizzi: Please remove this console.log

raffazizzi:
Besides the stray `console.log` this looks good to me @martindholmes . I wonder if bothering with the shadowdom is strictly necessary here: are you concerned about ID conflicts or mixing CSS from the two HTMLs? If not, it may suffice to just stick the data in the div you're already creating and hiding. Anyway, it certainly doesn't hurt to have it so this questions is more for my own edification.

martindholmes:
Thanks @raffazizzi -- I removed the stray console.log call. I think the shadow dom is the cleanest way to do this, and I don't think it involves any particular overhead or additional complexity. No need to add to the weight of the live dom.

Issue #2653, opened 2025-01-12
`@ed` and `@edRef`?


sydb:
I have this very vague recollection¹ that when `@edRef` was invented the intent was that it would be used as an alternate to `@ed`.  As evidence (besides the fact that it just makes sense), of the 19 examples that include either `@ed` or `@edRef`, none include both `@ed` and `@edRef`. However, `@edRef` is basically not mentioned in the prose of the _Guidelines_ at all,² so it is hard to know. Further, I quite suspect that as soon as I suggest that they should be in strict alternation with one another (i.e., only one or the other should be legal on any given instance of an element that is a member of att.edition, either because we put `org="choice"` on the `<attList>` that contains them or add some Schematron to enforce this alternation) @martindholmes will come up with a use case that demonstrates it is useful to have both.

But in the case where we leave it legal to have both of them on the same element instance, we need to say what it _means_. I submit the answer to that question is either

- a) they indicate separate editions, or
- b) they are different ways of indicating the same edition (in the same order).

Although I suppose we could say “these Guidelines do not assert the intended semantics when both are specified” or some such. (Ick!) Personally, I would prefer the “only one or the other is allowed on any given element” approach.

**notes**
¹  Vague because, I think, I was not on Council at the time and did not pay it much heed.
² Only mentioned once in DR to say “`<lb>` additionally bears the attributes `@ed` and `@edRef`, from its membership …”, not what to _do_ with them at all.

martindholmes:
You're right @sydb, I will. A sigil is not the same as a pointer, and I imagine one might well want to specify a sigil in the context of a specific edition while simultaneously pointing to an external bibliographical resource. But this is not my area of expertise, I'm so we need input from people who use these attributes regularly.

lb42:
I am pretty sure that when @edRef was invented it was precisely thought of as an alternative to @ed -- there are after all plenty of other attributes one might use for pointing to external bibliographic resources in general.

sydb:
I started work on this issue in anticipation of its discussion at the Community Call tomorrow. My plan had been to submit 2 PRs, one for “`@ed` or `@edRef` not both allowed” and one for when they **are** both allowed, figuring folks on the call tomorrow would pick one or the other PR, merge it, and close the other “won’t fix”.

But then two more points occurred to me.
### point the first
[...]

PR #2730, opened 2025-07-08
Addresses #2653

joeytakeda:
This PR proposes a bit of a middle of the ground approach to resolving ambiguity between `@ed` and `@edRef` (as discussed in #2653) by simply copying the discussion of the similar case of `@key` and `@ref` in `att.canonical`. 

I also tried to update the example based on @sydb's suggestion, but would be glad for any other suggestions.

### DIFF
--- a/P5/Source/Specs/att.edition.xml
+++ b/P5/Source/Specs/att.edition.xml
@@ -50,8 +50,9 @@
 	  still living their personal lives in zest and endless
 	  novelty of action, … I saw myself still
 	  preserving, though with increasing difficulty, my lucid
-	  con<pc force="inter">-</pc><pb n="291" break="no" edRef="#stapledon1937"/>sciousness;</p>
+	  con<pb n="291" break="no" edRef="#stapledon1937"/>sciousness;</p>
     </egXML>
+    <p>In the above example, the soft hyphen in Stapledon 1937 is omitted. Such decisions may be documented in the edition's declaration of editorial principles, e.g. with the <gi>hyphenation</gi> element in the <gi>teiHeader</gi>.</p>
   </exemplum>
   <remarks versionDate="2025-07-08" xml:lang="en">
     <p>These guidelines provide no semantic basis or suggested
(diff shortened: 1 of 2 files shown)


P5/Source/Specs/att.edition.xml
```diff
@@ -51,7 +50,15 @@
 	  still living their personal lives in zest and endless
 	  novelty of action, … I saw myself still
 	  preserving, though with increasing difficulty, my lucid
-	  con-<pb n="291" edRef="#stapledon1937"/>sciousness;</p>
+	  con<pb n="291" break="no" edRef="#stapledon1937"/>sciousness;</p>
     </egXML>
+    <p>In the above example, the soft hyphen in Stapledon 1937 is omitted. Such decisions may be documented in the edition's declaration of editorial principles, e.g. with the <gi>hyphenation</gi> element in the <gi>teiHeader</gi>.</p>
```
sydb, line 55: Just want to say that this is _brilliant_, in the sense that it is dramatically more concise than what I might have come up with.
That said, I still prefer replacing the example with one that does not involve a soft hyphen in the differences between editions. I think this para is good enough that we can merge this PR as is, but we should still be looking for a better example.

P5/Source/Specs/att.edition.xml
```diff
@@ -51,7 +50,15 @@
...
   </exemplum>
+  <remarks versionDate="2025-07-08" xml:lang="en">
+    <p>These guidelines provide no semantic basis or suggested
+      precedence when both <att>ed</att> and <att>edRef</att> are
+      provided. For this reason simultaneous use of both is not
+      recommended unless documentation explaining the use is provided,
+      probably in an ODD customization, for interchange.</p>
+  </remarks>
```
sydb, line 63: Since I had not realized @joeytakeda was working on this, I had written my own version, which I like a bit better:
~~~xml
  <remarks versionDate="2025-07-14" xml:lang="en">
    <p>These guidelines provide no suggestion as to the intended
    meaning if both <att>ed</att> and <att>edRef</att> occur on the
    same element. For this reason simultaneous use of both is not
    recommended for interchange unless documentation explaining the
    use is provided, probably in an ODD customizaiton.</p>
  </remarks>
~~~

sydb (changes requested):
Others (@joeytakeda & @trishaoconnor, in particular) should consider the alternate wording of the `<remarks>` which I have provided. Any outcome is fine with me: if they like what is currently on this branch better, or replace it entirely with what I provided, or merge the two. As long as someone checks the final result for spelling, etc., I think this can be merged.

ebeshero:
Reviewed by Council on 12 August 2025, and @joeytakeda 's language looks good to us.

Issue #2719, opened 2025-06-16
example should use `<ab>`, not `<p>`?


sydb:
Two possible problems with the 4th example in [3.13.2 Core Tags for Drama](https://www.tei-c.org/Vault/P5/4.9.0/doc/tei-p5-doc/en/html/CO.html#CODR) (the one preceded by the sentence “By contrast with the preceding examples, the following encodes an early printed edition without making any assumption about which parts are prose or verse”):

1. That preceding sentence made me presume that the preceding examples differentiated dramatic speech in prose from dramatic speech in verse. But they do not seem to.
2. If the example is intended to demonstrate how to encode dramatic speech _without_ asserting whether that speech is verse or prose, wouldn’t it make more sense to use `<ab>` instead of `<p>`? (To be fair to whoever created the example, it was created long before there was an `<ab>`.)

trishaoconnor:
Thank you @sydb for highlighting this. 

I have addressed point 2 and changed `<p>` to `<ab>` in PR #2731. 

Regarding point 1, I have looked at the preceding examples in [3.13.2 Core Tags for Drama](https://www.tei-c.org/Vault/P5/4.9.0/doc/tei-p5-doc/en/html/CO.html#CODR) and, if I understand correctly, I think the prose of Guidelines should not be changed. 

Here is my rationale for keeping the sentence "By contrast with the preceding examples, the following encodes an early printed edition without making any assumption about which parts are prose or verse":

1. The 1st example demonstrates dramatic speech in prose encoded with the `<p>` element.
2. The 2nd example is of dramatic speech encoded as a sequence of verse lines using the `<l>` element
3. The 3rd example is again of dramatic speech encoded as a sequence of verse lines using the `<l>` element and the `@who` attribute

It is my understanding that within this context the first three examples do differentiate between encoding dramatic speech in prose from dramatic speech in verse. Of course, if I have misunderstood, I will happily update the prose of the Guidelines for the fourth example.

trishaoconnor:
This issue has been addressed with PR [2731](https://github.com/TEIC/TEI/pull/2731)

PR #2731, opened 2025-07-09
Replace `<p>` with `<ab>` in Core Tags for Drama example

trishaoconnor:
This PR addresses point 2 of issue #2719 and replaces `<p>` with `<ab>` for the fourth example in the [Core Tags for Drama](https://tei-c.org/release/doc/tei-p5-doc/en/html/CO.html#CODR) section of the Guidelines. Council decided in the July 2025 call that this is a straightforward change and should be merged at your earliest convenience @hennyu and @bleekere.

### DIFF
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -5914,24 +5914,26 @@ are prose or verse:
         noise of Thunder and Lightning heard: Enter
         a Ship-master, and a Boteswaine.</stage>
       <sp>
-         <speaker>Master.</speaker> <p>Bote-swaine.</p>
+         <speaker>Master.</speaker> 
+        <ab>Bote-swaine.</ab>
       </sp>
       <sp>
-         <speaker>Botes.</speaker> <p>Heere Master: What cheere?</p>
+         <speaker>Botes.</speaker> 
+        <ab>Heere Master: What cheere?</ab>
       </sp>
       <sp>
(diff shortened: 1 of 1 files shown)


bleekere:
This seems perfectly fine to me and can be merged. Concerning the original issue #2719 though, @sydb also pointed out that the "the preceding examples differentiated dramatic speech in prose from dramatic speech in verse. But they do not seem to." Should the prose of the guidelines be modified, too?

trishaoconnor:
Thank you @bleekere for reviewing this PR and highlighting that the prose of the Guidelines might require updating in this PR too. 

@sydb and @bleekere, I have looked at the preceding examples in [3.13.2 Core Tags for Drama](https://www.tei-c.org/Vault/P5/4.9.0/doc/tei-p5-doc/en/html/CO.html#CODR) and, if I understand correctly, I think the prose of Guidelines should not be changed. 

Here is my rationale for keeping the sentence "By contrast with the preceding examples, the following encodes an early printed edition without making any assumption about which parts are prose or verse":

1. The 1st example demonstrates dramatic speech in prose encoded with the `<p>` element.
2. The 2nd example is of dramatic speech encoded as a sequence of verse lines using the `<l>` element
3. The 3rd example is again of dramatic speech encoded as a sequence of verse lines using the `<l>` element and the `@who` attribute

It is my understanding that within this context the first three examples do differentiate between encoding dramatic speech in prose from dramatic speech in verse. Of course, if I have misunderstood, I will happily update the prose of the Guidelines for the fourth example.

Issue #762, opened 2013-03-23
Broken links in get.xml on tei-c


TEITechnicalCouncil:
This ticket tracks a couple of questions that have gone unanswered on the Council list:

The page at

http://www.tei-c.org/Guidelines/P5/get.xml

links to this URL:

http://tei.oucs.ox.ac.uk/Query/

which no longer seems to point to anything. 

This one links to a page, but the page doesn't seem relevant to the link 
in get.xml:

`<p>This complex customization is given as an example of
                   how the Oxford University Computing Services use the
                   TEI for authoring its website.  See <ref 
target="http://www.oucs.ox.ac.uk/">http://www.oucs.ox.ac.uk/</ref>
                   to see documents validated against schemas generated 
from this ODD.</p>`

Are there any other candidate targets for these links, or should the relevant sections be deleted?

Original comment by: @martindholmes

PR #2733, opened 2025-07-22
use newer, hopefully better, way to declare Schematron

sydb:
Use a `<constraintDecl>` rather than `<constraintSpec>`s to hold variable and namespace declarations.

WARNING: will not be working entirely correctly until [Stylesheets #762](https://github.com/TEIC/Stylesheets/pull/762) is merged, because the `<xsl:key>` element is not being copied from the ODD to the RNG. (It does make it into the .isosch output file.)

### DIFF
--- a/P5/Exemplars/tei_jtei.odd
+++ b/P5/Exemplars/tei_jtei.odd
@@ -1875,6 +1875,23 @@
       </div>
       <div>
         <schemaSpec ident="tei_jtei" start="TEI" defaultExceptions="http://www.tei-c.org/ns/1.0 eg:egXML">
+	  <constraintDecl scheme="schematron">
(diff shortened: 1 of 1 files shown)


rvdb (approved):
Thanks, @sydb . Tested locally, and I confirm this fixes the copying of at least the embedded `sch:*` declarations when generating an RNG schema. Tested as well: https://github.com/TEIC/Stylesheets/pull/762 fixes the copying of `//constraintDecl/xsl:*` children to RNG.

ebeshero (approved):
Looks good to me! I'm happy it is a helpful improvement for JTEI! I just updated the branch and reviewed locally, trusting @rvdb 's review for the improvement to the JTEI ODD. So I'll merge this one in now.

Issue #2658, opened 2025-01-18
allow byline in closer


martinascholger:
The TEI Lite group proposes to allow `<byline>` in `<closer>`. The definition of `<byline>` is that it “contains the primary statement of responsibility given for a work on its title page or at the head or end of the work". While it is allowed in the `<opener>`, it is not yet allowed in the `<closer>` element.

jamescummings:
Fully support this idea. 👍

sydb:
@trishaoconnor is GO to create a new branch, change the content models and classes so that `<byline>` is an allowed child of `<closer>` and test to see if anything breaks. If nothing breaks, submit as PR.

trishaoconnor:
PR #2735 was created to add `<byline>` as an allowed child of `<closer>`. It passed all tests and has been merged.

PR #2735, opened 2025-07-28
Added `byline` to `closer` content model

trishaoconnor:
This PR addresses issue #2658 to allow `byline` in `closer`. Tested locally via Docker but it doesn't seem to pass tests on GitHub.


trishaoconnor:
Thanks to @sydb and @joeytakeda for highlighting that the tests were failing because I'd mistakenly added `div.modelBottomPart` to `<byline>`. Now that has been corrected ie. now that I have removed `div.modelBottomPart` from `<byline>` the local and GitHub tests are successful and it should be ok to merge @bleekere and @hennyu.

Issue #2718, opened 2025-06-13
entry/type="wordFamily" not fitting in 10.1


bansp:
> The content of the fragment at the end of [10.1](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/DI.html#DIBO):
> ```xml
> <entry type="wordFamily">
>   <entry type="hom" n="1">
>    <form>
>     <orth>grève</orth>
> 
>    </form>
>   </entry>
>   <entry type="hom" n="2">
>    <form>
>     <orth>grève</orth>
> 
>    </form>
>   </entry>
>  </entry>
>  ```
> is, very explicitly, NOT a word family.

 _Originally posted by @bansp in [#2488](https://github.com/TEIC/TEI/issues/2488#issuecomment-2967358456)_ (which is a closed issue but I was afraid that I'd otherwise forget about it)

I understand that it's a matter of showing that the outer `<entry>` can be typed. But that is fairly obvious, given the content model and the examples of other typed entries. I would venture a guess that it's more relevant to type the lowest-level entries than the grouping entries (but I can also very well believe that I am missing some nice use cases).

I would even argue that, in Laurent's [Lex-0 example](https://dariah-eric.github.io/lexicalresources/pages/TEILex0/TEILex0.html#index.xml-body.1_div.3_div.4_div.2) of an Arabic word family based on the same root, the outer `@type` is theoretically a bit spurious, although I can understand that it simplifies the querying and the statistics (which is enough to motivate the redundancy, coming from the fact that "wordFamily" co-occurs with `<form type="root">`, sitting right below that outer `<entry>`).

Long rambling short: 

- Please eliminate this specific type "wordFamily" from up there; if really needed, it could be "hom", demonstrating the redundancy. As it is now, it's very wrong.
- Perhaps consider using the Lex-0 example, to show where typing both outer and inner entries makes sense (even if slightly redundant).

ebeshero:
@bansp Would it be better just to remove the `@type` on the outermost `<entry>` element rather than encourage redundancy? Or is it preferable to set the same `@type` on outer and inner elements?

ebeshero:
Also, @bansp , if we replace this with the longer Lex-0 example, am I correct in thinking there's no issue with `@type='wordFamily'` in the Lex-0 case?

bansp:
[...]

PR #2746, opened 2025-08-12
fix dictionary example per #2718

sydb:
Per @bansp’s [request](#2718), we have fixed the `@type` values in an example and added a new example from LEX-0.
This work brought to you per TEI Council by @sydb, @ebeshero, and @martinascholger.


bansp:
Thanks a bunch! 
I'll review after I'm back from vacation, that is, in the middle of the next week, or so.

ebeshero:
I think what we have here is certainly an improvement because it's addressing the ticket. @HelenaSabel if you agree, I will let you make the call to merge this in to dev and make it part of the 4.10 release this week. Given the unfortunate timing (@bansp can't review til next week), perhaps we accept the risk that we've introduced an error on the grounds that we did at least correct the problem.

We can always open a new ticket to improve the new example for the next release if we need to do that!  
@HelenaSabel , the decision is yours. :-)

HelenaSabel:
I think it’s worth merging it.

Issue #2758, opened 2025-09-02
<speaker> should remain optional


cmil:
The changes for #2642 (#2698 and 38a6db4f5a20f07a3df010e330f0727859cdf75c) seem to have inadvertently made the `speaker` element a required child of `sp`. It has been optional before and I think it should remain so.

lb42:
This must be a simple mistake surely? May be another spot release is needed.

sydb:
I agree, this looks like just an error. But it is not an obvious, easy fix: just making the `( speaker, model.global* )` clause optional results in ambiguity because a member of model.stageLike might be from the previous clause or the following clause. (Which, of course, DTDs and XSD cannot handle.) Sigh.

I have assigned this to myself & @martindholmes since we were the responsible parties for the error. But if anyone else wants to tackle this in order to come up with an unambiguous version, please do!

Desired (but ambiguous) content model:
~~~
(
  ( model.global | model.stageLike )*,
  ( ( speaker, model.global* )? )?,
  (
    (
       (
          lg
        | model.lLike
        | model.pLike
        | model.listLike
        | model.stageLike
        | model.attributable
       ),
       ( model.global* | q )
     )+
  )
)
~~~
I am thinking that moving the 2nd reference to model.stageLike to the clause at the end that has the `q` (why is there a `<q>` allowed there?) would solve the problem, but cannot test now, I have to go catch a train …

ebeshero:
I just confirmed the error with this simple test file, and also confirmed that it does not appear in version 4.9.0 of the Guidelines. So we definitely introduced it with the 4.10.0 release. :-( 

```xml
<TEI xmlns="http://www.tei-c.org/ns/1.0">
   <teiHeader>
      <fileDesc>
         <titleStmt>
            <title>Title</title>
         </titleStmt>
         <publicationStmt>
            <p>Publication Information</p>
         </publicationStmt>
         <sourceDesc>
            <p>Information about the source</p>
         </sourceDesc>
      </fileDesc>
   </teiHeader>
   <text>
      <body>
         <div>
            <sp>
               <milestone unit="page"/><!--model.global-->
               <stage>hi</stage>
             <!--  <speaker>...</speaker>-->
               <lg><l>...</l></lg>
            </sp>
         </div>
      </body>
   </text>
</TEI>
```

ebeshero:
[...]

PR #2759, opened 2025-09-02
attempted fix for #2758

ebeshero:
@sydb @martindholmes @raffazizzi : This branch isn't passing tests yet, but I think we need a few more eyes on it to figure out how to fix it. 
It's an attempt to solve #2758 on the way to (most likely) another patch release.

### DIFF
--- a/P5/Source/Specs/sp.xml
+++ b/P5/Source/Specs/sp.xml
@@ -52,7 +52,7 @@
    )
     -->
     <!--
-	Where:
+        Where:
 spPart_sans = ( lg | model.lLike | model.pLike | model.listLike | model.attributable | q )
 spPart_con  = ( spPart_sans | model.stageLike )
     -->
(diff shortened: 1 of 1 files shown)


P5/Source/Specs/sp.xml
```diff
@@ -31,30 +31,21 @@
...
-      </alternate>
-      <sequence>
-        <elementRef key="speaker"/>
-        <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
-      </sequence>
-      <sequence minOccurs="1" maxOccurs="unbounded">
-        <alternate>
+        <alternate minOccurs="1" maxOccurs="unbounded">
+          <classRef key="model.global"/>
+          <elementRef key="speaker"/>
```
cmil: This would allow any number of `speaker` elements at any position inside `sp`, wouldn't it?

lb42: I agree with Carsten!

ebeshero: Yes, this solution is overly permissive in that sense—particularly in allowing `<speaker>` anywhere. (Changing it to zero or one is probably not so big a problem to fix as limiting where it can occur.) We could add Schematron to constrain it further, and I figured it would hold if we can't find a better way quickly! (But I suspect we can.) I hope this is at least a starting point for a better solution.

sydb:
@ebeshero opened this PR, and thus apparently cannot be a reviewer. Nonetheless, if anyone can check this branch out and test to see if your `<sp>`s break, that would be good.

In the meantime we should start setting up for a patch release soon. (E.g., establish Jenkins jobs.)

ebeshero:
Okay, I can confirm that @sydb 's fix not only passes tests, but is also working when I build tei_all.rng and validate my test file (attached). 
*  That test file currently doesn't contain a `<speaker>` element, and it's valid.
*  I can also position a `<speaker>` element in as the very first element, or after a `<stage>`element, or something from model.global. 
* I cannot position a `<speaker>` element after the spoken passge. 
* Nor can I position multiple `<speaker>` elements anywhere. 

So I think this will suffice for a patch. But I am wondering if we really want to continue with the complexity of this new content model for `<sp>`.

```xml
<content>
  <sequence>
      <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
      <alternate>
        <sequence>
          <sequence minOccurs="1" maxOccurs="unbounded">
            <classRef key="model.stageLike" minOccurs="1" maxOccurs="1"/>
            <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
          </sequence>
          <alternate>
            <sequence>
              <elementRef key="speaker" minOccurs="1" maxOccurs="1"/>
              <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
              <sequence minOccurs="1" maxOccurs="unbounded">
                <alternate minOccurs="1" maxOccurs="1">
                  <elementRef key="lg"/>
                  <classRef key="model.lLike"/>
                  <classRef key="model.pLike"/>
                  <classRef key="model.listLike"/>
                  <classRef key="model.attributable"/>
                  <elementRef key="q"/>
                </alternate>
                <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
              </sequence>
            </sequence>
            <sequence>
              <alternate minOccurs="1" maxOccurs="1">
                <elementRef key="lg"/>
                <classRef key="model.lLike"/>
                <classRef key="model.pLike"/>
                <classRef key="model.listLike"/>
                <classRef key="model.attributable"/>
                <elementRef key="q"/>
              </alternate>
              <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
              <sequence minOccurs="0" maxOccurs="unbounded">
                <alternate minOccurs="1" maxOccurs="1">
                  <elementRef key="lg"/>
                  <classRef key="model.lLike"/>
                  <classRef key="model.pLike"/>
                  <classRef key="model.listLike"/>
                  <classRef key="model.attributable"/>
                  <elementRef key="q"/>
                </alternate>
                <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
              </sequence>
            </sequence>
          </alternate>
        </sequence>
        <sequence>
          <elementRef key="speaker" minOccurs="1" maxOccurs="1"/>
          <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
          <sequence minOccurs="1" maxOccurs="unbounded">
            <alternate minOccurs="1" maxOccurs="1">
              <elementRef key="lg"/>
              <classRef key="model.lLike"/>
              <classRef key="model.pLike"/>
              <classRef key="model.listLike"/>
              <classRef key="model.attributable"/>
              <elementRef key="q"/>
            </alternate>
            <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
          </sequence>
        </sequence>
        <sequence>
          <alternate minOccurs="1" maxOccurs="1">
            <elementRef key="lg"/>
            <classRef key="model.lLike"/>
            <classRef key="model.pLike"/>
            <classRef key="model.listLike"/>
            <classRef key="model.attributable"/>
            <elementRef key="q"/>
          </alternate>
          <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
          <sequence minOccurs="0" maxOccurs="unbounded">
            <alternate minOccurs="1" maxOccurs="1">
              <elementRef key="lg"/>
              <classRef key="model.lLike"/>
              <classRef key="model.pLike"/>
              <classRef key="model.listLike"/>
              <classRef key="model.attributable"/>
              <elementRef key="q"/>
            </alternate>
            <classRef key="model.global" minOccurs="0" maxOccurs="unbounded"/>
          </sequence>
        </sequence>
      </alternate>
    </sequence>
 
  </content>


```

Is it too much? Would introducing `model.stageLike` to model.global allow a simpler solution? 


[sp-speaker-testing.xml](https://github.com/user-attachments/files/22106225/sp-speaker-testing.xml)

ebeshero:
Also @sydb , whether you take that as a review or not, I think in the interests of time, any one of us should review your solution on this branch. The fact that I made the branch shouldn't matter at this point.  I will argue that a simpler content model might be desirable for human readability. Whether mine is too simple depends on how imperative we find constraining the position of `<speaker>` to ensure it can't come after the spoken portion of a speech.

sydb:
Keep in mind the content model is a lot simpler after we create the “spPart_con” and “spPart_sans” classes (or macros), whatever they may be named.
My instinct is to do the emergency patch release without the nicety of the new classes for the sake of expediency. Then we can work on the classes, deciding on their description, names, etc., at our leisure over the next month or three.

sydb:
BTW, I do **not** think this PR should be accepted as-is — the changes in "ebb-fix-#​2758" should be merged into a new "release-4.10.2" branch, not "dev" (which is what this PR would do).
@ebeshero, do you want to change this PR, create a new one, talk me into merging directly into "dev", or override my objection and merge directly into "dev"?

lb42:
Adding stage to model.global would permit retaining the constraint that speaker if present must precede what is spoken which i think is important. And is obviously much simpler.

sydb:
The proposed content model requires `<speaker>` to precede what is spoken, or at least it is supposed to. And @ebeshero’s tests say that it does.

lb42:
@sydb my point was not that the proposed content model doesn't do the jpob -- clearly it does. My point is that the same objective might be obtained with a much simpler content model , just by making stage a global element. And simplicity and consequent explainability are desirable.

ebeshero:
Closing this PR since we transferred to the work of it to the 4.10.2 release branch, made from 4.10.1.

PR #2760, opened 2025-09-03
Another crack at fixing content model of `<sp>`.

sydb:
[Temporary storage for new version of content model, as release team wants it merged into the release branch immediately.]

This update changes the content model of `<sp>` in order to

- allow `<stage>` before `<speaker>`
- revert `<speaker>` to being optional (which was the error in 4.10.0)
- maintain determinism (i.e., be unambiguous)
- without adding any new classes or macros (because this is a patch release)

The basic idea is to use a sequence of 3 clauses, here represented in RELAX NG compact syntax with added clause numbers preceding:
~~~
 [1]   model.stageLike*,
 [2]   (
[2a]     ( speaker, ( model.spPart | model.stageLike ) )
         |
[2b]     model.spPart
       ),
 [3]   ( model.spPart | model.stageLike )*
~~~
But (of course) we also need to allow for global elements. Thus it would need to be a 4-part model, as we need an extra part at either the beginning or end to account for leading or trailing global elements; in this case I have put the extra part at the beginning and used `( A, global* )` instead of `A`, as is common TEI practice per [EDW69 chapter 8](https://tei-c.org/Vault/ED/edw69.htm#ID8):
~~~
 [1]   model.global*,
 [2]   ( model.stageLike, model.global* )*,
 [3]   (
[3a]     ( ( speaker, model.global*), ( ( model.spPart | model.stageLike ), model.global* ) )
         |
[3b]     ( model.spPart, model.global* )
       ),
 [4]   ( ( model.spPart | model.stageLike), model.global* )*
~~~
But since we did not want the overhead of adding a new class in a patch release, the actual content model is somewhat more complex:
~~~
model.global*,
( model.stageLike, model.global* )*,
  (
    ( ( speaker, model.global*), ( ( lg | model.lLike | model.pLike | model.listLike | model.attributable | q | model.stageLike), model.global* ) )
    |
    ( ( lg | model.lLike | model.pLike | model.listLike | model.attributable | q ), model.global* )
  ),
 ( ( lg | model.lLike | model.pLike | model.listLike | model.attributable | q | model.stageLike ), model.global* )*
~~~
It is my hope (I have not really asked anyone else yet) that at the next full release we can add something like a new model.spPart class such that we can reduce the complexity of the content model. In preparation for this case I have added the simpler content model (using “model.spPart” as the name of the new class) to the ODD file in a comment.

If _anyone_ can check out this branch and test that it works for you, that would be good. Even if not, though, I plan to merge it into the patch release branch (release-4.10.2) soon so that Mr. Jenkins can start his efforts.


martindholmes (approved):
Good to go, we believe.

PR #2767, opened 2025-09-07
easier-to-read Schematron for nested `<standOff>`

sydb:
I am very sure this replacement is easier to read.
I am reasonably sure this replacement does exactly the same thing.
But I am not at all sure this constraint makes any sense in the first place. Can a `<standOff>` ever appear as a descendant of a `<standOff>`?
In my current tired state I think the answer is “no”, and that this entire constraint specification ("nested_standOff_should_be_typed") should be deleted.

### DIFF
--- a/P5/Source/Specs/standOff.xml
+++ b/P5/Source/Specs/standOff.xml
@@ -15,10 +15,10 @@
   </content>
   <constraintSpec ident="nested_standOff_should_be_typed" scheme="schematron" xml:lang="en">
     <constraint>
-      <sch:rule context="tei:standOff">
-        <sch:assert test="@type or not(ancestor::tei:standOff)">This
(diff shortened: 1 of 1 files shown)


jamescummings:
standOff is contained only by TEI and TEICorpus; Those are only contained by themselves in some ways. So as far as I can see (without testing) there is no way to get a standOff inside a standOff unless one is changing the model classes to which standOff claims membership or the content model of other elements.  Is thathappening?

sydb:
Thank you, @jamescummings. That is exactly my reading. So despite the fact that I am proud of my tiny Schematron improvement, I am going to kill this PR as a “won’t fix” and submit another that just removes what appears to be a crazy constraint. (It is probably a holdover from pre-release iterations of `<standOff>`.)

PR #2769, opened 2025-09-07
remove extraneous never-used constraint

sydb:
Per conversation on #2767, remove requirement for `@type` on a nested `<standOff>`, because there is no such thing as a nested `<standOff>`.


bansp (approved):
Looks fine to me. Thanks, @sydb .

Issue #869, opened 2010-08-09
half title pages in TEI Tite


TEITechnicalCouncil:
TEI Tite says a half title page should be encoded using &lt;div1 type="half-title"&gt; (with titlePart elements as appropriate); however, neither Tite nor P5 allows use of titlePart within a div1.

Original comment by: @kshawkin

hcayless:
Who is responsible for Tite nowadays? We don't see why `<titlePage>` / `<docTitle>` shouldn't be recommended instead of `<div1>`. And recommend `<titlePage type="half">`.

kshawkin:
Council decided in 2012 to maintain Tite, and I have volunteered to continue implementing tickets even after finishing my term.  I was encouraged to exercise my judgement in proceeding ( http://lists.village.virginia.edu/pipermail/tei-council/2014/020321.html ), but I confess I've ignored most of these things.  Thanks for the guidance on this particular ticket.  I'll try to get back to this project ( http://wiki.tei-c.org/index.php/User:Kshawkin/My_Council_obligations/Plan_for_TEI_Tite ), which is part of my larger guilt list ( http://wiki.tei-c.org/index.php/User:Kshawkin/My_Council_obligations ).

kshawkin:
While Tite currently includes `<titlePage>`, it "includes only a limited set of all elements in the TEI." For that reason, I'm not sure that I want to recommend use of this niche element; instead, it might be better just to recommend use of `<div1>`, perhaps without using `<titlePart>` within it.  I'm still pondering this.

lb42:
No progress since March. Should this be reassigned?

kshawkin:
Webmastering is taking up all of my TEI time, but I'm hoping to offload most of those duties soon, which will allow me to return to my leftover Council obligations.  However, if you've got Council members looking for things to do, feel free to take it from me.

jamescummings:
This is currently our oldest open issue, dating from 2010-08-09 in SourceForge. Just making a note since I was looking up what the oldest one was for a presentation. ;-)

martindholmes:
I believe that nine years with not a word of progress and no complaint suggests that this is no longer relevant to anyone. However, I hereby confirm that the following is valid against the current TEI Tite schema:

~~~~~
<text xmlns="http://www.tei-c.org/ns/1.0">
  <front>
    <titlePage type="half">
      <docTitle>
        <titlePart>Testing</titlePart>
      </docTitle>
    </titlePage>
  </front>
  <body>
[...]

PR #2778, opened 2025-09-10
Fix for issue #869.

martindholmes:
Following a check to show that the Tite schema does actually seem to allow what its prose suggests, this is a minor amelioration to address issue #869 so that we can finally close our oldest open issue (thanks @jamescummings). One review should be enough for a merge, I think.


martindholmes:
Assigning the PR to @sydb because Council at the FTF agreed to the fix but he wanted a rewording, now done.

martindholmes:
@sydb Can we merge this now?

sydb (approved):

Issue #2787, opened 2025-09-15
gloss of `@rendition` should be “rendition reference”


sydb:
Per @torstenroeder’s suggestion, we should add a `<gloss>` to the `@rendition` of att.global.rendition that says “rendition reference”.

martinascholger:
Solved by https://github.com/TEIC/TEI/pull/2791

PR #2791, opened 2025-09-15
gloss for rendition reference #2787

torstenroeder:
add gloss in en,de,fr to @rendition that says "rendition reference" (and removed some ws)


P5/Source/Specs/att.global.rendition.xml
```diff
@@ -102,16 +78,13 @@ the TEI header.</p>
     <attDef ident="rendition" usage="opt">
       <gloss versionDate="2025-09-15" xml:lang="en">rendition reference</gloss>
       <gloss versionDate="2025-09-15" xml:lang="de">Darstellungsverweis</gloss>
-      <desc versionDate="2007-09-22" xml:lang="en">points to a description of the rendering or presentation used for this element in the
-        source text.</desc>
+      <gloss versionDate="2025-09-15" xml:lang="fr">interprétation référence</gloss>
```
martinascholger: I wonder if this should be "description du rendu" instead of "interprétation référence" - see `<desc xml:lang="fr">`.

martinascholger:
For the German translation I ask @hennyu and/or @torstenroeder to approve

P5/Source/Specs/att.global.rendition.xml
sydb: I do not speak French, but according to translate.google.com, neither would be precisely correct. It seems to like "référence à l'interprétation" or "référence de rendu".

martindholmes:
It looks like "interprétation référence" would be a better gloss for `@ana` than `@rendition`. Since "rendu" is the French gloss of the `<rendition>` element, I also vote for "description du rendu".

martinascholger:
I asked my colleague - he recommends `référence du rendu`

hennyu (changes requested):
The German description that Martina asked us to look at is fine to me. This french passage should be changed:
<gloss versionDate="2025-09-15" xml:lang="fr">interprétation référence</gloss>
It should be "référence du rendu", as suggested in the comments, before this PR is merged.

ebeshero:
The explanations and all the revisions are very clear, and @martinascholger has just explained them all to me. I am happy to approve and merge this PR.

PR #2807, opened 2025-09-16
worked on ticket 2316-lem

hennyu:
added speaker element to the content model of lem


martindholmes (approved):
Exactly as requested on the ticket.

torstenroeder (approved):

Issue #2786, opened 2025-09-15
Deprecate or rename <foreign>


raffazizzi:
The use of the term "foreign" can be problematic in multilingual contexts where no language (or the people that speak it) is arguably "foreign" to the one mainly used in a given doument. This has been raised as a concern by a few projects including:

* https://scholarlyediting.org/issues/41/kinship-and-longing/ (reported by @raffazizzi)
* https://ogham.celt.dias.ie/ (reported by @trishaoconnor)
* https://lyoninmourning.dhil.lib.sfu.ca/ (reported by @joeytakeda)
* https://adaptiveteinetwork.arts.ubc.ca/ (reported by @joeytakeda)

We think that `@xml:lang` is sufficient to indicate that a word or phrase belongs to some language other than that of the surrounding text, so it could be simply be used on `<seg>` or `<w>` to perform the same encoding action, or if a distinct element is necessary, we may want to reconsider its name.

lb42:
There is also the existing `<distinct>` element which some may prefer

raffazizzi:
`distinct` was brought up in conversation but we know at least of one projects was uncomfortable with its current gloss (we could adjust it), particularly the examples provided in the second clause. Quoted here for reference: "identifies any word or phrase which is regarded as linguistically distinct, for example as archaic, technical, dialectal, non-preferred, etc., or as forming part of a sublanguage."

joeytakeda:
My opinion is that `<foreign>` and `<distinct>` are elements with specific semantics, which are useful only if you want those semantics -- i.e. to say that your interpretation of this text is that this span of stuff is _foreign_  or _distinct_, and not just in a different language. The Guidelines should you want to say that this is in another language, then just use `<seg>`.

trishaoconnor:
Thank you for opening this ticket @raffazizzi. Recently several linguistic researchers highlighted that the term "foreign" was problematic within the context of encoding national languages such as Irish and Welsh. The explanation that the language is foreign to the language of the encoded file was understood but not satisfactory. 

[...]

PR #2814, opened 2025-09-16
<foreign> clarifications and more alternatives

raffazizzi:
This PR attempts to clarify and offer alternatives to the use of `<foreign>` in multilingual texts, or texts originating from multilingual contexts. See #2786.

I am initially setting this as draft to prompt feedback and suggestions on the proposed changes.

### DIFF
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -576,8 +576,7 @@ weak bees let them in.</q></egXML>
     where there was "<del>a</del> <seg xml:lang="sco">Deel a Sole</seg> upon the Shoe"
   </egXML>
     <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="COHQF-egXML-kbl" source="#COHQHF-eg-kbl">
-    <closer xml:lang="es">y firmo su Mrd. Con las demas partes Expressa
-      das. de que yo el Es.<hi rend="sup">no</hi> Doy fe
+    <closer xml:lang="es">y firmo su Mrd. Con las demas partes Expressadas. de que yo el Es.<hi rend="sup">no</hi> Doy fe
       <lb/>Derneville Fran.<hi rend="sup">co</hi> Broutin
       <lb/>D Leonardo Mazange Frãn.<hi rend="sup">co</hi> Liotau
       <lb/>Louis Lessier <emph xml:lang="fr">fils</emph>
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -584,7 +584,7 @@ weak bees let them in.</q></egXML>
   </egXML>
 </p>
 
-<p>The <gi>foreign</gi> element should additionally not be used to represent foreign words
+<p>The <gi>foreign</gi> element should additionally not be used to represent words in another language
 which are mentioned or glossed within the text: for these use the
 appropriate element from section <ptr target="#COHTG"/> below. Compare the
 following example sentences:
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -567,8 +567,8 @@ weak bees let them in.</q></egXML>
   provide a peg onto which the <att>xml:lang</att> may be attached.
 </p>
   
-<p>The <gi>foreign</gi> element should not be used in multi-lingual texts or texts originating from multilingual contexts because the text 
-  in question may not be <soCalled>foreign</soCalled> to its intended audience.
+<p>The <gi>foreign</gi> element should not be used in multilingual texts or texts originating from multilingual contexts because the words 
+  in question may not be <soCalled>foreign</soCalled> to their intended audience.
   In these cases it is also preferable to use <gi>seg</gi> with the appropriate <att>xml:lang</att>, or <gi>emph</gi> if the text is emphasized.</p>
   <p>The following examples originate from bilingual contexts, namely English and Scots and Spanish and French. 
     <!-- TODO: add source -->
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -428,7 +428,7 @@ phenomenon and alternate approaches to encoding it, this section
 describes ways of encoding the following textual features, all
 of which are conventionally rendered using some kind of highlighting:
 <list rend="bulleted">
-<item>emphasis, foreign words and other linguistically distinct uses
+<item>emphasis, words in other languages or other linguistically distinct uses
 of highlighting</item>
 <item>representation of speech and thought, quotation, etc.</item>
 <item>technical terms, glosses, etc.</item></list>
@@ -467,7 +467,7 @@ places in older texts, editorial corrections or additions, etc.</item></list>
  </p>
 <p>The textual functions indicated by highlighting may not be rendered
 consistently in different parts of a text or in different texts. (For
-example, a foreign word may appear in italics if the surrounding text is
+example, a word in another language may appear in italics if the surrounding text is
 in roman, but in roman if the surrounding text is in italics.)  For this
 reason, these Guidelines distinguish between the encoding of rendering
 itself and the encoding of the underlying feature expressed by it.
@@ -543,11 +543,11 @@ which are not discussed in this section include <gi>title</gi>
 (discussed in section <ptr target="#COBI"/>) and <gi>name</gi> (discussed
 in section <ptr target="#CONARS"/>).
  </p></div>
-<div type="div3" xml:id="COHQH"><head>Emphasis, Foreign Words, and Unusual Language</head>
+<div type="div3" xml:id="COHQH"><head>Emphasis, Words in Other Languages, and Unusual Language</head>
 <p>This subsection discusses the following elements:
 <specList><specDesc key="foreign"/><specDesc key="emph"/><specDesc key="distinct"/></specList>
 These elements are all members of the <ident type="class">model.emphLike</ident> class. </p>
-<div type="div4" xml:id="COHQHF"><head>Foreign Words or Expressions</head>
+<div type="div4" xml:id="COHQHF"><head>Words or Expressions in Other Languages</head>
 <p>Words or phrases which are not in the main language of the text
 should be tagged as such, at least where the fact is indicated in the
 text. Where the word or phrase concerned is already distinguished from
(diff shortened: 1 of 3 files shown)


P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
+  
+<p>The <gi>foreign</gi> element should not be used in multi-lingual texts or texts originating from multilingual contexts because the text 
+  in question may not be <soCalled>foreign</soCalled> to its intended audience.
+  In these cases it is also preferable to use <gi>seg</gi> with the appropriate <att>xml:lang</att>, or <gi>emph</gi> if the text is emphasized.</p>
+  <p>The following examples originate from bilingual contexts, namely English and Scots and Spanish and French. 
+    <!-- TODO: add source -->
+  <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="COHQF-egXML-aa" >
+    where there was "<del>a</del> <seg xml:lang="sco">Deel a Sole</seg> upon the Shoe"
+  </egXML>
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="COHQF-egXML-kbl" source="#COHQHF-eg-kbl">
```
ebeshero, line 578: Perfect example!

ebeshero (approved):
These strike me as sensible edits, and the example is brilliant! (Don't forget to add the source to BIB!)

martindholmes (approved):
These changes are a definite improvement. The only quibbles I have are tiny and have nothing to do with the intent or motivation for the ticket. 

I liked the new example with Latin (which is technically foreign to everyone now), up to "Bee Master", but the following sentence threw me because I have no idea what the text is, so the sudden mention of wax-moths let me momentarily puzzled. I don't think that final sentence is needed (although it might be enjoyable if you happen to know the text), and if I put myself in the position of a non-native, non-European speaker I can imagine reaching for a dictionary to find out what a wax-moth is and wondering how it's relevant to the "foreign" thing. So I would lose that sentence.

In the Scots example, though, I feel there's somehow not enough context for the words to mean anything much; perhaps it would help to gloss the phrase, again to help non-native speakers: " = 'a [great] deal of sole', or a substantial sole" (assuming that's what it actually means).

I wonder about mentioning `<emph>` here:

`In these cases it is also preferable to use <gi>seg</gi> with the appropriate <att>xml:lang</att>, or <gi>emph</gi> if the text is emphasized.`

If a term in a different language from the surrounding text is distinguished in some way such as italicization, I think it's most likely going to be because it's in another language rather than because it's being emphasized, and in any case, how would you distinguish the two? Perhaps it's unwise to suggest that simply being in another language might constitute a kind of emphasis. But I see your example below uses it. Is "fils" really emphasized, or just in another language and therefore italicized?

None of these are a big deal for me, though. If you want to go ahead and merge, I'll be happy.

lb42:
The wax moth example is not new! I chose it many years ago.

raffazizzi:
Thanks @martindholmes . I think @joeytakeda has asked his team for other Scots examples, so we may end up with a clearer one.

In the "fils" example, the emphasis is editorial; that is, the source document is a manuscript, but the editors want the word italicized because it is in another language than its surrounding text. At same time they don't want to call it foreign because it is't in the historical context of the document. Perhaps a `<seg rend="italic" xml:lang="fr">fils</seg>` would be more neutral and achieve the same effect...

lb42:
Surely that would imply that the word was italicised in the original ms? If this is all about how to render the markup in the edited version thats a job for the render ing software not the markup methinks.

raffazizzi:
Right, yes – this is probably not a good enough example to justify the use of `<emph>` as an alternative to `<foreign>`. It can still work for showcasing using `<seg>`.

lb42:
Also, while I'm here, I think an example using `<distinct>` would also be useful here. Happy to look for one !

martindholmes:
@lb42 Sorry, I must have missed the wax moth example before, or never focused on it. But my comment still stands: I think it would present a barrier to a non-native speaker (or even a native speaker who finds it difficult to skip over a puzzle).

martindholmes:
@raffazizzi If "fils" is not italicized in the original, then presumably there's no signal to say that it is emphasized in the original at all, so `<seg xml:lang="fr">` would probably  be safest.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
+    <closer xml:lang="es">y firmo su Mrd. Con las demas partes Expressadas. de que yo el Es.<hi rend="sup">no</hi> Doy fe
+      <lb/>Derneville Fran.<hi rend="sup">co</hi> Broutin
+      <lb/>D Leonardo Mazange Frãn.<hi rend="sup">co</hi> Liotau
+      <lb/>Louis Lessier <emph xml:lang="fr">fils</emph>
+    </closer>
+  </egXML>
 </p>
 
-<p>The <gi>foreign</gi> element should not be used to represent foreign words
+<p>The <gi>foreign</gi> element should additionally not be used to represent foreign words
```
trishaoconnor: Perhaps "should additionally not be used to represent words in a different language which are mentioned or glossed within the text..." to be consistent with the changes above and below?

trishaoconnor (approved):
Thank you @raffazizzi for your time and effort in adding this extra clarification and guidance. I welcome these edits, I agree that this is a definitive improvement to the _Guidelines_ chapter. I like the example provided for `<emph>` but I understand the concerns already raised here. Hopefully, the example that @joeytakeda has volunteered to search for would provide a better use case for <emph>. 
I only have one small suggestion for line 587 in the CO chapter so that it is consistent with the phrasing of the rest of the edits.

raffazizzi:
The Keywords of Black Louisiana team also provided English translations of the documents (in French+Spanish) that they edited. They used `emph[@xml:lang]` instead of `foreign` whenever they retained in the translation original terminology in French or Spanish from the documents. I would say that in these cases the words are indeed foreign to English, but the editors didn't want to impose "foreignness" even on the translation of this multi-lingual and culturally sensitive material.

Is this too unique a stance for the Guidelines? I think it's legitimate. Note that in this case the `emph` is the TEI's authors emphasis rather than the being derived from a source document.

Here is an example that could be added to the chapter instead of the one I presented earlier.

```xml
<p>
   <lb/>When asked about the statement of
      the <emph xml:lang="fr">sauvagesses</emph> he
   <lb/>replied no, he answers the question. 
</p>
```

Which is the editor's translation of:

```xml
<p>
     <lb/>Intérogé Surla Cite du sauvagesses il
     <lb/>répond que non, Il répond l'interoge… 
 </p>
```

lb42:
Wouldnt <distinct> be more appropriate than <emph> here?

P5/Source/Specs/foreign.xml
```diff
@@ -55,8 +55,13 @@ Ce ne seroyt que bon que nous rendissiez noz cloches...</egXML>
                         the phrase or words concerned. The global <att>xml:lang</att> attribute
                         should be used in preference to this element where it is intended to mark
                         the language of the whole of some text element.</p>
+    <p>In multilingual texts, or texts originating from multilingual contexts, this element may not be appropriate
+      for marking words or phrases as belonging to some language other than that of the
+      immediately surrounding text. The text  in question may not be <soCalled>foreign</soCalled> to its intended audience.
+      In these cases it is preferable to use <gi>seg</gi> with the appropriate <att>xml:lang</att>, or <gi>emph</gi> if the text is emphasized.</p>
```
tuurma, line 61: why not suggest the use of `distinct` here, if the word or phrase is linguistically distinct, why muddle things up with the generic `seg` or imply emphasis where it is a separate characteristic (a phrase may be both linguistically distinct and emphasized)?

lb42: I am pretty sure i also  suggested using distinct when this issue was first raised

lb42: I am pretty sure i also  suggested using distinct when this issue was first raised

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
-attribute and its values respectively). Where there is no other
-applicable element, the element <gi>foreign</gi> may be used to
-provide a peg onto which the <att>xml:lang</att> may be attached.
+attribute and its values respectively). 
 <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="COHQHF-egXML-me" source="#COHQHF-eg-5"><q>Aren't you confusing <foreign xml:lang="la">post hoc</foreign> with <foreign xml:lang="la">propter
 hoc</foreign>?</q> said the Bee Master. <q>Wax-moth only succeed when
 weak bees let them in.</q></egXML>
+  Where there is no other
+  applicable element, the element <gi>seg</gi> may be used to
+  provide a peg onto which the <att>xml:lang</att> may be attached.
```
tuurma, line 567: *wrapper* instead of *peg* perhaps, to avoid too flowery language (thinking about Raffs plea for simple English phrasing)

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
+</p>
+  
+<p>The <gi>foreign</gi> element should not be used in multilingual texts or texts originating from multilingual contexts because the words 
+  in question may not be <soCalled>foreign</soCalled> to their intended audience.
+  In these cases it is also preferable to use <gi>seg</gi> with the appropriate <att>xml:lang</att>, or <gi>emph</gi> if the text is emphasized.</p>
```
tuurma, line 572: how about recommending `distinct`? I always considered it like `foreign` but without the judgment

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
+attribute and its values respectively). 
 <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="COHQHF-egXML-me" source="#COHQHF-eg-5"><q>Aren't you confusing <foreign xml:lang="la">post hoc</foreign> with <foreign xml:lang="la">propter
 hoc</foreign>?</q> said the Bee Master. <q>Wax-moth only succeed when
 weak bees let them in.</q></egXML>
+  Where there is no other
+  applicable element, the element <gi>seg</gi> may be used to
+  provide a peg onto which the <att>xml:lang</att> may be attached.
+</p>
+  
+<p>The <gi>foreign</gi> element should not be used in multilingual texts or texts originating from multilingual contexts because the words
```
tuurma, line 570: I think *should not* is a bit strong, perhaps *might not be adequate*?

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
 <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="COHQHF-egXML-me" source="#COHQHF-eg-5"><q>Aren't you confusing <foreign xml:lang="la">post hoc</foreign> with <foreign xml:lang="la">propter
 hoc</foreign>?</q> said the Bee Master. <q>Wax-moth only succeed when
 weak bees let them in.</q></egXML>
+  Where there is no other
+  applicable element, the element <gi>seg</gi> may be used to
+  provide a peg onto which the <att>xml:lang</att> may be attached.
+</p>
+  
+<p>The <gi>foreign</gi> element should not be used in multilingual texts or texts originating from multilingual contexts because the words 
+  in question may not be <soCalled>foreign</soCalled> to their intended audience.
```
tuurma, line 571: on second thought the introduction of the intended audience and what they might think foreign is going far - and unnecessary. How about:

`The <gi>foreign</gi> element might not be suitable in multilingual texts or texts originating from multilingual contexts because the concept of <hi>foreignness</hi> is not necessarily adequate in such environments.`

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -600,7 +618,7 @@ indicated by the use of underlining.  As the following examples
 demonstrate, an encoder may choose whether or not to make explicit the
 particular type of rendition associated with the emphasis. If a source
 text consistently renders a particular feature (e.g. emphasis or words
-in foreign languages) in a particular way, the rendering associated
+in a different language) in a particular way, the rendering associated
```
tuurma, line 621: `certain` language instead of `different`?

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -558,15 +558,33 @@ specify additionally that its language distinguishes it from the
...
+    <closer xml:lang="es">y firmo su Mrd. Con las demas partes Expressadas. de que yo el Es.<hi rend="sup">no</hi> Doy fe
+      <lb/>Derneville Fran.<hi rend="sup">co</hi> Broutin
+      <lb/>D Leonardo Mazange Frãn.<hi rend="sup">co</hi> Liotau
+      <lb/>Louis Lessier <emph xml:lang="fr">fils</emph>
+    </closer>
+  </egXML>
 </p>
 
-<p>The <gi>foreign</gi> element should not be used to represent foreign words
+<p>The <gi>foreign</gi> element should additionally not be used to represent words in another language
```
tuurma, line 587: if you're not using it, then it's not `additionally`, methinks. If emphasis is meant, then maybe `In particular, the foreign element should not...`

tuurma (changes requested):
approve in principle, made some suggestions re wording

please consider if you really want to recommend `seg` and not `distinct` instead

P5/Source/Guidelines/en/CO-CoreElements.xml
tuurma: sorry, repeating myself, now I see ;-)

Issue #2317, opened 2022-06-25
<app> as sole content of a <sp> element


gabrielegan:
I propose that `<app>` should be allowed to be the sole content of a `<sp>` element. At the moment this is invalid:
```
  <sp who="#CHOLMLEY">
    <app type="textual">
      <lem wit="#MS1">
        <l>Not so, my lord, it does no good.</l>
      </lem>
      <rdg wit="#MS2">
        <p>I did not think so.</p>
      </rdg>
    </app>
  </sp>
```

The reason seems to be that an `<app>` can't be the sole content of an `<sp>`, since adding a single empty `<l>` element fixes the above, and the following is valid:
```
  <sp who="#CHOLMLEY">
    <l></l>
    <app type="textual">
      <lem wit="#MS1">
        <l>Not so, my lord, it does no good.</l>
      </lem>
      <rdg wit="#MS2">
        <p>I did not think so.</p>
      </rdg>
    </app>
  </sp>
```
Surely that makes no sense and there is no good reason why an `<app>` can't be the sole content of an `<sp>` element. If we
agree on that, I'd like to propose that the TEI standard is altered to fix this too.

raffazizzi:
This makes sense to me but we'll need to find a way to make sure the `<app>` contains the content model usually required by `<sp>`. Will discuss with the council.

ebeshero:
See also #2316

torstenroeder:
Subgroup F2F 2025 agreed that any of the elements currently available in `<sp>` should be available in any order (0 or more).

ebeshero:
Noted that during Council discussion of 2026-02-09, we concluded that the added example is sufficiently clear for those familiar with the `<app>` element, but adding new prose to the Guidelines might introduce confusion for those not so familiar. So we concluded that this ticket is resolved with PR #2817 .

PR #2817, opened 2025-09-20
Issue #2317: update content model for <sp>

torstenroeder:
This PR resolves issue #2317, updating the content model of `<sp>` and updating `detest.xml ` and `detest_xml_relaxng.log` accordingly.


sydb (changes requested):
These changes look perfect (in that they do exactly what Council wanted). However, the prose of the Guidelines still needs to be updated.

PR #2790, opened 2025-09-15
Allow `address` in `div` by adding it to `model.divTopPart` and `model.divBottomPart`

bleekere:
#2460

PR #2818, opened 2025-09-21
avoid ambiguous content in `<post>`

sydb:
The recently merged #2790 introduced an ambiguous content model problem for `<post>`, because its content refers to both model.phrase (of which `<address>` is a member) and to model.divBottom (of which `<address>` is newly a member) within the same `<alternate>`.

The obviously correct way to fix this (IMHO) is to use `<classRef key="model.divBottom except="address">`, but when I tried that the entire class was dropped. (Perhaps due to TEIC/Stylesheets#454?)

Thus this PR introduces a mildly ugly hack to get around this problem: the reference to model.divBottom has been changed to a list of references to the components of model.divBottom *except* for `<address>`. Ugly, but it works.

### DIFF
--- a/P5/Source/Specs/post.xml
+++ b/P5/Source/Specs/post.xml
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-Copyright TEI Consortium.
-Dual-licensed under CC-by and BSD2 licences
-See the file COPYING.txt for details.
--->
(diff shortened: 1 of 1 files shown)


HelenaSabel (approved):

martindholmes:
Ran the tests and everything looks good; this should be merged quickly because it's breaking builds on Jenkins.

martindholmes (approved):

PR #2819, opened 2025-09-21
improve messages emitted from Makefiles

sydb:
Back in 2021 I added little “I am running this target” messages to P5/Test/Makefile. I think it has made debugging a lot easier, and has not caused any problems (that I am aware of). So I did the same thing with P5/Exemplars/Makefile and P5/Makefile itself.
I did not do this to Source/Guidelines/en/Images/Makefile both because it only has 3 non-interdependent targets and (mostly) because I forgot it existed.


martindholmes (commented):
I like what I see, but I'll wait till I can run it before signing off. Also interested to see if it adds any noticeable time to the build.

martindholmes:
Works great, no detectable impact at all out speed.

Issue #2816, opened 2025-09-19
Modify points attribute to allow third coordinate in teidata.point


jamescummings:
Jun Ogawa raised an interesting problem (in successive conference presentations) that there needs to be a way to describe surfaces and zones in three dimensions. 
Currently the [points attribute](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.coordinated.html#tei_att.points) allows 3 to infinity [teidata.point](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-teidata.point.html) values. Although I think it may need more fully specifying, he and I both came up with the idea of adding an _optional_ third coordinate to the X,Y coordinates of teidata.point to enable the Z axis to be recorded.   

Currently the content model of teidata.point is: 
~~~xml
 <dataRef name="token" restriction="(-?[0-9]+(\.[0-9]+)?,-?[0-9]+(\.[0-9]+)?)"/>
~~~
which shouldn't be too hard to change to have an optional third point.

trishaoconnor:
Thank you @jamescummings and Jun Ogawa for raising this very interesting issue. This certainly seems like a very important consideration and your recommendation for adding an optional third coordinate to better describe surfaces and zones in three dimensions would potentially benefit many members of both the TEI and EpiDoc communities. The content model of `teidata.point` does seem to be a straightforward scenario. Hopefully testing will prove it to be so!

martindholmes:
+1 from me. I think we will  need to explain what we assume this third coordinate is relative to; in other words, where is 0 0 0. In the case of a flat surface, we basically assume top left, but there's less of an obvious anchor point in a three-dimensional system. Can we have some examples please?

junjun7613:
Thanks @jamescummings ! I have been considering 3D encoding in TEI (including a poster presentation at TEI2023), and I am convinced that there must be several different solutions to this issue. However, after the discussions at this year’s TEI conference with @jamescummings , @ebeshero and @martindholmes , I have come to regard the method using points attribute as the most appropriate. Would it be appropriate for me to attach here the article I am currently writing for submission to jTEI regarding this issue?

trishaoconnor:
@junjun7613 it would be great if you could attach your article if you wouldn't mind sharing it with us. It would be very helpful if you have a good 3d encoding example that we could cite in the Guidelines.

I've tested adding the optional z value and it passes local tests so I'm creating a pull request for others to review.

junjun7613:
[...]

PR #2821, opened 2025-09-23
Allowing optional third coordinate to `teidata.point` for #2816

trishaoconnor:
This PR addresses #2816 which requested allowing an optional third coordinate to `teidata.point`. I've tested the changes locally and it passes tests and I would welcome more eyes on this to check that all is well. 

It would also be helpful to include a good example of 3d encoding as well in this PR and Jun Ogawa will hopefully be able to supply us with one.

### DIFF
--- a/P5/Source/Specs/teidata.point.xml
+++ b/P5/Source/Specs/teidata.point.xml
@@ -5,7 +5,7 @@
   <desc versionDate="2010-10-17" xml:lang="en">defines the data type used to express a point in cartesian space.</desc>
   <desc versionDate="2024-08-08" xml:lang="ja">デカルト空間内の点を表現するデータ型を定義する。</desc>
   <content>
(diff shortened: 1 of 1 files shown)


sydb (approved):
The change @trishaoconnor has made seems perfectly reasonable, the regex looks correct, and it directly addresses the original request. Thus an approval.

But there are two huge caveats:

1. Having thought about it a bit more, I do not think this is the right way to do this. I think we can provide users further, more useful constraint, on the theory that almost no one wants both 2D and 3D points _on the same attribute_.
2. Even if this is the way Council decides to address this feature request, prose describing the datatype, prose discussing how to use it, and examples are all in order.

As for #​1 I will plan to submit a PR to demonstrate what I have in mind.

martindholmes:
@sydb I think I disagree with this: 

> almost no one wants both 2D and 3D points on the same attribute.

The `<gml:pos>` element `child::text()` contains a coordinate tuple which may have two or three coordinates. Since coordinate pairs or triples are almost always processed by tokenization, it's trivial to know how many you have. I don't see this as a problem at all.

PR #2824, opened 2025-10-02
restored a <sch:pattern> wrapper around global Schematron variables (…

rvdb:
…otherwise they would be skipped from the resulting compiled XSLT version used for validation in Oxygen)

This <sch:pattern> wrapper had been removed in subsequent commits.

### DIFF
--- a/P5/Exemplars/tei_jtei.odd
+++ b/P5/Exemplars/tei_jtei.odd
@@ -1875,23 +1875,25 @@
       </div>
       <div>
         <schemaSpec ident="tei_jtei" start="TEI" defaultExceptions="http://www.tei-c.org/ns/1.0 eg:egXML">
-	  <constraintDecl scheme="schematron">
+          <constraintDecl scheme="schematron">
             <sch:ns prefix="sch" uri="http://purl.oclc.org/dsdl/schematron"/>
             <sch:ns prefix="tei" uri="http://www.tei-c.org/ns/1.0"/>
             <sch:ns prefix="xs" uri="http://www.w3.org/2001/XMLSchema"/>
             <sch:ns prefix="xsl" uri="http://www.w3.org/1999/XSL/Transform"/>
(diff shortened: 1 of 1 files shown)


rvdb:
@sydb This is the proper resolution for preserving global Schematron variables, IMO. The other suggestion I made, adjusting `teiodds.xsl` so it would wrap `<tei:constraintDecl>` child elements in `<sch:pattern>` in the resulting RNG output would be too broad. When testing locally, it appeared the `tei:constraintDecl/sch:ns` declarations from the ODD file would be invalid when transformed to `sch:pattern/sch:ns` in the resulting RNG file. Hence this tweak to the Schematron fragments in the ODD file, which solves the issue.

martindholmes (commented):
I wonder if the best way to get the current TEI version is from Jenkins. Shouldn't there be somewhere on the main website to get this info?

rvdb:
@martindholmes Is anything blocking this PR?

martindholmes (approved):

martindholmes:
@rvdb I was wondering if you might respond to my question, but it doesn't matter. The TEI site is actually down more frequently than Jenkins.

martindholmes:
@rvdb I was wondering if you might respond to my question, but it doesn't matter. The TEI site is actually down more frequently than Jenkins.

rvdb:
@martindholmes Thanks for merging! And apologies, I had probably parked your question as a stylesheets issue, but have now realized 1) it is mentioning the `$tei.version.url` Schematron variable in the jTEI ODD file, and 2) that I seem to have asked myself the same question before :-) (see the comment at https://github.com/TEIC/TEI/blob/dev/P5/Exemplars/tei_jtei.odd#L1891).

Apparently, when editing a jTEI article in Oxygen, repeated calls to http://www.tei-c.org/release/doc/tei-p5-doc/VERSION (each time Schematron validation is triggered) seemed to hit a "Too Many Requests" error. That seemed not to be the case for the Jenkins version of that file.

Issue #2832, opened 2025-10-15
p5subset.xml is invalid against p5odds.rng


martindholmes:
There are over 200 errors, but they're all caused by importing `<biblStruct>` elements from the bibliography, but throwing away their contents (only their `@xml:id`s are required). If these were imported instead as `<bibl>` elements (as the existing `<bibl>` elements in the bibliography are), the problem would be solved and the file would be valid.

martindholmes:
See PR #2833.

YvesMarcoux:
I am also getting two errors caused by embedded `<egXML>` elements (lines 65179 and 65205).
The other 215 errors are caused by `<bibStruct>` elements, as described above.
So total of 217 errors.

sydb:
@YvesMarcoux! Wonderful to hear from you! (It never occurred to me you would be playing with TEI, let alone posting on an issue. : -)

Against what schema are you trying to validate the p5subset.xml file? I am betting tei_all or tei_odds, as in those schemas an `<egXML>` is not permitted inside an `<egXML>` (I do not know why not). But the p5odds schema (about which @martindholmes is complaining) explicitly does allow `<egXML>` elements to nest. See #2380.

I wonder if I address @martindholmes’ concerns on that ticket if I can drum up some support for allowing `<egXML>` in `<egXML>` in tei_all (and thus tei_odds).

martindholmes:
I have nothing against nesting egXMLs, It's just that they're awkward to process into nice documentation.

YvesMarcoux:
> [@YvesMarcoux](https://github.com/YvesMarcoux)! Wonderful to hear from you! (It never occurred to me you would be playing with TEI, let alone posting on an issue. : -)

@sydb Nice to hear from you too! I never was very far from the TEI, but I guess it didn't show... Also, contrary to now, I've never had yet to come to grips with ODDs and their chaining.

> Against what schema are you trying to validate the p5subset.xml file? I am betting tei_all or tei_odds, as in those schemas an `<egXML>` is not permitted inside an `<egXML>` (I do not know why not). But the p5odds schema (about which [@martindholmes](https://github.com/martindholmes) is complaining) explicitly does allow `<egXML>` elements to nest. See [#2380](https://github.com/TEIC/TEI/issues/2380).

Thanks for you answer, you're dead on: I was validating against tei_odds, hence my two extra errors. Thanks for the pointer to #2380.

[...]

PR #2833, opened 2025-10-15
Fix for issue #2832: make p5subset.xml valid against p5odds.rng.

martindholmes:
This just converts `<biblStruct>` elements into `<bibl>`s for the sake of validity.


lb42 (commented):
A valid bibl is definitely preferable to an invalid biblstruct

martindholmes:
@sydb @HelenaSabel Any objection to this?

PR #2834, opened 2025-10-18
update example

sydb:
I added a new event entry (for AFL811, which I did not know about when I generated original created) to event example, and also deleted an unused `<place>` from same example.


hennyu (approved):
The modified example looks good; Martina and me tested this successfully.

martinascholger (approved):

Issue #2649, opened 2025-01-04
Missing bibliography item


martindholmes:
The post.xml spec file includes this:

`Element adopted from the <ref target="#TEIcmc2012">DeRiK schema</ref>.`

which is presumably supposed to point to an item in the bibliography for the DeRiK schema, but that item isn't there.

martinascholger:
Deutsches Referenzkorpus zur internetbasierten Kommunikation (DeRiK), mentioned in https://journals.openedition.org/jtei/476.  Good example for the mentoring session.

PR #2836, opened 2025-10-24
Uhk 2649 - added missing bibliography item

hennyu:
Did this together with @martinascholger in a mentoring session addressing issue #2649


ebeshero:
The new / revised entries look good to me!

Issue #2838, opened 2025-10-29
require alternate attrs of `<dataRef>`


sydb:
The `<dataRef>` element allows the ODD writer to specify the datatype being referred to in one of three ways:

- `@key` referring to a TEI datatype
- `@name` referring to a W3C XSD datatype
- `@ref` referring to any other datatype (presumably user-defined, probably in Jeni Tennison’s Datatype Library Language, but we do not say _anything_ about how to do this, we barely even mention that you can)

Those three attributes are intended to be, and are, mutually exclusive. You cannot use both `@key` and `@name` on the same `<dataRef>`. Good. The problem is that not only are they mutually exclusive, they are optional. They should not be, one of them should be required. That is, a `<dataRef>` — an element that exists to reference a datatype — that does not have any attribute that says to which datatype it refers, is broken.

The attributes are currently defined as follows (here dropping the `<gloss>` and `<desc>` elements):
~~~xml
    <attList org="choice">
      <attDef ident="key">
        <datatype><dataRef key="teidata.xmlName"/></datatype>
      </attDef>
      <attDef ident="name">
        <datatype><dataRef key="teidata.xmlName"/></datatype>
      </attDef>
      <attDef ident="ref">
        <datatype><dataRef key="teidata.pointer"/></datatype>
      </attDef>
    </attList>
~~~
which is wrong because the default value of `attDef/@usage` is "opt". Thus this results in `( @key? | @name? | @ref? )` which gives the user a choice between three _optional_ attributes, which means she is allowed to use _no_ attribute.
Instead this should be
~~~xml
    <attList org="choice">
      <attDef ident="key" usage="req">
        <datatype><dataRef key="teidata.xmlName"/></datatype>
      </attDef>
      <attDef ident="name" usage="req">
        <datatype><dataRef key="teidata.xmlName"/></datatype>
      </attDef>
      <attDef ident="ref" usage="req">
        <datatype><dataRef key="teidata.pointer"/></datatype>
      </attDef>
    </attList>
~~~
which would result in `( @key | @name | @ref )`, which would correctly require that one and only one of the three attributes be present on a `<dataRef>`.

martindholmes:
Fix is merged, but Council agrees that we should add a test for this.

sydb:
Fixed months ago in #2867.

PR #2839, opened 2025-10-29
Require that one and only one of the referring attrs on `<dataRef>`

sydb:
This address #2838.


martindholmes (approved):
I wonder if it's worth adding a test for this.

bleekere (approved):

bleekere:
Passes local tests, too, so afaic this can be merged. Though it would be good to think about Martin's comment: add a test for this? Do we want to discuss that first, or can we merge and open a new issue for the test?

Issue #2850, opened 2025-12-17
attRef/@name should be replaced with @key for consistency with elementRef


HelenaSabel:
See https://github.com/TEIC/TEI/issues/2282#issuecomment-3292757277

sydb:
- `<classRef>` — uses `@key`
- `<elementRef>` — uses `@key`
- `<macroRef>` — uses `@key`
- `<moduleRef>` — uses `@key`
- `<specGrpRef>` — uses `@target`, but points to an `@xml:id`, so that makes sense
- `<listRef>` — does not have a referring attribute (it contains refering elements)
- `<attRef>` — **uses `@name`**
- `<dataRef>` — uses `@key` (it also has `@name` to refer directly to an XSD datatype)

I am in favor of deprecating `attRef/@name` in favor of `attRef/@key`. Yes, it is a backwards-incompatible change that some might think is merely a cosmetic change. But it is not just cosmetic — the truth is the thing you use to refer to an attribute is **not** its name, it is its `@ident`. Its name might also include a namespace (if there is an `@ns` in force) or might be entirely different (if an `<altIdent>` is used).

martindholmes:
100% agree on `attRef/@name`. But I don't agree that `specGrpRef/@key` makes sense; surely it should be `@ref` or `@target` or one of the other pointer attributes? The pattern is presumably that keys point to idents and refs point to ids.

sydb:
Oh, error on my part, @martindholmes. `<specGrpRef>` uses `@target`, not `@key`. (Hence it makes senses.) I have updated it in table, above. (Which means anyone reading your comment immediately above is going to think you are on drugs. :​-​)

PR #2852, opened 2025-12-31
add (and use) `@key` of `<attRef>`

sydb:
This is intended to address  #2850.

Needs to be built against the "sydb_795_attRef_key" branch of the Stylesheets or for TEIC/Stylesheets#796 to be merged first.


sydb:
Note: The fact that the tests here fail is uninteresting — this branch needs to be built using the "sydb_795_attRef_key" branch of the Stylesheets.

Issue #2744, opened 2025-08-12
Better explain the use of attributes that do the same thing


joeytakeda:
This arises out of the #2653, which adapted the remark for `@key` and `@ref` to discourage the simultaneous use of `@ed` and `@edRef`.  I'm wondering if we should have a principled approach -- or at least some better discussion -- of these kinds of attributes: attributes that do more-or-less the same thing, but one points specifically at a thing and one uses a token. 

A non-comprehensive list of these (from a quick look through [Appendix D](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/REF-ATTS.html)): 

* `@ed` and `@edRef`
* `@key` and `@ref`
* `@rend` and `@rendition`
* `@lemma` and `@lemmaRef`
* `@script` and `@scriptRef`
* `@scribe` and `@scribeRef`
* `@unit` and `@unitRef`

By and large, the guidance has been "We don't know what it means if both are used, but if you want to use both, then document what it means to you," but how this has been expressed varies: sometimes using schematron, sometimes remarks, and sometimes saying nothing at all.

Of course, these may not all be the same kind of thing, but for ones that are, it would be nice to have some sort of broad guidance about these kinds of attributes (and possible some sort of term for the kinds of pairings, but I can't think of a nice name for them) that outlines why the GLs have both, that the GLs have no real guidance about what it means to use both, and some guidelines about how encoders ought to document it if they decide to use both.

jamescummings:
I would add to your list by suggesting that with rend&rendition there is also the style attribute. I'd take a harder line and say that in customising the TEI it is the obligation of those doing so to do certain things. Decide between ed and edRef, between rend/style/rendition, between key and ref, to give value lists for any teidata.enumerated attribute that doesn't have a valList, etc. TEI customisations Which don't do this and more aren't non-confirmant, just not as finished as they should be. ;-) 

Where there is a choice between a URI-based attribute and a magic token meaningful only to the project, I'm off the opinion that the former should be preferred by protects unless they have good reason not to. Call me a pointing purist. ;-) 

(I am, of course, also a complete hypocrite!)

HelenaSabel:
## Overview
- `@ed` and `@edRef` (`att.edition`): definition clear, remark about use
- `@key` and `@ref` (`att.canonical`): definition remarks (same as `@ed` and `@edRef` and a specific one) about use 
[...]

PR #2853, opened 2026-01-03
improving remarks describing attrs with similar semantics

sydb:
First pass at applying changes agreed to on 2025-09-15 on ticket #2744. I am sure there is more to do, for att.measurement in particular.


P5/Source/Specs/att.global.rendition.xml
```diff
@@ -130,22 +121,27 @@ the TEI header.</p>
...
+        <p>The <att>rendition</att> attribute is used in a very
+        similar way to the <att>class</att> attribute defined for
+        XHTML but with the important distinction that its function is
+        to describe the appearance of the source text, not necessarily
+        to determine how that text should be presented on screen or
+        paper.</p>
+        <p>If <att>rendition</att> is used to refer to a style
+        definition in a formal language like CSS, it is recommended
+        that it not be used in conjunction with <att>rend</att>.
+        Where both <att>rendition</att> and <att>rend</att> are
```
raffazizzi: @sydb I think we're supposed to remove this language about rendition overriding rend.

raffazizzi (changes requested):

joeytakeda (approved):
Looks good to me!

Issue #2306, opened 2022-06-15
tagdocs elements are available in silly contexts


martindholmes:
The `<elementSpec>` element (to take one example) is available inside 126 different elements, which include `<u>`, `<witness>`, `<epigraph>`, and `<figDesc>`, to mention a few of the silliest. The ATOP team, in trying to figure out namespace hierarchies and other processing complexities, would really appreciate it if this could be cleaned up. It's surely arguable that the only logical parents of `<elementSpec>` are other tagdoc elements, specifically `<specGrp>` and `<schemaSpec>`. Similar issues apply to `<classSpec>` and other members of `model.oddDecl`, which is a member of `model.inter`, leading to its availability all over the place. Could we make better use of `model.oddDecl` to solve this problem?

sydb:
My 2¢ — Unless someone (@lb42? @jamescummings? @laurentromary?) has a good reason why model.oddDecl should be a member of model.inter, I think we should just remove it as a corrigible error. (Which means that `<specGrp>` would then have to be made a member of model.inter, as it _does_ make sense there. Now that I am thinking about it, should `<specGrp>` be a member of model.oddDecl?)

sydb:
So not having heard an objection from @lb42 nor @jamescummings, and having gotten a thumbs-up from @laurentromary, I think we are good to go in removing model.oddDecl from model.inter (and taking care of the various fallout problems).

Question — Do we think this has to be deprecated? (I think our policy says “yes”, because this is a schema change that might invalidate a users’ file (however unlikely that is), we need to deprecate. But it is a bit of a pain, because we can’t just use `@validUntil`.)

bansp:
I think it's a very bad practice to remove something skipping the agreed procedure and using ignorance as the argument/justification .
The Guidelines are released into an open world. The fact that a brilliant Council member can't imagine something to be used in a 'strange' context doesn't mean that it is not being used (maybe wrongly, according to some/most practices, but still in good faith) :-) . I believe that the Council should in such cases _always_ keep its end of the agreed (implied) bargain.

sydb:
[...]

PR #2855, opened 2026-01-04
most oddDecl elements must be descendant of `<schemaSpec>` or `<specGrp>`

sydb:
Address #2306 by:

- Issue a DEPRECATION WARNING if a `<classSpec>`, `<constraintSpec>`, `<dataSpec>`, `<elementSpec>`, `<macroSpec>`, `<outputRendition>`, or `<specGrpRef>` is not inside a `<schemaSpec>`, a `<specGrp>`, or a `<egXML>` (due to testing-swaps-namespace thing).
- Change _Guidelines_ so that all of our such elements are in fact inside `<specGrp>`
- Add a test (`<elementSpec>`-inside-`<head>`)

In 9–10 months time the deprecation warning should become an error.

Note: This is currently a DRAFT PR, as I have not yet added a note to Appendix G (deprecations). Hope to get to that tonight, but figure the rest can be looked at without that.


martindholmes:
With @sydb and @HelenaSabel : We tested this and determined that there are 981 cases that violate these constraints within p5subset.xml (spec elements are children of divs there). Therefore we chose to postpone this merge until after the February release, so that we can also add wrapping specGrps to make p5subset.xml valid (for which another ticket already exists), and check that there's no fallout from that change in the Stylesheets ODD processing.

martindholmes:
Coming back to this: it seems that there are a lot of irrelevant changes to the way the XInclude namespace is declared and used in the PR, which is a bit confusing. @sydb Was there any intention here, or is this just meaningless fallout from the use of an XML parser/serializer that does things differently? Could we remove those changes from the future PR to make it easier to evaluate?

sydb:
While I admit it is mildly annoying, the XInclude prefix change was intentional, and I do not think there is an easy way to revert it, @martindholmes. But one of the reasons I did not feel bad about making these changes is the realization that almost every line that got changed was going to have whitespace changes, anyway. (Which you already consider problematic.)
If this is really bothering you may I suggest that we either
- find someone else to review this, or
- ask @HelenaSabel & @hennyu to review the code changes, and ask you to test that it works? (Although note that there is already an `<elementSpec>` inside a `<head>` in detest.odd which is correctly flagged.)

Either way I think we really should try to get this into the upcoming release, but now that 5 months have gone by we also need to think about the deprecation period. It is currently set to end on 2026-09-03 after which an ODDdecl element in the wrong place would become an error. Also, I am wondering if the `<sch:assert>` on line 32 should have a `@role` of "nonfatal" or "warning". (In P5 we have 110 Schematron assertions & reports, of which 11 have a `@role` attr:
~~~
      6 nonfatal
      3 warning
      2 error
~~~
and there is no mention of when to use `@role` in [TCW 24](https://tei-c.org/documentation/tcw24/).)

martindholmes:
The branch apparently still has conflicts. @sydb  What puzzles me about the XInclude namespace stuff is why it's in this PR; each PR should address one issue distinctly, surely? If the XInclude declarations need changing, that should be a separate issue and a separate PR.

I also agree that the deprecation period needs to be extended to at least a year from the next release.

P5/Source/Specs/model.oddDecl.xml
```diff
@@ -30,7 +30,7 @@
       <sch:pattern id="oddDecls_in_specGrp_or_schemaSpec" abstract="true">
         <sch:rule context="$toe">
           <sch:assert test="ancestor::tei:specGrp | ancestor::tei:schemaSpec | ancestor::tei:egXML">
-            DEPRECATION WARNING: Sometime after 2026-09-03 the &lt;<sch:value-of select="name(.)"/>>
+            DEPRECATION WARNING: Sometime after 2028-03-09 the &lt;<sch:value-of select="name(.)"/>>
```
ebeshero, line 33: @sydb This makes sense to me...some of the others make me ponder...

P5/Test/expected-results/detest_odd_schematron.log
```diff
@@ -1,17 +1,17 @@
-at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:schemaSpec[1]/tei:elementSpec[7]/tei:remarks[1] —
-   Error: both the versionDate and xml:lang attributes on "remarks" are required when it is a child of "elementSpec".
+at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:head[1]/tei:elementSpec[1]/tei:desc[1] —
```
ebeshero, line 1: @sydb I have questions about this XPath: Why is the log isolating this very specific path: `//div[1]/head[1]/elementSpec[1]/desc[1]`?

P5/Test/expected-results/detest_odd_schematron.log
```diff
@@ -1,17 +1,17 @@
...
 at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:schemaSpec[1]/tei:elementSpec[10]/tei:attList[1]/tei:attDef[1]/tei:gloss[1] —
    Error: both the versionDate and xml:lang attributes on "gloss" are required when it is a child of "attDef".
-at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:schemaSpec[1]/tei:elementSpec[10]/tei:remarks[1] —
-   Error: both the versionDate and xml:lang attributes on "remarks" are required when it is a child of "elementSpec".
 at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:specGrp[1]/tei:elementRef[1] —
    An element reference is not repeatable when part of a schema specification (and thus this <elementRef> should not have @minOccurs or @maxOccurs).
 at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:schemaSpec[1]/tei:elementSpec[17]/tei:content[1]/tei:alternate[1]/tei:elementRef[1] —
    An element reference within a content model must refer to a locally defined element specification (and thus this <elementRef> should not have @source).
+at /tei:TEI[1]/tei:text[1]/tei:body[1]/tei:div[1]/tei:head[1]/tei:elementSpec[1] —
```
ebeshero, line 13: here, too: Why isolate only this path to elementSpec, as child of `div/head`?

ebeshero (commented):
@sydb I had a look at this today and didn't understand a couple of the tests. It's likely that I'm misunderstanding something, but I wonder why we need the two tests that I flagged here, for `elementSpec` as a child of `/text/body/div[1]/head[1]`?

P5/Test/expected-results/detest_odd_schematron.log
hennyu: This is something that is still hard for me to review, but could it be the case that Syd meant to create one test here that allows to check if there is any `elementSpec` that is directly a child of an element (here `<head>`) that it should not be a child of? I understood it as just one of many other possible cases and that this one was added as an example test to check whether such cases are detected at all. However, I agree with you, @ebeshero that it is probably possible to create a more general test?

hennyu (commented):
As far as I can tell it looks good to me. I agree with @martindholmes , though, that the namespace changes for the XIncludes could have been done separately, but of course, now this is cleaned up alongside the other changes in this PR.
@ebeshero still had a question about the specific test case with `elementSpec` inside `head`.

P5/Test/expected-results/detest_odd_schematron.log
sydb: Maybe because it is late and I have not gotten enough sleep lately, and thus am dimmer than usual, but I do not understand the question, @ebeshero. (That said, @hennyu has the right basic idea, above, I think.) Perhaps if I review what is going on here that will help. Maybe not, but I will give it a try, anyway.

[Note: all filepaths are relative to …/TEI/P5/Test/.]

The file we are looking at here is expected-results/detest_odd_schematron.log. It is the (expected) output from validating detest.odd against ../p5odds.isosch. Remember that detest.odd and detest.xml are files that contain deliberate errors for testing. Anyhoots, this .log file is arranged in pairs of lines. The first of each pair has the word “at” followed by an XPath followed by an em dash; the second of each pair has the corresponding error message that got flagged at the XPath on the previous line.

The particular XPath you have flagged is “/TEI[1]/text[1]/body[1]/div[1]/head[1]/elementSpec[1]” which refers to an `<elementSpec>` element on lines 74–77 of detest.odd. The text inside that `<elementSpec>` says (in part) “should be flagged as an error: elementSpec should not be in head”. And, indeed, it **was** correctly flagged as an error (well, a deprecation warning) by the new code. The reason that XPath shows up as a diff in GitHub is because that `<elementSpec>` was added to detest.odd in this PR specifically to test that the new code (also added in this PR) would correctly flag it as a problem. (I do not understand why the following line, the msg itself, is not also shown as a difference by GitHub.)

Is that helpful?

ebeshero: Yes--somehow I thought you'd made a separate Schematron rule especially for this path--at least that is what I remember wondering about. I'm pretty tired myself this week (sigh). Anyway, yes, that explains it and certainly  the general Schematron test should easily address this error.

ebeshero (approved):

P5/Source/Specs/classSpec.xml
```diff
@@ -42,13 +36,20 @@ that is a group of
...
+      </sch:pattern>
+    </constraint>
+  </constraintSpec>
   <attList>
     <attDef ident="type" mode="change" usage="req">
       <desc versionDate="2005-10-05" xml:lang="en">indicates whether this is a model class or an attribute class</desc>
       <desc versionDate="2007-12-20" xml:lang="ko">모델 부류 또는 속성 부류 여부를 나타낸다.</desc>
       <desc versionDate="2007-05-02" xml:lang="zh-TW">指出這是結構元素集或是屬性元素集。</desc>
       <desc versionDate="2008-04-05" xml:lang="ja">当該クラスがモデルクラスか属性クラスかを示す。</desc>
-      <desc versionDate="2007-06-12" xml:lang="fr">indique si c'est une classe de modèles ou une classe d'attributs.</desc>
```
martindholmes, line 51: This is inconsistent. I don't see why this period would be deleted when other periods stay in other languages. If the standard is for all descs to end with periods, or not to, then we have a lot of other work to do, but it doesn't belong on this PR. I would say reverse this, then raise another ticket for consistency in desc format.

martindholmes (approved):
With the one caveat about the deleted period, this looks good.

P5/Source/Specs/classSpec.xml
sydb: It was not changed on purpose.

Issue #2860, opened 2026-01-08
entryFree should be able to nest


martindholmes:
Arising from a discussion in the help channel of the TEI Slack: The `<entryFree>` element has a content model which allows a lot of things -- that's what it's for, really, to allow a much freer arrangement of content for contexts where `<entry>` is too constrained. However, while the `<entry>` element can nest, the `<entryFree>` element can't. This just seems like a simple oversight to me, and would be easily fixed.

PR #2861, opened 2026-01-08
For issue #2860: make entryFree nestable, and also allow it as a chil…

martindholmes:
…d of entry.

I've done this by adding it to both model.entryPart and model.entryPart.top. The rationale for making it available inside `<entry>` is this:

It's not unusual for a born-digital dictionary to be constructed around a historical print dictionary, as a sort of hybrid, where the transcription/encoding of the original source entries is accompanied by a more structured version of the data. With `<entryFree>` available in `<entry>`, the structured born-digital version of the data can also host a transcription of the source entry from which it was derived. I think this would be very useful, and I don't see any downsides.

### DIFF
--- a/P5/Source/Specs/entryFree.xml
+++ b/P5/Source/Specs/entryFree.xml
@@ -24,6 +24,8 @@
     <memberOf key="att.lexicographic"/>
     <memberOf key="att.sortable"/>
     <memberOf key="model.entryLike"/>
+    <memberOf key="model.entryPart"/>
+    <memberOf key="model.entryPart.top"/>
   </classes>
   <content>
     


laurentromary:
This seems to be a natural move and puts the two elements in coherence. The implementation is the appropriate one. Green light for me.

kdepuydt:
What would be added to this entryFree to make the distinction between the less structured content and the extracted structured content?

bansp (approved):
My take on this is very much like Laurent's: this appears to be a move towards coherence of the schema, and also (additionally or as a consequence) one that facilitates a gradual, phase-based, digitisation process, where today's `<entryFree>` may gravitate towards tomorrow's `<entry>`, depending on the given project's resources and scheduling.

Issue #2459, opened 2023-08-18
<address> should become member of att.typed 


StefanDumont:
A discussion at a workshop on TEI-XML encoding of correspondence resulted in the article ["Addresses"](https://encoding-correspondence.bbaw.de/v1/addresses.html) by Marjam Trautmann, in which several problems with `<address>` are being discussed.

One of the problems is that there is no proper way to distinguish multiple addresses on a letter or postcard, e.g. addresses of sender and receiver. At the moment this isn't possible within `<address>`, since it has no attribute `@type` and therefore some cumbersome workarounds are necessary (see [Trautmann's article](https://encoding-correspondence.bbaw.de/v1/addresses.html#c-2) for details). 

So we ask, if it is possible to make `<address>` a member of `att.typed`?

sydb:
I am not sure I grok (or agree with) the use case listed in the OP. (Nor am I sure I disagree.) But since `<address>` can obviously be repeated and categorized, and there is a user request, it should (IMHO) be a member att.typed.

But I am concerned that `@type` is not the attribute that those encoding correspondence really want. Wouldn’t the `@type` of `<address>` assert the addressing system used (e.g., by country postal system, by internal business system, etc.)? I think `@role` or `@of` would be the attribute for whether this is the address of the sender or recipient, no?

sabineseifert:
In Trautmann's article there are three cases mentioned:
1 ) sender/addressee: I also think `@role` fits better than `@type`. This means, we also need to add `<address>` to `att.naming`.
2 ) and 3 ) private address/business address and postal address/e-mail address: For these cases, `@type` seem to be fitting.

Examples for the Guidelines would be nice which I happily provide.

I am a bit puzzled by the mentioning of `@of` which I cannot find in the Guidelines?!

lb42:
It's true that this is probably overloading the use of @type -- and not the only such case. Would a typology of addresses  distinguish an address used as a destination differently from one used as a source? If so what happens when the same address originally typed as a source appears later as a destination (maybe those postcards didnt get replies, but probably others did!)  So I agree  that @type='sender' is suboptimal. But I don't think using @role is quite right either for much the same reason. Suppose the phrase "sent from" appeared in the source before the address, what would you do? Presumably that's not part of the address, so it would not go inside the element. It would ideally be a sibling label element, within some other grouping element .. probably ab or seg . And in the absence of such a label (or even in its presence) you could plausibly attach @type = sender to that grouping element.
________________________________

[...]

PR #2844, opened 2025-11-26
Add more attributes to address

rettinghaus:
This PR 

* makes address part of `att.canonical` and `att.typed`
* adds a special `@role` attribute

closes #2459

PR #2863, opened 2026-01-13
new attrs for `<address>`

sydb:
This PR is intended to address #2459.

I have added `<address>` to att.canonical & att.typed, and given it a `@role`.

The main things for reviewers to examine are the lists of sample & suggested values for `@role` and `@type`, respectively. I just made them up from thin air, and some tweaking may well be in order. (My knowledge of types of addresses is very US-centric, I am afraid.)

I wonder aloud whether we should put this `@role` in a class that is defined in ND, so only users who include the names & dates module have that attribute.


sydb:
Oy!
This PR should probably be ignored as a duplicate, and #2844 considered instead.

Issue #2695, opened 2025-04-02
spGrp should self-nest


lb42:
The element spGrp is primarily useful for the encoding of musical numbers especially when more than one speaker/singer participates. In a sense, it is divLike but it does not tesselate the text. It permits interruption of the  act-scene-speech hierarchy, like a nested text or quote.   However, unlike div, ispGrp cannot self-nest.

I propose two changes : first, the content model of spGrp should be expanded to permit spGrp as an explicit alternate along with sp and the members of  model.stageLike and model.global. Second the attribute @org should be added, with values "sequential" and "parallel", indicating whether the  components of the spGrp are to be performed in sequence or in parallel. This seems a more natural and simpler approach to the problem of representing simultaneous singing of different words by different performers than the appropriation of the @sync attribute (which is intended for speech transcription, rather than performance). 

Here's an example from a 19c light opera (Rophino Lacy's version of Cinderella) , but there are plenty of others

![Image](https://github.com/user-attachments/assets/0a1c7604-7092-4d6f-bc67-999074ea12a4)

lb42:
P.S. A better way would be to add spGrp to the att.divLike class and expand the associated valList for @org

lb42:
And here's how I would like `to` encode the above example:
```

 <stage>A Page brings a seat, which is placed next to the throne, on its R., Cinderella occupies
      it, the sisters sit on her R. H., and the fête continues. A fanciful dance by appropriate
      characters, followed by a party of Tyrolese singers and dancers, who enter dressed in the
      costume of their nation. While some dance, the others accompany them by their voices
      alone.</stage>

     <spGrp>
      <head>TYROLIENNE.</head>
       <spGrp org="parallel">  
         <sp >
       <speaker>Men.</speaker>
       <l>Whilst to joy we sing inviting, </l>
       <l>With our strains thy steps uniting </l>
       <l>With thy smiles our pains requiting </l>
       <l>Lovely maid, our eyes delight.</l>
      </sp>
      <sp>
       <speaker>Women.</speaker>
       <l>Swift as the flash </l>
       <l>That mocks the sight </l>
       <l>Thou seem'st a bird </l>
       <l>In airy flight.</l>
      </sp>
 </spGrp>
      <sp>
       <speaker>Together.</speaker>
       <l>When, home returning, </l>
       <l>   We leave these cool fountains, </l>
       <l>   In our native mountains </l>
       <l>Thy praise we'll recite.</l>
      </sp>
[...]

PR #2866, opened 2026-01-20
remodeling spGrp to self-nest for #2695

ebeshero:
See if you agree with how I modified the content model and description of `<spGrp>` and the somewhat more generalized new values I've added for `@org` on `att.divLike`.

Right now, this branch isn't passing tests because I haven't added the new source for the example to BIBL. But when I build the Guidelines schema (with `make exemplars`) I'm validating a test file with @lb42 's example on the ticket. I notice that my new values for `@org` are breaking test1 and test2 in Stylesheets so we'll need to do some repairs—but we should review them anyway.

### DIFF
--- a/P5/Source/Guidelines/en/DR-PerformanceTexts.xml
+++ b/P5/Source/Guidelines/en/DR-PerformanceTexts.xml
@@ -655,8 +655,8 @@ disguise, as in the following examples:
    <castItem><role xml:id="hh">Henry Higgins</role></castItem>
 </castList>
 <sp who="#hh">
-   <speaker>The Notetaker</speaker>
-   <p> ... </p>
+   <speaker>The Note Taker</speaker>
+   <p>Live where you like; but stop that noise.</p>
 </sp></egXML>
 <!-- GB Shaw, Pygmalion --></p>
 <p>If the speaker attributions are completely regular (and may thus be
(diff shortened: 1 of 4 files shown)


ebeshero:
This is passing tests now. @lb42 , how does this look, and do you want to do something different with the BIBL entry for your example? I'm not sure it needs more than this update to the spec if we think the chapter prose in https://tei-c.org/release/doc/tei-p5-doc/en/html/DR.html#DRSPG is strong enough.

joeytakeda:
Reviewed in Jan 20, 2026 community call (with thanks!) -- group notes that, by and large, this look great and agrees that prose may need to be updated to be a bit more explanatory (which @lb42 has offered to draft). 

However, one small quibble is the addition of both `unison` and `parallel` as values. Community Call group suggests just using `parallel` to allow for broader use (as `unison` is semantically quite similar to `parallel`, but may not be as widely applicable).

lb42:
The bibliographic record for the nice new example should probably not use the date 1830 since this is the date of first performance, not the publication date, which is unknown. And it should definitely spell the publisher's name correctly ("Hailes Lacy") . I suggest adding the serial title as well "Lacy's Acting Edition, volume 18, No. 0262" but that's not essential.

lb42:
As Joey says, it's not clear why both "unison" and "parallel" are provided, or what the distinction between them is. I would stick with just "parallel" as a third value, complementing the existing values of "composite" (no claim is made as to how child elements should be processed), and "uniform" (all children are to be precessed in sequence).  Have you a use case for "UNISON"?

lb42:
There's a lot I would rewrite in the existing DR chapter, but none of it is particularly wrong.  Perhaps the section https://tei-c.org/release/doc/tei-p5-doc/en/html/DR.html#DRSIM might be moved to follow https://tei-c.org/release/doc/tei-p5-doc/en/html/DR.html#DRSPG to show how spGrp is useful for things other than musical numbers. Certainly the Shavian  example needs to be corrected to use @org rather than @type 
~~~~
<spGrp org='parallel'  rend="braced">
 <sp>
  <speaker>Lady Utterword</speaker>
  <p>Mr Mangan!</p>
 </sp>
 <sp>
  <speaker>Captain Shotover</speaker>
  <p>Whats that?</p>
 </sp>
 <sp>
  <speaker>Hector</speaker>
  <p>Ha! ha! Do. Do.</p>
 </sp>
 <sp>
  <speaker>Ellie</speaker>
  <p>Please dont.</p>
 </sp>
 <stage type="delivery">in consternation</stage>
</spGrp>
~~~~

martindholmes (changes requested):
I agree with @lb42 and the community group that the value "unison" is essentially no different from "parallel", unless the latter is supposed to suggest speaking different text while the former means the same text; but if speakers are speaking the same text in unison, that would be a single speech, surely?

ebeshero:
@lb42 @joeytakeda @martindholmes I've implemented your recommended revisions for this branch, including:
* removed the "unison" value for `@org` (I agree, not necessary, just an idea I had while initially working on the ticket.)
* rearranged DR to move #DRSIM after #DRSEG. (That works nicely, yes.)
* fixed the BIBL entry for the Lacy's play edition.
* While in the neighborhood of DR, saw some ancient discussion over references to G. B. Shaw plays. Shaw is an old favorite of mine, so I'm familiar with his punctuation quirks (really he was being as minimalist as possible), and just updated both BIBL entries to Shaw to the proper published Constable editions.  I just looked this up, and apparently [Shaw called apostrophes "uncouth bacilli"](https://mindthegap1.substack.com/p/uncouth-bacilli) and wanted them to be eliminated from English writing! (He didnt win that–see what I did there?) but his plays are properly published without such "uncouth bacilli".

P5/Source/Specs/spGrp.xml
```diff
@@ -3,29 +3,86 @@
...
           <elementRef key="sp"/>
+          <elementRef key="spGrp"/>
           <classRef key="model.stageLike"/>
-        </alternate>
-      
+        </alternate>   
     </sequence>
   </content>
+  <exemplum xml:lang="en">
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-spGrp-egXML-RophCind" source="#RphCnd">
```
sydb: Any reason not to have the `@xml:id` follow the usual pattern? (**All** of the 1,576 other `<egXML>` IDs that start with "gi-" end with a disambiguation component comprising a hyphen and two lowercase letters, a–z.)

P5/Source/Guidelines/en/DR-PerformanceTexts.xml
```diff
@@ -744,6 +744,72 @@ Levant</head>
...
+      speech,  and the four speeches with which it is to be associated are 
+      grouped by means of the <gi>spGrp</gi> element. The <att>rend</att> 
+      attribute is used to specify the fact that the three speeches were 
+      grouped by the brace in the copy text. Producing a readable version 
+      of the text which simulates the original printed effect may however
+      require more complex markup and processing.
+    </p>
+    <p>More powerful and more precise mechanisms for temporal alignment are
+      defined in chapter <ptr target="#TS"/>.  These would be appropriate for
+      encodings the focus of which is on the actual performance of a text
```
sydb: Wouldn’t “… for encodings which focus on the actual performance …” say the same thing in a simpler way?

P5/Source/Specs/spGrp.xml
```diff
@@ -3,29 +3,86 @@
...
+      
+      <stage>A Page brings a seat, which is placed next to the throne, on its R., Cinderella occupies
+        it, the sisters sit on her R. H., and the fête continues. A fanciful dance by appropriate
+        characters, followed by a party of Tyrolese singers and dancers, who enter dressed in the
+        costume of their nation. While some dance, the others accompany them by their voices
+        alone.</stage>
+
+      <spGrp>
```
sydb, line 36: I am concerned about the implications of the inconsistent use of space in this example.

- 18 of 42 end tags are preceded by an extraneous space. I do not think this alone is a problem, but since I am writing a comment, anyway …
- One start tag has an extraneous space before the TAGC delimiter (‘>’).
- Four metrical line elements (`<l>`) start with 1–3 occurrences of U+00A0. Is this standard encoding practice for indented lines? (I have always used `@rend` on `<l>` to indicate the indentation, but `<space>` exists for this purpose, and is quite reasonable, too. Use of U+00A0 does not strike me as _wrong_, but it does strike me as fragile, and thus perhaps not something we should be recommending.)

sydb (approved):
I do not feel qualified to review the changes to BIB.

P5/Source/Specs/spGrp.xml
ebeshero: Okay, fine! I'm changing it. :-) Never thought to look at all of them, but was just pairing the egXML id to the id I made for the new BIB entry.

P5/Source/Guidelines/en/DR-PerformanceTexts.xml
ebeshero: Yep. Revised.

P5/Source/Specs/spGrp.xml
ebeshero: Thanks for catching this, @sydb ! I went through with this tidy regex `(\S) (</)` and located 27 extraneous spaces before the end tags, all of which I've now removed. And I carefully removed the ones at the start tags. (I agree, it's not a good idea to suggest we handle indentation that way, and I'm pretty sure it doesn't matter for this example, but if someone objects we can put a nice `@rend` in there instead.)
Fixed the weird space in the start tag as well.

ebeshero:
@martindholmes Did the build for this come out okay in PDF / HTML? I can rename the branch if that will help, and also make sure I can build from Jenkins a little later today.

martindholmes (approved):
Test build on Jenkins works well; PDF looks good.

PR #2867, opened 2026-01-31
add test for #2838 (or, more precisely, for #2839, the fix for #2838)

sydb:
Add check to detest.xml¹ that a `<dataRef>` element with **no** attributes, or any **two** of `@key`, `@name`, & `@ref` is flagged as an error.

**Note**
¹ Because adding it to detest.odd would break the build, as these errors are caught be RELAX NG, not by Schematron.


sydb:
Well _that’s_ an interesting test failure. Not sure what to do about it off the top of my head. Have to go offline to walk dogs, hope to look in ~1 hr.

martindholmes (commented):
Tested, and reviewed. Looks good.

PR #2870, opened 2026-02-16
improved prose for remarks for `@role` of `<address>`

sydb:
Title says it all.


martinascholger:
Looks good to me and is definitely an improvement.

PR #2872, opened 2026-03-01
Delete extraneous "the".

sydb:
Title kinda says it all.
This does not need a separate reviewer and assignee IMHO. If one person (assignee or anyone else) looks & approves he or she should just merge.


ebeshero:
The "the" sure was extraneous! :-)

PR #2885, opened 2026-04-16
Clarifications on the use of the word 'prefix' for issue #2001.

martindholmes:
I have limited the changes to two clarifications, one in the spec file for `<prefixDef>` and one in the chapter prose, since I believe the likelihood for confusion is small, and since the scheme name actually does function as a prefix, and the RFC itself uses the word "prefix".


jamescummings (approved):
Seems straightforward clarification to me

sydb (approved):

PR #2888, opened 2026-04-23
proposed att.linguistic.dependency class for lightweight dependency syntax description

### LINKED ISSUE #2889: add att.linguistic.dependency class for lightweight description of syntactic dependency structures
This ticket accompanies PR ##2888 that modifies part of chapter 18 (AI-AnalyticMechanisms.xml) and adds a new attribute class (att.linguistic.dependency.xml).

The content of the PR has been co-authored by Andreas Nolda (@nolda ) and myself, and incorporates feedback from the LingSIG (it was presented and in discussed at our virtual and F2F meetings, as well as in separate e-mail exchanges) and from a wider audience at the TEI Conference in Kraków in 2025 (the latter was responsible for the autumnal bunch of modifications, including a changed approach to attribute naming).

### Extremely brief summary

The PR proposes a new attribute class called `att.linguistic.dependency` with three attributes that, together, provide a lightweight means for annotating syntactic dependency relations between the contents of the word-level elements `<w>` and `<pc>`. This new attribute class extends the existing attribute class att.linguistic for lightweight linguistic annotation, complementing the latter's attributes `@lemma`, `@lemmaRef`, `@pos`, `@msd`, and `@join`. It thereby allows for encoding not only the output of lemmatisers, part-of-speech taggers, and morphological analysers, but also that of dependency parsers in a straightforward, simple way. Crucially, the proposed enhancement can encode the existing Universal Dependencies descriptions (which come in a plain-text format known as CoNLL-U), and we have a proof-of-concept script that does that, but it can also be applied to a description of (we believe) any traditional dependency-grammatical relationships.

The summary here is brief so that attention can be focused on the parts that are meant to do the actual explanatory and illustrative job:
* proposed addition to ch. 18: https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/release/doc/tei-p5-doc/en/html/AI.html#AILADEP

* proposed new spec: https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/ref-att.linguistic.dependency.html

Note that the links above point to versions of the Guidelines that recognise the added attribute class, so they use, at the top, PIs different than those from regular Guidelines:
```xml
<?xml-model href="https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
<!--Note the above PI substitution that enables us to use the added spec file. The original PI is below:
<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>-->
```

bansp:
This PR is connected to issue #2889 .

It proposes changes to two existing parts of the Guidelines: `Guidelines/en/AI-AnalyticMechanisms.xml` and `Specs/att.linguistic.xml`, and adds `Specs/att.linguistic.dependency.xml` along several graphics.

Note that, at this point, the PR is not going to pass the checks, because of the added Spec. The modified/added docs (`Guidelines/en/AI-AnalyticMechanisms.xml` and `Specs/att.linguistic.dependency.xml`) begin with the following PI:

```xml
<?xml-model href="https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
<!--Note the above PI substitution that enables us to use the added spec file. The original PI is below:
<?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>-->
```

Here is the relevant pocket of Mr. Jenkins of Paderborn: https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/

And the particular relevant links (which work thanks to the changed nvdl PIs) are:
* proposed addition to ch. 18: https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/AI.html#AILADEP
* proposed new spec: https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/ref-att.linguistic.dependency.html


P5/Source/Guidelines/en/Images/dependency4.png
sydb, line 1: I suspect this figure would be easier to read if the words lined up under each index number were left-aligned. (The human eye is much better at “popping” back to the same spot it started … it does not like to have to fish around for where a line starts.) This is also true of some other figures, but in those the columns are so short it really doesn’t matter.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
     <w lemma="forth" pos="av" orig="sorth">forth</w>
   </egXML></p>
 </div>
+
+<div type="div3" xml:id="AILADEP">
+<head>Syntactic dependency relations between word-level elements</head>
```
sydb: Is a dependency relationship always between two words (or abbreviations, or punctuation, or whatever), or can it be among three or more? (In which case the “between”s on this line and two lines later should perhaps be “among”s?)

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+<p>For the purpose of encoding linguistic annotations of syntactic dependency
+relations between word-level elements, the Guidelines provide attributes that
```
sydb: As a general rule, we refer to the TEI Guidelines themselves as “these Guidelines”, but also as “the TEI Guidelines” often enough. This is not a hard-and-fast rule, though, and is not documented in the [style guide](https://tei-c.org/documentation/tcw24/).

Occurences of “guidelines” and preceding word (case insensitive):
~~~
    243 these guidelines
     68 tei guidelines
     36 the guidelines
      6 p5 guidelines
      5 present guidelines
      1 der guidelines
      1 encoding guidelines
      1 formulating guidelines
      1 libraries guidelines
      1 p3 guidelines
      1 those guidelines
      1 title guidelines
~~~
(This list was extracted from text nodes only, i.e. occurrences in comments or PIs were ignored. I _think_ I used this branch’s version, but I am not sure.)

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+<mentioned>books</mentioned>, the latter node is said to be the
+dependent node and the former its head node.</p>
+<p>An equivalent rendition of this graph is given below, where the direction of
+the arrowless arcs is implicitly given by the vertical dimension. For
+convenience, the horizontal positions of the nodes are made explicit in terms of
+numerical, 1-based, token indices.
+<figure>
+<graphic width="200px" url="Images/dependency2.png"/>
+</figure>
+As it is implied from this rendition, the node with token index 2 (labelled with
```
sydb: This phrasing (“As it is implied from this rendition, …”) feels clumsy to me, but I am not sure I can do much better.

Perhaps one of
- “As implied from this rendition …”
- “This rendition (correctly) implies that …”
- “This rendition makes it clear that …”
- “It is clear from this rendition …”
- “The node with token index 2 (…) is clearly the …”
or something like that?

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+<mentioned>buys</mentioned>) is the <term>root</term> of the dependency graph:
+from this node, every other node in the graph can be reached on a path of arcs
```
sydb: I think maybe “every other node in the graph can be reached by following a path of arcs from top to bottom”

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+by following them from top to bottom.</p>
+<p>Using the <att>depN</att> and <att>depR</att> attributes defined by the <ident
+type="class">att.linguistic.dependency</ident> class, the dependency relations
+in the above example can be encoded in TEI XML as follows:
```
sydb: I would prefer to avoid the construct “TEI XML” and just say “… may be encoded as follows:” or “… can be encoded in TEI as follows:”, but in truth we use “TEI XML” in the Guidelines _a lot_. So I think the horse has already left this barn.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+</egXML>
+Here, <att>depN</att> attributes represent the indices of the tokens marked up
+with <gi>w</gi> and <gi>pc</gi> elements. Dependency relations are encoded by
+means of <att>depR</att> attributes of the <gi>w</gi> and <gi>pc</gi> elements
+in terms of a colon-separated pair consisting of the token index of their head
+node and a label for the type of the dependency relation.</p>
+<p>A widely used notation for annotations of dependency relations and other
+linguistic information is the plain-text, column-based <term>CoNLL-U</term>
+format, used by <ref target="https://universaldependencies.org/">Universal 
+Dependencies</ref> (UD) as well as by related frameworks. A
```
sydb: Either “(UD)**,** as well as by related frameworks.” or “(UD) and related frameworks.”, no?

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+<eg xml:space="preserve">1	She	she	PRON	_	_	2	nsubj	_	_
+2	buys	buy	VERB	_	_	0	root	_	_
+3	books	book	NOUN	_	_	2	obj	_	SpaceAfter=No
+4	.	.	PUNCT	_	_	2	punct	_	_</eg>
+The first column gives the indices of the tokens in the current sentence, the
+second column the tokens themselves, the third column the corresponding lemma,
+the forth column a UD part-of-speech tag, the seventh column the token index of
+the syntactic head of the current token, and the eighth column a label of the
+type of the dependency relation between the token and its head. The very last,
+tenth, column can be used for miscellaneous information, such as whether or not
```
sydb: I do not understand “… there is space to add after the token when joining them”. Maybe “such as whether or not a space should be inserted between this and the following token when joining them”?

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+  <pc depN="4" depR="2:punct" pos="PUNCT" lemma=".">.</pc>
+</s>
+</egXML>
+As in the previous example, <att>depN</att> attributes represent the indices of
+the tokens marked up with <gi>w</gi> and <gi>pc</gi> elements; the
+index <val>0</val> for the <soCalled>virtual</soCalled> root node is left
+implicit. Dependency relations are again encoded by means of <att>depR</att>
+attributes of the <gi>w</gi> and <gi>pc</gi> elements. Note that the <gi>w</gi>
+element with the token <mentioned>buys</mentioned> in its content now also has a
+<att>depR</att> attribute. Part-of-speech tags and lemmas are given as values
```
sydb: “… given as the values …”, no?

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+  <w depN="1" depR="2:nsubj" depE="2:nsubj|4:nsubj" pos="PRON PRP" msd="Case=Nom|Gender=Fem|Number=Sing|Person=3|PronType=Prs" lemma="she">She</w>
+  <w depN="2" depR="0:root" depE="0:root" pos="VERB VBZ" msd="Mood=Ind|Number=Sing|Person=3|Tense=Pres|VerbForm=Fin" lemma="buy">buys</w>
+  <w depN="3" depR="4:cc" depE="4:cc" pos="CCONJ CC" lemma="and">and</w>
+  <w depN="4" depR="2:conj" depE="0:root|2:conj" pos="VERB VBZ" msd="Mood=Ind|Number=Sing|Person=3|Tense=Pres|VerbForm=Fin" lemma="sell">sells</w>
+  <w depN="5" depR="2:obj" depE="2:obj|4:obj" pos="NOUN NNS" msd="Number=Plur" lemma="book" join="right">books</w>
+  <pc depN="6" depR="2:punct" depE="2:punct" pos="PUNCT ." lemma=".">.</pc>
+</s>
+</egXML></p>
+<p>As in the CoNLL-U format, the enhanced dependency graph <!-- with additional
+dependency relations --> is represented by pipe-separated list values of
```
sydb: Hmmm …
Maybe 
~~~xml
is represented using <att>depE</att> attributes. Each <att>depE</att> attribute
holds a list of pipe-separated values, each of which is a colon-separated pair: the
token index of the head node and a dependency relation type.
~~~
or some such?

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1190,168 @@ sake of readability).
...
+<att>depE</att> attributes. The verb form <mentioned>sells</mentioned> with
+token index 4, for instance, has <val>0:root|2:conj</val> as its <att>depE</att>
+value, to be interpreted as being both the dependent node of a root relation to
+the <soCalled>virtual</soCalled> root node with index 0 as well as the dependent
+node of a conjunct relation to the node with index 2 (the verb form
+<mentioned>buys</mentioned>). Concurrent part-of-speech tags are given as
+space-separated list values of <att>pos</att> attributes, and <att>msd</att>
```
sydb: Whoa!
Since when do `@pos` attrs allow spaces or multiple values? (Answer: at least for the last 9 years.)

It turns out the datatype of `@pos` is teidata.text, which seems overly permissive to me. Just sayin’. 

Anyway, as far as I can tell, the _Guidelines_ never use nor define the semantics of multiple values on `@pos`. I am not here suggesting that the proposed use of, e.g., "NOUN NNS" to mean something like “the NOUN pos and the NNS pos are equivalent, and both are applicable here” is wrong, only that if that is what multiple space-separated values on `@pos` means then

1. The datatype of `@pos` should be something like `<datatype minOccurs="1" maxOccurs="unbounded"><dataRef key="teidata.word"/></datatype>`; and
2. the semantics should be explained somewhere (probably in the discussion of `@pos` in att.inguistic, but maybe in the chapter or both).

sydb (approved):
[Note: I have not looked at the classes, yet; have to go do other things, but did not want to delay saying anything until I could say everything.]

I am not really competent to say whether or not TEI should support dependency syntax description (my instinct is “yes”) nor whether the proposed mechanism is the _right_ way to do this or not (no idea — I can say it seems sufficient to the task, but have no idea whether there might be other, better, methods). That said, I can make, and have made, several suggestions on wording, etc.

I think this could be merged either before or after consideration of most of the wordsmithing I suggest in the individual comments. The only thing I think actually _needs_ to be thought through before merging is the datatype and use of `@pos`.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Please kindly check the result when convenient -- we are in the process of addressing all your remarks, @sydb , and this one just happens to the be the first in the queue :-)

P5/Source/Guidelines/en/Images/dependency4.png
bansp: In our experience, the way it is done now (centred under the relevant syntactic node) is the most commonly found rendition in linguistic writing. However, we suspect that what led you to make the remark, Syd, is the "tails" of those columns that group feature value assignments -- and these, indeed, can be improved, by aligning the 'equals' signs -- a new suggestion will be coming up tomorrow.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: In the computational approaches that we know about, the relationship always holds between two items. So we'd rather keep the "between" -- but thanks for the double-check!

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: We now put "these Guidelines" in this spot, and we have also removed "in TEI XML" from two places, because it was redundant (and I was sure I saw a remark from you addressing that phrase, but maybe I only imagined it ;-)).

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: We're taking the first suggested option -- thanks!

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Accepted, thanks!

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Done, thanks.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Taking the latter, thanks!

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Accepted, thanks.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Some such!  :-) Taken with a minor modification, thanks!

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: We are very grateful for this -- we have speculated on that point as well.
We have modified the att.linguistic spec by adding one more exemplum, and modified slightly the wording in the passage at hand. (Tried to be minimalistic about that -- please see if that's OK.)

bansp:
Ooops, I intended to say that I added the `@ident` attribute (to silence Schematron), and my fingers produced something different -- sorry about that.

bansp:
I have aligned the display of example sentences with the way it is done in [att.linguistic](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.linguistic.html) .
The main issue here is the sentence-final fullstop, which is also analysed (and given a POS such as 'PUNCT'). By using `<q>`, we keep that example-final dot nicely separated from the sentence-final one.

Examples from att.linguistic:

> The English sentence ‘We're going to Brazil.’ tagged with (...)

Current example from [att.linguistic.dependency](https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/ref-att.linguistic.dependency.html) (while Mr. Jenkins is still digesting the new commit):

> Universal Dependency (UD) annotation of the English sentence _She buys books._. 

(Two dots, the first italicised together with its example sentence.)

In the [proposed addition to ch. 18](https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/AI.html#AILADEP), we see:

> In this directed graph, every node is labelled with one token of the sentence _She buys books._ The nodes (...)

-- which is a little typo, but I itch to correct it by also switching from `<mentioned>` to `<q>` -- and let me do that, simply for us all to see how it gets displayed, even if the final decision is different.

Of course, it can never be easy, because Ch. 18 (and likely the rest of the Guidelines) contain some other sentence-sized elements inside `<mentioned>`, and I think I'm even responsible for the second one:

1.
```xml
 consider the following segmentation of the English
S-unit <mentioned>I didn't do it</mentioned>.
```


2.
```xml
<p>The following example shows a German sentence <mentioned>Wir fahren in den
Urlaub.</mentioned> (<q>We're going on vacation.</q>) annotated with all the attributes discussed
above.
```

Example (1) above arguably contains an error -- the analysed sentence contains a `<pc>.</pc>`, which is not mentioned there.

In example (2), I was looking for a way to distinguish between the example sentence and its translation.

Now, still for the sake of discussion, let me intervene above (see [the Guidelines, ch. 18](https://tei-c.org/release/doc/tei-p5-doc/en/html/AI.html#AILALW) for comparison). I'm going to use `<q>` in example (1) and do the following in example (2):

```xml
<p>The following example shows a German sentence <q>Wir fahren in den
Urlaub.</q> ("We're going on vacation.") annotated (...)
```

-- because I have found literal `"` double quotes used in this chapter (but not for a translation -- search for "Gottfried").

Commit follows.

bansp:
I _am_ going to try to stop myself from further interventions!

But, while I was admiring how nicely Andreas has modified our most complex graph diagram, it occurred to me that, to many eyes, what we are presenting here is heavily pen-and-paper theoretical stuff.  (In my eyes, it is an utterly practical, pragmatic approach, with quite a few compromises along the way, made for the sake of it being close to universally applicable. Some theorists would squint at it -- but it _works_.)

So, I thought, let us add a footnote with a URL to an [online tool](https://www.let.rug.nl/kleiweg/conllu/) that has been there for years, where each of our listings can be pasted, and which produces an interactive graph representation. Let me therefore try to see where I can place that footnote -- perhaps early is better.

bansp:
Let me explain my last bunch of probably weird-looking commits: I was experimenting with the Jenkins settings, and after one try, everything crumbled, sending me into panic mode and into trying various ways to recover. Only belatedly did I realise that the breakage came from the dependency on Stylesheets, which happened to get broken exactly when I thought I had "15 minutes to experiment" (sigh, Murphy's Laws). 

Now the Stylesheets are back, but what I want to keep after the flurry of panic commits is a followup of [Syd's remark elsewhere](https://github.com/TEIC/TEI/discussions/2628) that made me realise that our PI at the top of the added file (`att.linguistic.dependency.xml‎`) as well as the modified chapter (`AI-AnalyticMechanisms.xml`) unnecessarily references an artefact derived by the `dist` target of the Makefile -- I can simplify that and resign from `dist` entirely, which should remove one obstacle to our work on this PR, namely the time needed to wait until our modifications get rendered (going from 30+ minutes to 3+ minutes seems worth it).

When I thought I inadvertently deleted all Jenkins builds, I came up with the following URL for the validation PI:
`https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/ws/P5/p5.nvdl`

-- it doesn't reference any build but rather the work space directly. There is still a bit of magic in all this and I feel more like a witch doctor than a physician, but it works and our additions validate again, so I'd like to keep that, even though the artefacts are back there now.

bansp:
I would also be grateful for a hint if I should keep this branch continually updated against TEI-C:dev. It is becoming a bit tiring, without knowing if it's useful to the reviewers and when it may be critical, and when it's completely spurious. 

I am now merging two outstanding commits because I can see that the TEI-dev line in the Jenkins was successful after they'd got merged. But then I'll try to stop thinking about this until I get a hint that I should act. Cheers.

martindholmes:
@bansp Generally it's a good idea to merge dev into your branch regularly, so you'll find out quickly if a commit to dev has broken your build.

bansp:
Thanks, @martindholmes -- I try :-)

I realise that this is just one of many issues and I realise that the Council is sometimes (if not most often) overwhelmed with the amount of extra work on your backs. 

It would be great -- if possible -- to get at least a ballpark estimate of when Andreas and I should potentially reserve time for the next step, in case new issues crop up. We've addressed the issues Syd mentioned, and a bit more on top of that, and it would be great to know if there is a chance for a next round before the end of this month, or the next month, etc.

I'm not asking for this to be prioritised, just for the information on how far down the queue this issue is. If it's not deep, I'd try to monitor the TEI repo for changes more often. Otherwise, I'd set a weekly reminder in my calendar. Any chance you could give me an estimation, based on what you know?

martindholmes:
@bansp We're in refrigeration for the next release right now, and the release should happen on Tuesday. I would suggest that immediately following the release, you merge dev again, and then assign the PR to someone (not just reviewers, an assignee). That can be me if you like, and then it'll be my job to get reviews done and see about merging.

bansp:
Thanks, Martin! Noted, will do.

bansp:
Thanks for your kind offer, @martindholmes. Assigned you, having merged the dev branch into the PR.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1146,8 +1151,8 @@ have <att>join</att> with the value <val>right</val> on the tokens
 <mentioned>victim</mentioned> and <mentioned>surfaced</mentioned>, or whether information
 on a single direction of adjacency is enough. Strategies vary, and it is important to
 document them in the TEI header.</p>
-<p>The following example shows a German sentence <mentioned>Wir fahren in den
-Urlaub.</mentioned> (<q>We're going on vacation.</q>) annotated with all the attributes discussed
+<p>The following example shows a German sentence <q>Wir fahren in den
+Urlaub.</q> ("We're going on vacation.") annotated with all the attributes discussed
```
martindholmes, line 1156: @bansp Why was the `<q>` element replaced with literal quotes here?

bansp: As in https://github.com/TEIC/TEI/pull/2888#issuecomment-4920759476 -- I was looking for a way to handle this, and double quotes are used elsewhere, so I thought I'd demonstrate what it would look like and then we'd see.

A fairly widespread convention in linguistic publications is to use italics for the in-line examples from the/an object language, followed by a translation in double quotes in the language of the publication. Of course, we want this fragment and the whole of the chapter to be aligned with the rest of the Guidelines, but I wasn't sure what that alignment should look like.

P5/Source/Specs/att.linguistic.dependency.xml
```diff
@@ -0,0 +1,187 @@
...
+  <desc versionDate="2025-12-02" xml:lang="en">provides attributes for
+    annotating dependency relations between word-level elements.</desc>
+  <classes/>
+  <attList>
+    <attDef ident="depN" usage="opt">
+      <gloss versionDate="2025-12-02" xml:lang="en">token index</gloss>
+      <desc versionDate="2025-12-02" xml:lang="en">encodes the token index or index
+        range of a word-level element in a syntactic dependency relation.</desc>
+      <datatype><dataRef key="teidata.word"/></datatype>
+      <valDesc versionDate="2025-12-02" xml:lang="en">is usually an
```
martindholmes, line 18: On this `<valDesc>` and the following ones, see:

https://github.com/TEIC/TEI/issues/2938

bansp: Hmm, it would be a pity to take them out -- they do have explanatory value, don't they. But especially if that should become a debated blocker, we'd rather get rid of the element.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
```diff
@@ -1187,6 +1192,176 @@ sake of readability).
     <w lemma="forth" pos="av" orig="sorth">forth</w>
   </egXML></p>
 </div>
+
+<div type="div3" xml:id="AILADEP">
+<head>Syntactic dependency relations between word-level elements</head>
```
martindholmes: Guidelines headings should be Title Case.

P5/Source/Specs/att.linguistic.dependency.xml
```diff
@@ -0,0 +1,187 @@
...
+        relation.</desc>
+      <datatype><dataRef key="teidata.text"/></datatype>
+      <valDesc versionDate="2025-12-02" xml:lang="en">should use a colon to
+        separate the head’s token index and <!-- the label for --> the type of
+        the dependency relation, and a pipe symbol as a list-item separator.</valDesc>
+    </attDef>
+  </attList>
+  <exemplum xml:lang="en">
+    <p>Universal Dependency (UD) annotation of the English sentence <q>She buys
+    books.</q>.</p>
```
martindholmes: @bansp There are two periods here.

P5/Source/Specs/att.linguistic.dependency.xml
```diff
@@ -0,0 +1,187 @@
...
+        <w depN="3" depR="4:cc" depE="4:cc" pos="CCONJ CC" lemma="and">and</w>
+        <w depN="4" depR="2:conj" depE="0:root|2:conj" pos="VERB VBZ" msd="Mood=Ind|Number=Sing|Person=3|Tense=Pres|VerbForm=Fin" lemma="sell">sells</w>
+        <w depN="5" depR="2:obj" depE="2:obj|4:obj" pos="NOUN NNS" msd="Number=Plur" lemma="book" join="right">books</w>
+        <pc depN="6" depR="2:punct" depE="2:punct" pos="PUNCT ." lemma=".">.</pc>
+      </s>
+    </egXML>
+  </exemplum>
+  <exemplum xml:lang="en">
+    <p>Universal Dependency (UD) annotation of the English sentence
+      <q>She likes coffee and he tea.</q>.</p>
```
martindholmes: @bansp Two periods here.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Oh that's truly ugly. But we'll conform, of course.

P5/Source/Specs/att.linguistic.dependency.xml
bansp: Nope. It's my comment https://github.com/TEIC/TEI/pull/2888#issuecomment-4920759476 again. There's a single period, at the end of the sentence. The other (linearly first)  '.' character plays no role in this very sentence -- or it plays the same role as any 's', '.', ':' would, inside the example sentence: it is part of the analysed string. Let me stress: a _string_ or a _character sequence_. It's a pure coincidence that the analysed string ends in a period.

We could surely reword the wrapper sentence -- but that's pure cosmetics. If you think it's better to put some word fluff after the example and before the period, I will of course suggest something.

P5/Source/Specs/att.linguistic.dependency.xml
bansp: As above. Word fluff surely possible, if you think that's better.

P5/Source/Specs/att.linguistic.dependency.xml
bansp: I failed to come up with a non-nauseating suggestion for this, so I changed the last period to a colon. Please kindly see if that's digestible. Same for the other 'double' period.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
bansp: Changed, thanks.

P5/Source/Guidelines/en/AI-AnalyticMechanisms.xml
martindholmes: We should take a look for similar cases and follow the model there. My instinct would be to use `<q>`, but I could be wrong.

P5/Source/Specs/att.linguistic.dependency.xml
martindholmes: I need to find out why Council decided to avoid `<valDesc>`, and what was supposed to be used instead. Possibly just `<remarks>`, because that's regularly translated into the other languages, whereas `<valDesc>` probably isn't.

P5/Source/Specs/att.linguistic.dependency.xml
martindholmes: I think this is purely a question of style. For me, too, it's logical to end the example sentence with a period, and then end its containing sentence with its own period, but in most English-language style guides, this would be deemed wrong; the final period would be omitted, so the one within the quotation would serve dual function. Again, we should be able to find other examples elsewhere in the Guidelines.

bansp: It's just occurred to me that, after the example character sequence, we could add a fragment like "(analysed together with the final period)". It sounds a bit cheesy perhaps, but it avoids the double-dot clumsiness, and points out a detail perhaps not immediately obvious to the general reader.

P5/Source/Specs/att.linguistic.dependency.xml
bansp: To potentially speed things up, I could redo them as `<remarks>` and lift the potential time pressure. But I guess it makes sense to keep it as-is for now, to demonstrate to the Council that the element isn't useless. So let me wait for a signal with this one.

bansp:
I have pushed two cosmetic fix commits and I am a bit afraid that I may need to postpone until late August other pieces of ch. 18 cosmetics, e.g. adding a dot to the running 'Kruger' example where appropriate -- part of the issue is that the phrasal annotation with CLAWS is probably not easily reproducible nowadays, and quite possibly some other parser should be used there -- something that would not pose a challenge to the reader the way CLAWS does (PennBank would be better and clearer, I think, in terms of labelled bracketing).

1. I think that it might make a lot of sense to adopt `<mentioned>` as the wrapper for pieces of object language of whatever size (and again, a minor issue concerns what to do with `<q rend="display">` in [18.4.1](https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/AI.html#AILAGD) that should probably be `<egXML>`). 
2. And then use `<q>` for translations into the language of the narration,
3. The "double dot" problem is not a problem of the Guidelines, so the way to handle it should be editorial -- cook something sensible up to disrupt the sequence of double dots, but also `!</mentioned>,` and other clusters where the first punctuation character belongs to an example, and the other belongs to the narrative stream.

The above means re-examining the two spec files and section 18.4 (this PR), and then the rest of the Guidelines, notably the entire ch. 18, the Dictionary chapter, the Speech chapter (probably no issues there, I've skimmed it looking for trouble and finding none), the CMC chapter, and probably the graph and FR chapters.

I have no problem doing that with the Council's blessing -- but maybe in a separate pull request, so that this one is not spread thin and dragged out?

martindholmes:
@bansp I think I would now recommend that you redo the valDescs as remarks, since the Council is not likely to investigate the valDesc issue soon.

bansp:
@martindholmes Thanks. Pushed one change to depN. If that one's OK, I'll handle the rest analogously.

Edit: did that for all three. The result is at https://jenkins-paderborn.tei-c.org/view/LingSIG/job/TEIP5-LingSIG-att.linguistic.dependency/lastSuccessfulBuild/artifact/P5/Guidelines-web/en/html/ref-att.linguistic.dependency.html

martindholmes:
Thanks @bansp! We should be able to move forward with this now.

Issue #2892, opened 2026-05-05
Add <gloss> to att.placement


ebeshero:
Per this [discussion on the TEI-L list](https://lists.psu.edu/cgi-bin/wa?A2=TEI-L;8f1da487.2605&S=) let's add `<gloss>` to att.placement so that people don't have to create workarounds for it like `<note type="gloss" place="margin">...</note>`.

trishaoconnor:
See PR #2899 for adding `<gloss>` to `att.placement`.
See PR #2902 for updates to the prose in [Terms and Glosses](https://tei-c.org/release/doc/tei-p5-doc/en/html/CO.html#COHTG) to mention that `<gloss>` is a member of `att.placement` and to highlight the suggested values that might be most relevant for encoding glosses.

PR #2899, opened 2026-05-09
add `<gloss>` to `att.placment`

trishaoconnor:
This PR addresses #2892  and adds `<gloss>` to `att.placement`.

### DIFF
--- a/P5/Source/Specs/gloss.xml
+++ b/P5/Source/Specs/gloss.xml
@@ -20,6 +20,7 @@
     <memberOf key="att.cReferencing"/>
     <memberOf key="att.cmc"/>
     <memberOf key="att.declaring"/>
+    <memberOf key="att.placement"/>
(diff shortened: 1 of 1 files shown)


sydb (approved):
Seems perfectly OK to me, and passes tests.

No example is added, but it is already the case that most of the members of att.placement don’t have an example of its use. (And how many times does the same thing need to be exemplified, anyway?)

Members of att.placement (in alphabetical order) preceded by number of examples with the `@place` attribute:
~~~
18 add
   addSpan
   del
   delSpan
   div
   figure
 7 fw
   gloss      <!-- new -->
 1 head
 3 label
 6 metamark
   mod
   notatedMusic
28 note
   noteGrp
   postscript
   redo
   restore
   retrace
24 rt
   sp
   speaker
 1 stage
   subst
   substJoin
   trailer
   undo
   witDetail
~~~

Issue #2898, opened 2026-05-06
overly complex XPath logic


sydb:
The "spec-in-module" constraint in att.identified currently looks like this (line numbers added):
~~~xml
[01]  <constraint>
[02]    <sch:rule context="tei:elementSpec[ @module ]
[03]                      |tei:classSpec[ @module ]
[04]                      |tei:macroSpec[ @module ] ">
[05]      <sch:assert test="( not( ancestor::tei:schemaSpec | ancestor::tei:TEI | ancestor::tei:teiCorpus ) )
[06]                        or
[07]                        (
[08]                          not( @module )
[09]                          or
[10]                          (
[11]                            not( //tei:moduleSpec )
[12]                            and
[13]                            not( //tei:moduleRef )
[14]                          )
[15]                          or
[16]                          ( //tei:moduleSpec[ @ident = current()/@module ] )
[17]                          or
[18]                          ( //tei:moduleRef[ @key = current()/@module ] )
[19]                        )">
[20]      Specification <sch:value-of select="@ident"/>: the value of
[21]      the module attribute ("<sch:value-of select="@module"/>")
[22]      should correspond to an existing module, via a moduleSpec or
[23]      moduleRef</sch:assert>
[24]    </sch:rule>
[25]  </constraint>
~~~
This test, which was last worked on > 13 years ago by SPQR, strikes me as overly complex.

First, the overall structure is needlessly complicated with at least one extra level of parens. If we replace each actual XPath snippet with the line number it is on, the structure boils down to:
~~~
[05]
or
(
  [08]
  or
  (
    [11]
    and
    [13]
  )
  or
  [16]
  or
  [18]
)
~~~
Which might be easier to grok with less whitespace:
~~~
[05] or ( [08] or ( [11] and [13] ) or [16] or [18] )
~~~
If we replace the compound `( [11] and [13] )` with `[12]`, we get
~~~
[05] or ( [08] or [12] or [16] or [18] )
~~~
I submit that the parens add nothing (but confusion) to the logic of the test.

If we examine the structure of [12] more closely, it is `( not(A) and not(B) )`. Logically that is precisely euqivalent to `not( A or B )`.

Furthermore, line 08 (and thus 09) is (are) completely reduntant: the rule is only being fired when a `@module` attribute is present.

And lastly, rather than test the values of `@ident` of `<moduleSpec>` separately form the values of `@key` of `<moduleRef>`, why not test them at the same time?

Thus I think the entire thing should be reduced to:
[...]

PR #2900, opened 2026-05-09
make "spec-in-module" test less confusing

sydb:
Replace the `<constraint>` "spec-in-module" with that which was recommended on issue #2898.

### DIFF
--- a/P5/Source/Specs/att.identified.xml
+++ b/P5/Source/Specs/att.identified.xml
@@ -13,25 +13,27 @@
     <memberOf key="att.combinable"/>
   </classes>
   <constraintSpec ident="spec-in-module" scheme="schematron" xml:lang="en">
+    <!--
+      The @test below comprises three expressions in a disjunction.
+      Only if all three are false() is the error message generated;
+      i.e., if any one of the three is true() no message is
+      generated. The first two test for whether or not we are in a
+      circumstance in which the basic question, “does the module
+      @module exist?” can be meaningfully asked. The sub-tests are:
+      1) This specification element (which has a @module attr) is free-standing; i.e. it is NOT a descendant of <schemaSpec>, <TEI>, nor of <teiCorpus> — this is TRUE for the individual files in which we store the TEI P5 specifications, but false for p5.xml (or p5subset.xml) itself, and false for any proper customization file.
+      2) There are no <moduleSpec> nor <moduleRef> elements in this file.
+      3) In this file there is either a <moduleSpec>’s @ident or a <moduleRef>’s @key which matches this specification’s @module.
+    -->
     <constraint>
       <sch:rule context="tei:elementSpec[ @module ]
                         |tei:classSpec[ @module ]
                         |tei:macroSpec[ @module ] ">
         <sch:assert test="( not( ancestor::tei:schemaSpec | ancestor::tei:TEI | ancestor::tei:teiCorpus ) )
                           or
-                          (
-                            not( @module )
-                            or
-                            (
-                              not( //tei:moduleSpec )
-                              and
-                              not( //tei:moduleRef )
-                            )
-                            or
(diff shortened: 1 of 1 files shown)


jamescummings (approved):
I have not tested this rule, but the XPath looks good to me. (And the logic behind it seems to make sense, though I did have to read the very helpful note three times.)

martindholmes:
@sydb If I'm understanding correctly, the last test is the one most likely to fire for most well-formed customizations:

`( //tei:moduleSpec/@ident | //tei:moduleRef/@key ) = current()/@module`

so wouldn't it make sense to put that one first, to avoid having to evaluate the others? (I'm assuming there's no intervening optimization taking place somewhere.)

sydb:
Short answer, @martindholmes, is I have no idea. That is, no idea which way might be more efficient. But in my brain it is easier to figure it out if the steps are “First, see if it even makes sense to test; then, if so, do this test”. In truth, I think of `rule/@context` as setting up whether the test should be fired or not, so to me it probably should be
~~~xml
<sch:rule context="( tei:elementSpec[@module] | tei:classSpec[@module] | tei:macroSpec[@module] )[
                     ( ancestor::tei:schemaSpec | ancestor::tei:TEI | ancestor::tei:teiCorpus )
                     and
                     ( //tei:moduleSpec | //tei:moduleRef ) ]">
  <sch:assert test="( //tei:moduleSpec/@ident | //tei:moduleRef/@key ) = current()/@module">
    Specification <sch:value-of select="@ident"/>: the value of the module attribute
    ("<sch:value-of select="@module"/>") should correspond to an existing module, via a
    moduleSpec or moduleRef</sch:assert>
</sch:rule>
~~~
That is, put _all_ the conditions in `rule/@context` and only the test in `assert/@test` (or `report/@test`). But I thought other folks might balk at that complexity being in `@context`, especially since we have had all the complexity in `@test` for well over a decade, now.
But if you think that is as easy or easier for a human to follow, I would be happy to change.

P.S. I think it even makes the explanatory comment simpler, but not sure, I have not written it yet. :smile:

sydb:
I have now re-written to the new “all the logic about whether we should be looking to see if the module is defined or not in the `@context`” method. I like this better in large part because the comment is easier process. Or at least I think so. @jamescummings?

martindholmes (commented):
Just a quibble, but could I ask for a period at the end of the sch:assert? We haven't been consistent in this in the spec files, but we should be. If it's a full sentence, it should end with a period. I should do a trawl through the rest of our Schematron and standardize this.

sydb:
> … could I ask for a period at the end of the sch:assert? …

Done.
See also #2905.

martindholmes (approved):
Looks good!

ebeshero:
@sydb I think by now we have most of us agreeing with the logic and syntax, but I am wondering if we actually tested it on an actual ODD in the wild? I think I could do so this afternoon and if it works, just merge this...

sydb:
Good question, @ebeshero. I do not know if I have ever seen this error either before or after my change. Hang on, I will try testing it (both in dev branch and this branch) against all the ODDs in the ATOP collection …

sydb:
(You wanna a build a test ODD file that _should_ fail this test?)

ebeshero:
Yes--that's what I was thinking... Also, I've just been testing the branch and so far all's well there.

sydb:
Ha!
There are ~61 ODD files in the ATOP hierarchy … I tested all of them. In those 61 files there are 14 errors flagged by this constraint (the "spec-in-module" constraint) in 5 files. _All_ 28 messages are different … after this change, they end in a period! :smile: 
So it looks to me like it works exactly as expected. (In all 14 cases the XPath flagged as an error is the same before & after this change, the only difference is the ultimate period.)

ebeshero:
Well! On the strength of that beautiful testing, I'm just going ahead and merge this branch! :-D

Issue #1516, opened 2016-10-10
appInfo spec is somewhat unintuitive;   interpretation of pointing elements and of the lack thereof


bansp:
I was about to mention the appInfo element to a fellow Freedict developer, in the hope that he would be able to look at the [spec](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-appInfo.html) and move from there, when the cute fragment `<ptr target="#P1"/>` caught my eye.

OK, this is explained in the [text of the section](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/HD.html#HDAPP), but in the brief spec description the pointer is completely unintuitive. I thought at first that it targeted a description of the tool elsewhere in the header.

I would like to suggest a more thorough description in the spec, and perhaps also including `<ref>` as an alternative to `<ptr>`, especially if more than one part is affected and they may be meaningfully labelled.

Another thing I wondered about is the interpretation of `<application>` in the absence of any pointing element inside. I understand that in such cases, it applies to the entire document (pre- or post-inclusion?), but it may be worth mentioning in the description as well.

Definitely a low-priority issue but worth raising, I thought. HTH

martindholmes:
I absolutely agree. This particular thing was my first contribution to the TEI and I didn't really do a great job of it. My initial feeling was that as other applications began to make use of these elements, more features might be added and the functionality fleshed out a bit, but it turned out that they were not widely used. In fact, other than the Image Markup Tool which first gave rise to the feature, I don't know of any application using these elements. Perhaps their time has come, or perhaps the opposite: they may have proved to be nothing but clutter and might be deprecated.

laurentromary:
This is a perfect timing to revisit this element since we need it for describing th tools involved in standOff annotation. One of the difficulties I have encountered is also its position in the header which prevent it being used in alternation with a case where we deal with human annotator. Maybe appInfo should be made a member of model.respLike and/or better articulated with respStmt.

bansp:
So your suggestion. Laurent, is to either make it a member of both model.encodingDescPart and model.respLike, or to use respStmt where and how, please? (I can't see the other option clearly)

martindholmes:
Or why not just point at it with a respStmt/resp?

laurentromary:
Just thinking aloud:
`<respStmt>
 <resp>...</resp>
 <appInfo>...</appInfo>
</respStmt>`
:-}
Pointing is an overhead, really.

martindholmes:
[...]

PR #2901, opened 2026-05-09
tweaks re: `<application>`

sydb:
Minor improvements to the wording, encoding, and punctuation of the discussion of the `<appInfo>` and `<application>` elements. I have made these changes and this PR as part of #1516, but it does not even come _close_ to closing that issue. These are just helpful tweaks.

In both the tagdoc for `<application>` and section 2.3.11 (#HDAPP) we find (emphasis added by me):

> … has an interest in two parts of **a** document which was last saved …

I am wondering if that indefinite article should be the definite article “the”, or even something like “the TEI document in which this `<appInfo>` appears”?


martindholmes (commented):
These changes look good to me.

PR #2902, opened 2026-05-10
Update prose on `<gloss>` to mention `att.placement` membership

trishaoconnor:
This PR addresses issue #2892 by updating the [Terms and Glosses](https://tei-c.org/release/doc/tei-p5-doc/en/html/CO.html#COHTG) section of the Guidelines to state that the `<gloss>` element claims membership of `att.placement`. I also highlighted the suggested values which might be most relevant for encoding glosses that occur in the margins of a text as well as above or below a line of text.

### DIFF
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -1053,7 +1053,7 @@ overshadows the historical content</quote>.</egXML>
       <att>place</att> attribute for indicating where the
       gloss is with respect to the text.
       <specList>
-        <specDesc key="att.placement" atts="place"/>
+        <specDesc key="att.placement"/>
       </specList>
       The most relevant suggested values of <att>place</att> for
       glosses that occur in the margin of the text are <val>top</val>, <val>bottom</val>,
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -1048,8 +1048,16 @@ overshadows the historical content</quote>.</egXML>
         <specDesc key="term"/>
         <specDesc key="gloss"/>
       </specList>
-      These elements are also members of
-      the class <ident type="class">model.emphLike</ident>.
+      These elements are members of
+      the class <ident type="class">model.emphLike</ident>. As a member of the class <ident type="class">att.placement</ident>, the <gi>gloss</gi> element also carries the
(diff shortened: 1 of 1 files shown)


P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1048,8 +1048,16 @@ overshadows the historical content</quote>.</egXML>
...
         <specDesc key="gloss"/>
       </specList>
-      These elements are also members of
-      the class <ident type="class">model.emphLike</ident>.
+      These elements are members of
+      the class <ident type="class">model.emphLike</ident>. As a member of the class <ident type="class">att.placement</ident>, the <gi>gloss</gi> element also carries the
+      <att>place</att> attribute for indicating where the
+      gloss is with respect to the text.
+      <specList>
+        <specDesc key="att.placement" atts="place"/>
```
sydb: I am hesitant about the addition of `@atts` in this line. While I am not really _against_ it, seems to me the only thing it achieves is that if we ever add another attribute to att.placement (e.g. `@subplace` :​-​) it would not show up here. That might be a good thing, but I think it more likely to be a bad thing.

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -1218,7 +1226,7 @@ overshadows the historical content</quote>.</egXML>
             <specDesc key="rb"/>
             <specDesc key="rt"/>
           </specList>
-          The <gi>rt</gi> element is a member of <ident type="class">att.placement</ident>, and thus the
+C          The <gi>rt</gi> element is a member of <ident type="class">att.placement</ident>, and thus the
```
sydb: Looks like just a boo-boo. The leading ‘C’ should be deleted.

sydb (changes requested):
Only major thing that needs to be fixed before merge is the typo on line 1229.

P5/Source/Guidelines/en/CO-CoreElements.xml
trishaoconnor: Thank you @sydb  for catching this embarrassing typo.

P5/Source/Guidelines/en/CO-CoreElements.xml
trishaoconnor: Thank you for bringing this to my attention @sydb, I have removed `@atts` from line 1056 now.

PR #2903, opened 2026-05-12
Update oversight in 2 ⇒ 2 or 3 values for a point

sydb:
When data.point was changed from 2 values to 2 or 3 values the `<remarks>` were not changed to match.


martindholmes (approved):
Looks good.

PR #2909, opened 2026-06-16
limit how long test for website waits

### LINKED ISSUE #2908: P5 tests should correctly fail when the TEI website is down
The TEI Test suite depends on the TEI website being available. The make target "testmeta2010" tests this by trying to `curl` a file from the TEIC website, but if the site is not available, this may hang for many minutes. @sydb and I believe that the curl command should be modified from:

`curl -s --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml`

to

`curl --silent --max-time 15 --show-error --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml` 

which means the the command will timeout after 15 seconds. At the moment Jenkins will fail the tests on the Error 28 which would be thrown at the command line by this timeout, but we could make the build ignore this by modifying the tei-log-parse-rules in the TEIC/Jenkins repo.

There are other parts of the tests (detest.odd) which depend on the TEI site being up, and those will require additional work.

We should consider whether any tests depending on the TEI website being up should be able to fail gracefully, so that other tests can continue to be run even when the TEI site is down.

sydb:
In order to fix #2908, with @martindholmes, added a `--max-time` switch to the `curl` command that protects the "testmeta2010" test from running when the website is down.

### DIFF
--- a/P5/Test/Makefile
+++ b/P5/Test/Makefile
@@ -81,7 +81,7 @@ testmeta2010:
 	@echo " "
 	@echo "--------- P5/Test/Makefile work on target $@ ---------"
 	@echo check that TEI web site is up and responding ...
-	-TEISITE=`curl -s --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml`
+	-TEISITE=`curl --silent --max-time 15 --show-error --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml`
 ifdef TEISITE
 	@echo ... it is, so run tests
 	$(ANT) -f antruntest.xml -DoddFile=testmeta2010.odd -Dtestfile=testmeta2010.xml validateodd


P5/Test/Makefile
```diff
@@ -81,7 +81,7 @@ testmeta2010:
 	@echo " "
 	@echo "--------- P5/Test/Makefile work on target $@ ---------"
 	@echo check that TEI web site is up and responding ...
-	-TEISITE=`curl -s --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml`
+	-TEISITE=`curl --silent --max-time 15 --show-error --head http://www.tei-c.org/Vault/P5/1.5.0/xml/tei/odd/p5subset.xml`
```
ebeshero, line 84: 15 seconds, yes?

ebeshero (approved):

raffazizzi:
This seems to have created dozens of undefined prefix "xlink" errors...

PR #2910, opened 2026-06-22
regularize Schematron msgs that do not end in dot

sydb:
Mostly put dots at end of Schematron msgs that are full sentences; but also put element names into pointy brackets, put at-sign in front of attr names, etc.

Note1: This PR replaced #2905 because I changed the branch name.

Note2: Regularizing the use of characters to indicate element names, attr names, and attr values for Schematron msgs that already end in dot is handled in branch "sydb_regularize_undotted_Schematron_msgs" by PR #2911.


hennyu:
This looks good to me.

hennyu (approved):
This looks good to me.

PR #2911, opened 2026-06-22
Regularize Schematron msgs that already end in dot 

sydb:
Regularize those Schematron msgs that end in dot so that they use standard characters to indicate element names, attribute names, and attribute values. Those that did not end in dot are addressed in #2910.


torstenroeder (approved):
all positive

torstenroeder (commented):
all tests passed. is it ok that a path from your local machine appears in some @documents attributes?

sydb:
Good catch. Yes, it is OK insofar as it is not a problem for the TEI build process (which does not look at the `@documents` attr.)
It is not OK in that it is feels like a small invasion of privacy that should be avoided. But that horse left the barn over a decade ago, I’m afraid. (That is, a path on my local machine was probably first checked into a public repo over 10 years ago.)

Issue #2913, opened 2026-06-22
Constraints disallowing a single child of `<sequence>`, `<alternate>`, and `<interleave>` should be removed


martindholmes:
Following a discussion of a presentation (https://github.com/TEIC/atop/blob/dev/Documentation/presentations/Council_repeatability_after_auto-deletion.odp) by @sydb at today's Council meeting, Council arrived at a consensus to the effect that the problem that arises when deletions result in a single wrapper element ( `<sequence>`, `<alternate>`, or `<interleave>`) containing only one child (e.g. `<elementRef>`) can be circumvented if this scenario is allowed and considered valid. Therefore the `<constraintSpec>`s "alternatechilden", "interleavechilden", and "sequencechildren" should be removed.

The idea would then be that the `@minOccurs` and `@maxOccurs` on the wrapper elements would act in the normal way, combining with those on the single child, to give the actual range of instances allowed.

PR #2915, opened 2026-06-23
allow singleton child of ODD grouping element

sydb:
This PR addresses #2913 by removing from P5 the constraint that an `<alternate>`, `<interleave>`, or `<sequence>` must have two or more children, and adding that same constraint to p5odds and tei_customization.

It also fixes the bug discovered in #2734 wherein `<valList>` was allowed to occur as a child of `<alternate>` more than once. (Which should not be allowed, in large part because the Stylesheets drop the entire construct when this happens.)

### DIFF
--- a/P5/Source/Specs/alternate.xml
+++ b/P5/Source/Specs/alternate.xml
@@ -10,18 +10,24 @@
     <memberOf key="model.contentPart"/>
   </classes>
   <content>
-    <alternate minOccurs="1" maxOccurs="unbounded">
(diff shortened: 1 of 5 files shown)


martindholmes (commented):
First problem when building P5:
TEI/P5/p5.xml:112670:25: error: element "interleave" from namespace "http://www.tei-c.org/ns/Examples" not allowed in this context

We're working on that now.

martindholmes:
Tests are passing -- merging now.

Issue #2896, opened 2026-05-06
Add `@ident` to all `<remarks>` in the P5 source


martindholmes:
Council agrees that whatever the final process for issue #2826, all `<remarks>` in the P5 source should have `@ident`.

ebeshero:
Check if the needed constraint is added to be sure whether this is solved for the 4.12 release. @martindholmes

martindholmes:
I'm pretty sure this was solved by #2918, so I'm just going to close it.

PR #2917, opened 2026-06-29
add `@ident`s to `<remarks>`, INCLUDING the language

sydb:
WARNING: This PR and #2918 are mutually exclusive — they are two different ways to address issue #2896. We should merge one or the other, but **not** both.

This PR adds a unique `@ident` to every `<remarks>` element in the Guidelines. (It does NOT change the 9 cases of `<remarks>` within `<egXML>`, at least not yet.) The values are constructed from the ancestor `@ident` and `@xml:id` values and the `@xml:lang` is appended.  

Also manually deprecate optional `@ident` to required as of 2027-06-01 — output is suboptimal IMHO.


martindholmes:
I think this is NOT the right approach; I prefer #2918. Since the `@xml:lang` is already present on the element, it makes no sense to echo it on the `@ident`.

sydb:
Both @martindholmes and @martinascholger assert that we should use #2918 instead. Since I agree, closing this as we won’t be using it.

Issue #2826, opened 2025-10-08
There should only be one `<remarks>` element per language


martindholmes:
The constraint enforcing this already exists, in p5odds.odd. It makes good sense because `<remarks>` is not identifiable (no `@ident`) and therefore it is not possible to specify, other than by `@xml:lang`, which `<remarks>` element should be replaced in an ODD customization; and the element name itself, `<remarks>`, clearly indicates that it is designed to contain multiple remarks. We suggest that the constraint from p5odds.odd be ported over to the P5 schema and thus made universal.

sydb:
Personally I think this needs discussion only insofar as whether the same constraint (only one in any given language) should apply to `<desc>` and `<gloss>` (when used as a child of `*Spec`), too.

sydb:
Council (subset) discussion reveals that we really think this constraint should be applied to `<gloss>` and `<desc>` in P5 (not just p5odds), but that for `<remarks>` we think `@ident` should be required. Both these changes should be made via hand deprecation, of course.

jamescummings:
I note the remarks does have an `@ident` now, unlike what @martindholmes said when originally posting. I'm not sure if that solves the issue however. 

I'm assuming the limit of one per language would be within its containing parent element, rather than say for the elementSpec as a whole, since obviously one might have remarks about particular attributes or valItems? 

Seems reasonable to me otherwise.

martindholmes:
Council agrees that we can go ahead with the subissue, and then the ATOP team should then make a formal proposal on whether/how/where `@ident` should appear on all `<remarks>`.

martindholmes:
The subissue is done and closed, so what remains to be decided by Council is:

1. Should we add a constraint such that only one `<remarks>` element per language is allowed?
2. Should we apply that only in p5_odds, or make it general for all odds?
3. Should we extend it to `<gloss>` and `<desc>`?

Personally I think: yes to 1 and 3, but only in p5_odds.

raffazizzi:
> 1. Should we add a constraint such that only one <remarks> element per language is allowed?

Yes and I would do this generally, not just for p5_odds. The plural indicates that this is a container for multiple remarks. Splitting them into multiple `<remarks`> elements would perhaps make sense if it allowed `@type`, but it doesn't. So a stricter schema that matches out reference implementation seems saner.

> 3. I would add the constraint only in p5_odds; `<gloss>` and `<desc>` aren't plural and have `@type`. Perhaps a schematron warning in the general schema?

[...]

PR #2918, opened 2026-06-29
add `@ident`s to `<remarks>`, EXCLUDING the language

sydb:
WARNING: This PR and #2917 are mutually exclusive — they are two different ways to address issue #2896. We should merge one or the other, but **not** both.

This PR adds a NON-unique `@ident` to every `<remarks>` element in the Guidelines, but the `<remarks>` CAN be uniquely identified by combing the `@ident` value with the value of `@xml:lang`. (It does NOT change the 9 cases of `<remarks>` within `<egXML>`, at least not yet.) The values are constructed from the ancestor `@ident` and `@xml:id` values.  

Also manually deprecate optional `@ident` to required as of 2027-06-01 — output is suboptimal IMHO.


martindholmes:
IMHO this is the correct way to address this problem. The `@ident` identifies the points being made in the `<remarks>` element, not the element itself; it should be shared across `<remarks>` elements making the same point in different languages. This is somewhat analogous to the fact that `<attDef>s` which are different customizations of a single source `<attDef>` share the same `@ident`. Since we also believe (in #2826) that in P5, there should only be one `<remarks>` per language, this means that in P5 there should only be one `@ident` per context, and its form should be entirely predictable and therefore constrainable via p5odds.odd.

martinascholger:
I agree with @martindholmes - I also vote for excluding the language.

martindholmes (approved):
With @sydb, added a Schematron rule and updated test to account for it.

Issue #2919, opened 2026-06-30
Test/expected-results/detest_odd_schematron.svrl should not be in the repo


martindholmes:
@sydb and I observe: the file in the title is generated during the Test process, and its contents are then processed to create a text file containing only the salient parts of it. In 2024 it ended up being added to git, but it shouldn't be there.

sydb:
Noting that if `Test/expected-results/detest_odd_schematron.svrl` is _not_ in the repo (and it probably should not be), then `Test/expected-results/detest_odd_schematron.txt` needs to be in the repo instead. The .txt version is derived from the .svrl file and is what the `detest` target actually uses.

sydb:
Note also a comment in the Test/Makefile:

> Yes, it would be more efficient to store expected-results/detest_odd_schematron.txt in the repo rather than generating it every time it is to be tested, but it seems to me this generation process is wonky enough that it would be difficult to reproduce every time the Schematron is updated.

That was written in 2024-05, and given that the generation process it is talking about has not given anyone any trouble in > 2 years, seems like it is time to stop generating the expected results every time.

ebeshero:
If it speeds up the testing process to stop dynamic generation of the .svrl file, then we should remove it and add it to the .gitignore if it's still going to be generated in future.


What triggers the making of the .svrl file in Expected Results? That seems pretty weird.

sydb:
Oh, good point, probably should add it to .gitignore. And while it does speed up the testing process by a few tenths of a second, that is not why I think we should do this. We should do it because it makes the testing process easier to follow, understand, and update.

The .svrl file in expected_results/ is created when we test validate a file (in this case the detest.odd file) against Schematron.

martindholmes:
All working as expected, so merged and closing.

PR #2920, opened 2026-06-30
make Test/Makefile less confusing

sydb:
Address the (minor) problem raised in #2919 by a minor re-write of the detest: target of the Makefile in the Test/ directory. (And also, of course, remove the .svrl file from this repo and add the new .log file which is used instead of the previous .txt file.)


sydb:
Note that of the 4 files changed, there is pretty much no point in looking at the differences of the output files (`P5/Test/expected-results/detest_odd_schematron.log`, `P5/Test/expected-results/detest_odd_schematron.svrl`, and `P5/Test/expected-results/detest_odd_schematron.txt`); the only thing important to read is the `P5/Test/Makefile` itself.

The main changes in this PR are:

- use the `Utilities/svrl_text_only.xslt` program to generate the Schematron messages-only file instead of a  `xmllint | perl | egrep` pipeline¹
- use the name detest_odd_schematron.**log** instead of detest_odd_schematron.**txt** for that output to match the others in the same target
- keep the expected_results/ version of that .log file in the repo, so we do not have to re-generate it every time²

**Notes**
¹ In truth, the pipeline is not particularly complex, and _much_ faster than running Saxon.³ But the difference in runtime is almost certainly negligible compared to the whole test process, and the XSLT both does a better job (because it reports the XPath to the problem location, which can occasionally be useful) and, I am betting, is easier for most TEIers to understand.
² Although, as mentioned above, the time cost is minimal.³
³ Running the pipeline takes ~0.004 s on my system; running that XSLT in Saxon takes ~0.30 s. But keep in mind, the XSLT is doing more, so that is not a real apples-to-apples comparison, but rather a reflection of how much extra time you will now have to wait — ~0.296 s, at least on my desktop machine. :smile:

martindholmes:
Tests pass locally for me.

joeytakeda (approved):
Tests pass for me too (using Docker)

sydb:
Sorry, @martindholmes — fixed now!
(And thanks for the catch.)

Issue #2724, opened 2025-06-22
Provide @perf on actor


lb42:
The attribute `@perf` (performance) is available on elements `<move>` and `<tech>`; in the former case it "identifies the performance or performances in which this movement occurred as specified by pointing to one or more performance elements";  in the latter   it "points to one or more performance elements documenting the performance or performances to which this technical direction applies". As discussed recently on TEI-L, if this were  also available on the `<actor>` element, it would be possible to associate an actor's name with a specific performance. This would greatly simplify the encoding of `<castList>`a presenting details of multiple performances.   Such an extension clearly retains the semantics of the existing attributer and meets a clearly defined need. 

I leave it to Council to decide whether this should be met by defining the attribute afresh on the `<actor>` element, or by defining a new attribute class att,performed _vel sim_, with members actor, move, and tech.    The former would be quicker and easier; the latter more elegant.

lb42:
My summary of the discussion appeared on TEI-L Wed 18 Jun, 19:00

joeytakeda:
Agreed. And I personally think this should be an attribute class with the membership that @lb42 suggests plus all members of model.stageLike. 

One question that arose in the discussion— and perhaps this is another ticket — is whether `@perf` must point to performance elements or if it could also point to an event (in the case where a performance is not explicitly documented in the text itself or if there are details of the performance omitted in the text but captured in a standoff event). I see no harm in loosening that constraint.

joeytakeda:
Council subgroup at F2F 2025 agrees that `@perf` should be available on all members of model.stageLike, `actor`, as well as `castList`, `castItem`, and `role`  and that it should be able to point to any kind of event (usually event or performance element(s))

lb42:
I am not sure why @perf is needed on castList or castItem, which would more usually be nested within an appropriate `<performance>` element. Also, how exactly should the attribute be provided -- via a (new?) class, or explicitly defined on the relevant element?  
My choice would be for the latter, and only as an addition to the attList for `<actor>` . This is simple, does the job, and can always be expanded upon later, in case of demonstrable need.

lb42:
Here's an example  where the cast list references two distinct performances :
~~~~
<performance>
    <p xml:id="L1025P1">First performed at the Strand Theatre, on the 18th of July, 1832.</p>
    <p xml:id="L1025P2">[also at the Royal Victoria in July 1834]</p>
  </performance>

   <castList>
    <head>Characters.</head>
    <castItem>
[...]

PR #2921, opened 2026-07-05
Create `att.performed` and update documentation (#2724)

joeytakeda:
This PR addresses #2724 by:

* Creating a new attribute class (`att.performed`) that defines the `@perf` attribute
* Making `actor` a member of `att.performed`
* Making all members of `model.stageLike` also members of `att.performed`
* Updating DR slightly to use @lb42 's example


P5/Source/Guidelines/en/DR-PerformanceTexts.xml
```diff
@@ -329,8 +328,39 @@ Or:
 </performance></egXML>
 <!-- La Machine Infernale, Jean Cocteau -->
  </p>
+  
+  <p>Performance-specific information (e.g. different actors in different performances) can be encoded using the <att>perf</att> attribute, provided by the <ident>att.performed</ident> attribute class:
+    <specList>
+      <specDesc key="att.performed"/>
+    </specList>
+   . The <att>perf</att> attribute allows one or more pointers to documentation (given through e.g. the <gi>performance</gi> or <gi>event</gi> elements) to associate an element (such as <gi>actor</gi>) with one or more performances:
```
sydb: While I concede that this _should_ not be wrong, only one other `<specList>` in the Guidelines is followed by a period, and it looks awful. (And it has no intervening whitespace). It is in [TC](https://www.tei-c.org/Vault/P5/4.9.0/doc/tei-p5-doc/en/html/TC.html) — to find it search for “but not rdgGrp”.

P5/Source/Guidelines/en/DR-PerformanceTexts.xml
```diff
@@ -329,8 +328,39 @@ Or:
...
+    
+  <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="DRPERF-egXML-perf" source="#DRPERF-eg-13.1">
+    <performance xml:id="L1025P1"><p>First performed at the Strand Theatre, on the 18th of July, 1832.</p></performance>
+    <performance xml:id="L1025P2"><p> Royal Victoria Theatre in July 1834</p></performance>
```
sydb: Remove leading space before “Royal”?

P5/Source/Guidelines/en/DR-PerformanceTexts.xml
```diff
@@ -772,7 +802,7 @@ Levant</head>
             <p>Ha! ha! Do. Do.</p>
           </sp>
           <sp> <speaker>Ellie</speaker>
-            <p>Please dont.</p>
+            <p>Please <sic>dont</sic>.</p>
```
sydb, line 805: Yay!

P5/Source/Specs/att.performed.xml
```diff
@@ -0,0 +1,31 @@
...
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="drama" type="atts" ident="att.performed">
+  <desc versionDate="2026-03-03" xml:lang="en">provides attributes to specify performance-based information.</desc>
+  <attList>
+    <attDef ident="perf" usage="opt">
+      <gloss versionDate="2007-07-04" xml:lang="en">performance</gloss>
+      <gloss versionDate="2007-12-20" xml:lang="ko">연기</gloss>
+      <gloss versionDate="2008-04-06" xml:lang="es">función</gloss>
+      <gloss versionDate="2008-03-30" xml:lang="fr">jeu</gloss>
+      <gloss versionDate="2007-11-06" xml:lang="it">rappresentazione</gloss>
+      <desc versionDate="2026-03-03" xml:lang="en">identifies the
```
sydb: The i18n folks have a mild preference for `<desc>` and `<gloss>` elements to be on a single line.

P5/Source/Specs/att.performed.xml
```diff
@@ -0,0 +1,31 @@
...
+        performance or performances in which this element occurred as
+        specified by pointing to one or more <gi>performance</gi> elements.</desc>
```
sydb: I would ditch “as specified … elements.” here, and instead add
~~~xml
<remarks versionDate="2026-07-05" xml:lang="en">
  <p>Performances are typically indicated with <gi>performance</gi> or <gi>event</gi> elements.</p>
</remarks>
~~~
or some such.

sydb (commented):
If we were to apply the Cayless Consideration (do not add something that has not been requested) we would add `@perf` to fewer elements. Personally it makes a lot of sense to me on `<actor>`, `<tech>`, `<caption>`, `<sound>`, `<stage>`, and `<move>`; I am not convinced about having it on `<castGroup>`, `<castItem>`, `<castList>`, and `<role>`, but nor can I say it would never be useful.

P5/Source/Guidelines/en/DR-PerformanceTexts.xml
joeytakeda: Nice catch -- I agree that, even if it isn't illegal, the trailing period is not ideal

P5/Source/Guidelines/en/DR-PerformanceTexts.xml
joeytakeda: :)

P5/Source/Specs/att.performed.xml
joeytakeda: Thanks for the suggestion, @sydb. I've shortened the `<desc>` (and put it on one line) and moved that info into a `<remarks>`. I updated the wording, using the remarks for `att.locatable` as a model:

```xml
  <remarks versionDate="2026-07-05"
        ident="att.performed-attr.perf-remarks" xml:lang="en">
        <p>The <att>perf</att> attribute contains one or more pointers to a canonical description of the performance, typically described by a <gi>performance</gi> or <gi>event</gi> element.</p>
      </remarks>
```

joeytakeda:
Having now addressed @sydb 's very helpful suggestions, I think this can now be merged (if at least one or two more reviewers agree)

martindholmes (commented):
@joeytakeda I think the descs you've commented out should actually remain in place; they're marked as being out of date by their versionDate, and leaving a slightly-misaligned (but accurate for the previous incarnation of the attribute) version in place is surely better than nothing.

Other than that, this looks good to me.

ebeshero:
Thanks @joeytakeda ! Merging now.

PR #2922, opened 2026-07-05
Updates for 3D object encoding

trishaoconnor:
This PR is for the prose updates necessary for completing #2816. 

- [x] Updated the description for the `att.coordinated` class to also mention a three dimensional coordinate system. 
- [x] Updated the description of `@points` to include mention of three dimensional coordinate system and mention z axis.
- [x] Update the description of `<surface>` to mention "three dimensional"
- [x] Update the description of `<zone>` to mention "three dimensional"
- [x] Include example of 3D encoding
- [x] Update section [12.2.2.1 Advanced Uses of `<surface>` and `<zone>`](https://tei-c.org/release/doc/tei-p5-doc/en/html/PH.html#PH-surfzone) in [Chapter 12: Representation of Primary Sources](https://tei-c.org/release/doc/tei-p5-doc/en/html/PH.html) with reference to encoding three dimensional objects.


ebeshero:
Fantastic work, @trishaoconnor! I will try to give all the edits a thorough reading before Council's P5 meeting on Tuesday, as this will be very timely for the release... (and see if I can help with the breaking tests if you need that)!

martindholmes:
I'm not sure what's happening with the tests here -- the errors seem very outdated:

"Schematron warning: The use of the `value` element
is deprecated, and will be considered invalid after 2021-07-09."

But tests pass locally for me, so I don't know what's happening. Any ideas?

P5/Source/Specs/surface.xml
```diff
@@ -78,8 +78,46 @@ celui-ci.</desc>
...
+          </zone>
+          <zone xml:id="inscription01_back"></zone>
+        </surface>
+      </sourceDoc>
+    </egXML>
+  </exemplum>
   <remarks ident="surface-remarks" versionDate="2011-11-20" xml:lang="en">
-    <p>The <gi>surface</gi> element  represents any two-dimensional space on
+    <p>The <gi>surface</gi> element represents any two-dimensional or three-dimensional space on
 some physical surface forming part of the source material, such as a piece of
```
martindholmes, line 83: Although the element name is `<surface>`, I wonder if "some physical surface" here should take account of the fact that we're now potentially working in three dimensions. Should we use something like "matrix" here?

sydb: Sorry, what writing are you imagining that someone would want to encode with TEI that is _not_ on a 2D or 3D surface? A holographic projection, perhaps? Words projected via VR goggles?

This sounds like it might be a case of making the description more confusing for all current users (and the vast majority of future users) for the potential benefit of an imagined future user that may never materialize.

sydb:
> I'm not sure what's happening with the tests here -- the errors seem very outdated:
> 
> "Schematron warning: The use of the `value` element is deprecated, and will be considered invalid after 2021-07-09."
>  
> But tests pass locally for me, so I don't know what's happening. Any ideas?

Tests fail locally for me. The problem, though, is not the Schematron warning, but above that seven outright duplicate `@xml:id` errors caused, I believe, by the duplication of an example in both Source/Guidelines/en/PH-PrimarySources.xml and Source/Specs/surface.xml. The `<egXML>` elements are missing their required `@xml:id` attributes, but occur on lines 808 and 82, respectively.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -658,67 +720,128 @@ used to define a polygon of any shape using this coordinate system:-->
...
+                     <surface ulx="50" uly="20" lrx="400" lry="280">
+                        <zone ulx="0" uly="0" lrx="500" lry="321">
+                           <graphic
+                              url="http://upload.wikimedia.org/wikipedia/commons/5/50/Handschrift.karlsruhe.blb.jpg"
+                           />
+                        </zone></surface>
+                  </facsimile></egXML>
+            </p>
+            <p>Finally, when encoding a three-dimensional object, it is essential to include the
+               Z-axis as well as the X and Y axes. You can represent the spatial information of a
```
sydb: One has a hyphen, the other two do not. I am not sure which is more correct, but I prefer with a hyphen. (And I note Wikipedia uses the hyphen.) In which case
~~~
Z-axis as well as the X- and Y-axes
~~~

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -658,67 +720,128 @@ used to define a polygon of any shape using this coordinate system:-->
...
+               three-dimensional object using the <att>points</att> attribute to construct a
+               polygonal shape. When you are using the <gi>surface</gi> and <gi>zone</gi> elements
+               within a three-dimensional context, the <gi>surface</gi> element represents the
```
sydb: I would prefer “in” to “within”, as the latter are more likely to be mis-construed (at least by this American) as being in an XML descendant.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -658,67 +720,128 @@ used to define a polygon of any shape using this coordinate system:-->
...
+               entire three-dimensional object. The <gi>graphic</gi> element can then be used to
+               reference the three-dimensional model of the object. All the zones of interest on the
```
sydb: reference **a** three-dimensional model …

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -658,67 +720,128 @@ used to define a polygon of any shape using this coordinate system:-->
...
+               object can be defined using the <gi>zone</gi> element. For example, if a monument
+               features both a relief and an inscription, the <gi>zone</gi> element can be used to
+               delineate each feature respectively. You can use the <att>points</att> attribute on
+                  <gi>zone</gi> element to represent the three-dimensional spatial information for
```
sydb: I prefer either “… use the `@points` attribute on `<zone>` to represent …” or “… use the `@points` attribute on (the|a|any) `<zone>` element to represent …”.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -728,10 +851,10 @@ used to define a polygon of any shape using this coordinate system:-->
       <head>Scope of Transcriptions</head>
 
       <p>When transcribing a primary source, whether using <gi>text</gi> or <gi>sourceDoc</gi>,
-         scholars may wish to record information concerning
-         individual readings of letters, words, or larger units.They
-         may also wish to include other editorial material, such as comments on the status or
-         possible origin of particular readings, corrections, or text supplied to fill lacunae. </p>
+         scholars may wish to record information concerning individual readings of letters, words,
+         or larger units.They may also wish to include other editorial material, such as comments on
```
sydb: May as well insert the missing space between “units.” and “They”.

P5/Source/Specs/att.coordinated.xml
```diff
@@ -60,7 +60,7 @@ rectangular space.</desc>
       <datatype><dataRef key="teidata.numeric"/></datatype>
     </attDef>
     <attDef ident="points">
-      <desc versionDate="2011-12-17" xml:lang="en">identifies a two dimensional area by means of a series of pairs of numbers, each of which gives the x,y coordinates of a point on a line enclosing the area.</desc>
+      <desc versionDate="2011-12-17" xml:lang="en">identifies a two or three dimensional area by means of a series of pairs of numbers, each of which gives the x,y, or z coordinates of a point on a line enclosing the area.</desc>
```
sydb, line 63: 1. When used for a three dimensional area isn’t it a sequence of triples, not pairs?
2. I’m wondering (aloud) if the “, each of which …” should be in a `<remarks>` rather than here.

sydb (changes requested):

trishaoconnor:
Error message from the initial builds failing:
```
   <ERROR>id 'inscription01' used more than once</ERROR>
   <ERROR>id 'inscription01_front' used more than once</ERROR>
   <ERROR>id 'inscription01_front_relief' used more than once</ERROR>
   <ERROR>id 'inscription01_front_text' used more than once</ERROR>
   <ERROR>id 'inscription01_ft_l2' used more than once</ERROR>
   <ERROR>id 'inscription01_ft_l3' used more than once</ERROR>
   <ERROR>id 'inscription01_back' used more than once</ERROR>
```

Have now updated the `xml:id`s and also deleted the example from `<surface>`, but the builds are still failing.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -794,18 +794,20 @@ used to define a polygon of any shape using this coordinate system:-->
...
-               features both a relief and an inscription, the <gi>zone</gi> element can be used to
-               delineate each feature respectively. You can use the <att>points</att> attribute on
-                 a <gi>zone</gi> element to represent the three-dimensional spatial information for
-               these zones of interest. <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="PH-surfzone-egXML-jo" xml:lang="und" source="#UND">
+            <p>Finally, if you are using the <gi>surface</gi> and <gi>zone</gi> elements to encode a
+               three-dimensional object, it is essential to include the Z-axis as well as the X- and
+               Y- axes. You can represent the spatial information of a three-dimensional object
+               using the <att>points</att> attribute to construct a polygonal shape. When you are
+               using the <gi>surface</gi> and <gi>zone</gi> elements in a three-dimensional context,
+               the <gi>surface</gi> element represents the entire three-dimensional object. The
```
sydb: But it doesn’t. It represents the _surface_ of the entire 3D object. Imagine you got the chance to write “fascist!” on one of the tires of Elon Musk’s bicycle. If you were to encode this, the `<surface>` element would represent the surface of the tire, not its inner tube.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -794,18 +794,20 @@ used to define a polygon of any shape using this coordinate system:-->
...
+                  <gi>graphic</gi> element can then be used to reference a three-dimensional model
+               of the object. All the zones of interest on the object can be defined using the
```
sydb, line 804: I presume that `<graphic>` could also reference a photograph of the object, no? Do we intend “3D model” to include photographs? My instinct is that a photograph is not a model, but rather an image, of an object; but others may disagree.

P5/Source/Guidelines/en/PH-PrimarySources.xml
```diff
@@ -794,18 +794,20 @@ used to define a polygon of any shape using this coordinate system:-->
...
-               in a three-dimensional context, the <gi>surface</gi> element represents the
-               entire three-dimensional object. The <gi>graphic</gi> element can then be used to
-               reference a three-dimensional model of the object. All the zones of interest on the
-               object can be defined using the <gi>zone</gi> element. For example, if a monument
-               features both a relief and an inscription, the <gi>zone</gi> element can be used to
-               delineate each feature respectively. You can use the <att>points</att> attribute on
-                 a <gi>zone</gi> element to represent the three-dimensional spatial information for
-               these zones of interest. <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="PH-surfzone-egXML-jo" xml:lang="und" source="#UND">
+            <p>Finally, if you are using the <gi>surface</gi> and <gi>zone</gi> elements to encode a
+               three-dimensional object, it is essential to include the Z-axis as well as the X- and
```
sydb, line 798: In chapter WD we use “z-axis” with a lowercase ‘z’.
I do not know which, if either, is “correct”. I have seen both used often. (Heck, Wikipedia uses both in the same [article](https://en.wikipedia.org/wiki/Axes_conventions).)

sydb: BTW, my **very** mild preference would be to uppercase the one in WD, as using lowercase in “x- and y-axes” does not look as good as uppercase IMHO.

sydb (approved):
Minor nit-picks would make it better, but nothing that should stop it from being published, I don’t think.

trishaoconnor:
Thank you @sydb for your feedback. These suggestions are most helpful. I've applied those changes. 

- [x] Update the WD chapter to change the z-axis instances from lowercase to uppercase.

P5/Source/Guidelines/en/PH-PrimarySources.xml
trishaoconnor: Thank you for finding those examples in the WD chapter. I've updated those lowercase instances to uppercase for consistency. The Y-axis was also mentioned in the WD chapter and written in uppercase, so it was especially weird to have the z-axis in lowercase.

Issue #2873, opened 2026-03-10
Add `att.locatable` (`@where`) to `<death>` and `<birth>`


frederik-elwert:
Currently, the element [`<event>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-event.html) has the attribute class `att.locatable`, which consists of a single attribute, `@where`. However, the two elements [`<death>`](https://tei-c.org/release/doc/tei-p5-doc/de/html/ref-death.html) and [`<birth>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-birth.html) lack this class.

The prose text of the guidelines [describes](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ND.html#NDPERSEpe) `<birth>` and `<death>` as two specialized elements for particular events. Thus, it would be expected that they can hold the same attributes as the `<event>` element itself. And particularly, recording the place of birth and death makes immediate sense.

We are currently designing the data structure for a prosopography file, and there recording the place of death is required. Thus, we would make immediate use of this attribute.

martindholmes:
+1 from me.

ebeshero:
Council on 2026-04-10 agrees this should be implemented.

joeytakeda:
Thanks for this request, @frederik-elwert ! I've now done this in PR #2923 , but (as I note there) we could use some more documentation and examples of use. I also think that we could make many more things members of att.locatable (floruit comes to mind, but so do most of the model.persStateLike elements: occupation, persName, residence, socecStatus, affiliation?), but that can be a new ticket.

ebeshero:
Completed (as noted above) with https://github.com/TEIC/TEI/pull/2923 .

PR #2923, opened 2026-07-05
Make birth and death members of att.locatable (addresses #2873)

joeytakeda:
Fairly straightforward addition of att.locatable to birth and death. I've added a couple of examples (both derived from The Lyon in Mourning where we incidentally had customized the `<birth>` and `<death>` already to have `@where`!), but did not touch any of the prose or examples in ND.

### DIFF
--- a/P5/Source/Specs/birth.xml
+++ b/P5/Source/Specs/birth.xml
@@ -83,7 +83,7 @@
   </exemplum>
   <exemplum xml:lang="en">
     <!--Example derived from https://lim.dhil.lib.sfu.ca/BURKE1.html-->
-    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-birth-egXML-where">
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-birth-egXML-wh">
       <birth where="https://www.wikidata.org/wiki/Q662158">Born on the Island of <placeName>North Uist</placeName></birth>
     </egXML>
   </exemplum>
--- a/P5/Source/Specs/birth.xml
+++ b/P5/Source/Specs/birth.xml
@@ -16,6 +16,7 @@
     <memberOf key="att.datable"/>
     <memberOf key="att.dimensions"/>
     <memberOf key="att.editLike"/>
+    <memberOf key="att.locatable"/>
     <memberOf key="att.naming"/>
     <memberOf key="att.typed"/>
     <memberOf key="model.personPart"/>
@@ -78,6 +79,12 @@
       <birth when="1960-12-10">In a small cottage near <name type="place">Aix-la-Chapelle</name>,
         early in the morning of <date>10 Dec 1960</date>
          </birth>
+    </egXML>
+  </exemplum>
+  <exemplum xml:lang="en">
+    <!--Example derived from https://lim.dhil.lib.sfu.ca/BURKE1.html-->
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-birth-egXML-where">
+      <birth where="https://www.wikidata.org/wiki/Q662158">Born on the Island of <placeName>North Uist</placeName></birth>
     </egXML>
   </exemplum>
   <listRef>
--- a/P5/Source/Specs/death.xml
+++ b/P5/Source/Specs/death.xml
@@ -71,10 +71,10 @@
     </egXML>
   </exemplum>
   <exemplum xml:lang="en">
-    <!--Example taken from https://lyoninmourning.dhil.lib.sfu.ca/BURKE1.html-->
-    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-death-egXML-where">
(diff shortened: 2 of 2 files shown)


P5/Source/Specs/birth.xml
```diff
@@ -80,6 +81,12 @@
          </birth>
     </egXML>
   </exemplum>
+  <exemplum xml:lang="en">
+    <!--Example derived from https://lim.dhil.lib.sfu.ca/BURKE1.html-->
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-birth-egXML-where">
```
sydb: The value of `@xml:id` should be "gi-birth-egXML-??", where each ‘?’ is a lowercase letter, a–z, which you can just pick. (If you pick two that are already present, there will be an error and you will have to pick a different pair.)

P5/Source/Specs/birth.xml
```diff
@@ -80,6 +81,12 @@
          </birth>
     </egXML>
   </exemplum>
+  <exemplum xml:lang="en">
+    <!--Example derived from https://lim.dhil.lib.sfu.ca/BURKE1.html-->
```
sydb, line 85: I was going to complain that the `<egXML>` should have a `@source` instead of this comment, but having looked at that web page I think I understand why not. 🙂

P5/Source/Specs/death.xml
```diff
@@ -69,6 +70,12 @@
       <death when="1960-12-10">Passed away near <name type="place">Aix-la-Chapelle</name>, after suffering from cerebral palsy. </death>
     </egXML>
   </exemplum>
+  <exemplum xml:lang="en">
+    <!--Example taken from https://lyoninmourning.dhil.lib.sfu.ca/BURKE1.html-->
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-death-egXML-where">
```
sydb: Given that `<birth>` was from BURKE1, could we use something else for this? Perhaps an example that hits closer to home and includes just structured information?
~~~xml
<death when="2024-08-16" where="#US-NM-039-5467024"/>
~~~
 And maybe its prose equivalent?
~~~xml
<death when="2024-08-16" where="#US-NM-039-5467024">Fri 16 Aug 24 in Española, NM</death>
~~~
(You might prefer using a prefix like `gn:` instead of `#`, I don’t care.)

sydb (commented):
OK as is, but improvements can easily be made.

P5/Source/Specs/death.xml
joeytakeda: I like this example very much indeed -- I'm happy with the just structured info version plus it's nice to have more examples of prefixes too.

P5/Source/Specs/birth.xml
joeytakeda: Ah got it --we should probably codify that in p5odds, eh?

sydb: I suppose so, but it is not a simple algorithm. (And in fact, I am not sure I even _kept_ the program that created most the IDs while we were in Buenos Aires.) It is something like

~~~
  IF ancestor::*/@xml:id THEN 
    take its value
  ELSE
    take mapped name of root element (one of "class", "data", "gi")
    append hyphen
    append root element’s @xml:id
    IF we are in a classSpec//attDef THEN
      append “-attr-”
      append ancestor::attDef/@ident
append “-[element_name]-”
append random two-letter disambiguation
~~~
Where [element_name] is the name of the elment for which you are making up an ID (figure, table, or egXML, I think).

P5/Source/Specs/death.xml
sydb: Excellent!
I will have to check whether it should be Española or Sante Fe, but we can merge it now and I will update later iff needed.

P5/Source/Specs/death.xml
```diff
@@ -69,6 +70,12 @@
       <death when="1960-12-10">Passed away near <name type="place">Aix-la-Chapelle</name>, after suffering from cerebral palsy. </death>
     </egXML>
   </exemplum>
+  <exemplum xml:lang="en">
+    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="gi-death-egXML-wh">
+      <death when="2024-08-16" where="gn:US-NM-039-5467024"/>
+    </egXML>
+    <p>In this example, the value of <att>where</att> demonstrates the use of private URI scheme and custom prefix (<val>gn:</val>), which would typically be defined by a <gi>prefixDef</gi> in the document's <gi>teiHeader</gi>.</p>
```
ebeshero, line 77: This explanation is super helpful! Looks good to me.

ebeshero:
@joeytakeda @sydb I just reviewed this and it looks fantastic to me, but I wondered if you're both happy with this for the release? Syd's last comment that "improvements can easily be made" had me wondering, but I'm really thinking we should just merge this now.

ebeshero (approved):
The revisions to this PR look good to me, and the explanations of how to use the `@where` with a private URI scheme are very helpful.

PR #2927, opened 2026-07-13
use item rather than sequence comparison to save ?

### LINKED ISSUE #2926: maxGeneralEntitySizeLimit is mean to my laptop
I tried building P5 on my laptop this morning. It failed, so I went ahead and built it on my desktop (w/o problem). When I had some more time this evening I tried again on my laptop and found the error was a bit mysterious (at least, to me):

> Error on line 48237 column 1188 of Schematron1.xml: SXXP0003   Error reported by XML parser: JAXP00010003: The length of entity "[xml]" is "100,001" that exceeds the "100,000" limit set by "jdk.xml.maxGeneralEntitySizeLimit".
> SXXP0003   Error reported by XML parser: JAXP00010003: The length of entity "[xml]" is "100,001" that exceeds the "100,000" limit set by "jdk.xml.maxGeneralEntitySizeLimit".

I think of this a kinda high priority ticket, as this problem occurred on the laptop I plan to take to Vancouver.

sydb:
This is a fix to #2926, although I do not actually understand _why_ it fixes it.

I don’t know what maxGeneralEntitySizeLimit is nor how to set it. But the error was flagged on one of the nearly 221,000 occurrences of `position() &gt; 1`. The vast majority (if not all) of these are in the p5.odds test "only-one-each-lang". On a hunch, and reasoning it is better to use item comparison (` gt `) the sequence comparsion (`>`, often written as `&gt;`) anyway, I changed all 18 sequence comparisons to item comparisons and the problem went away.


peterstadler (approved):
This looks technically sane and valid to me (given that no side of the comparison will be a sequence at any time)

martindholmes:
This also seems reasonable to me. @joeytakeda If you agree, go ahead and merge.

joeytakeda:
This looks good to me! Merging now

PR #2928, opened 2026-07-14
Added paragraph advising to redefine defaultExceptions on schemaSpec …

### LINKED ISSUE #2837: Conflict with xenoData when introducing elements in new namespaces in ODD customizations
I discovered this problem while working with students to customize the CBML ODD (for Comic Book Markup Language). 

The issue appears  when:
* we introduce elements in a new namespace (e.g. `http://www.cbml.org/ns/1.0`) 
*  we wish to be checking IDs/IDREFs (with this setting turned on in oXygen)

The output Relax NG in XML syntax in these conditions generates this error on `xenoData`:

```xml
<define name="anyElement_xenoData_1">
      <element>
         <anyName>
            <except>
               <nsName ns="http://www.tei-c.org/ns/1.0"/>
               <name ns="http://www.tei-c.org/ns/Examples">egXML</name>
            </except>
         </anyName>
         <zeroOrMore>
            <attribute>
               <anyName/>
            </attribute>
         </zeroOrMore>
         <zeroOrMore>
            <choice>
               <text/>
               <ref name="anyElement_xenoData_1"/>
            </choice>
         </zeroOrMore>
      </element>
   </define>
```

The Relax NG error is on the <element> line here, and reads, “conflicting ID-types for attribute “id” from namespace “http:// [www.w3.org/XML/1998/](http://www.w3.org/XML/1998/) namespace” of element “panel” from namespace “http:// [www.cbml.org/ns/1.0](http://www.cbml.org/ns/1.0)”. (When I experimented with removing the cbml:panel definition in the ODD, the error would emerge again on this definition of anyElement_xenoData_1, just mentioning a different CBML element.)

To temporarily solve the problem, simply removing `<xenoData>` from the teiheader module entirely, or modifying `<xenoData>` in an elementSpec to exclude the CBML namespace worked. So this is my current patch (thanks to @raffazizzi's advice on the TEI Slack): 

```xml
<elementSpec ident="xenoData" mode="replace">
    <content>
        <anyElement except="http://www.cbml.org/ns/1.0"/>
    </content>
</elementSpec>
```

However, we think this should not be necessary. Is there a solution to the problem that doesn't compel ODD customizers to modify the `<xenoData>` element defintion when we introduce elements in custom namespaces? The `ANY` or `anyElement` in our content model for `<xenoData>` seems to be the problem. @raffazizzi suggests in Slack that perhaps `<anyElement>` can be adjusted by the ODD processor to exclude new namespaces introduced in an ODD customization?

Here is [my customization of CBML](https://github.com/newtfire/textEncoding-Hub/blob/master/Class-Examples/CBML/CBML-ODD/cbml.odd) for reference on this ticket. (It includes the patchwork to revise `<xenoData>` in an elementSpec at the very end of the file (line 376).

raffazizzi:
…when adding an element in a non-TEI namespace. Closes #2837


P5/Source/Guidelines/en/USE.xml
```diff
@@ -658,6 +658,10 @@ that practice. -->
             <!-- ... -->
           </schemaSpec></egXML>
       </p>
+      <p>When a customization introduces a new element, and therefore a new namespace, it is recommended to redefine <att>defaultExceptions</att> on <gi>schemaSpec</gi> with any new namespaces because elements matched by <gi>anyElement</gi> (when present in a customization) must exclude any element that may carry an <att>xml:id</att> attribute. This is particularly important if the target schema language is RELAX NG and ID/IDREF checking is desired and failure to do so results in <soCalled>conflicting ID-types</soCalled> validation errors.
```
sydb: 1.  I would prefer “it is recommended” be toned down. (This may be me just repeating myself, I don’t remember, but _I_, for one, do _not_ recommend this method.)
2. “it is recommended to” sounds a bit clumsy, anyway.
3. While not horribly long, this first sentence is quite a long one. So maybe chopping it up is a good plan?
4. It is kinda weird that the first clause switches from the singular (“a new namespace”) to the plural (“any new namespaces”).
5. The parenthetical “(when present in a customization)” seems to me to be mildly misleading. We mean (and I daresay the parenthetical technically says) when present in the output RELAX NG schema produced by ODD software to perform validation of files against the customized version of TEI. But a reader might think it means the customization ODD file itself, which likely does not include `<anyElement>`. (The customization RELAX NG likely does include `<anyElement>` because it is in the content model of `<xenoData>`.) I am not sure my suggested wording, below, is much better, though. Feel free to tweak.
6. The `<anyElement>`s need to exclude any element which may bear an ID attribute (even if it is an optional attribute, and even if it is not called `@xml:id`). So in fact, if the newly defined element(s) are not members of att.global and do not define their own ID attrs, one could avoid this problem. (Bad thing to do, IMHO.) That said, I am not sure we should bother differentiating between an ID attr and `@xml:id` here. We generally do not use the more general phrasing, and in most of the places we do I think it is a corrigible error. (Here it would not be an error, but it might still be confusing, idunno.)
7. The phrasing “is particularly important” is misleading — it is only important when RELAX NG is used in DTD compatibility mode. (Actually, only when the “ID/IDREF/IDREFS” feature of compatibility mode is being used, but I don’t think we need to hammer that detail.)
8. “… ID/IDREF checking is desired” — this is not the name of any feature of RELAX NG or DTD compatibility mode, it is what oXygen _calls_ the “ID/IDREF/IDREFS” feature of DTD compatibility mode. Other processors have other names. (E.g., `jing`, in the description of the `-i` switch calls it “checking of ID/IDREF/IDREFS”.) We should not be implicitly reinforcing that oXygen is the only way to edit TEI files. 🙂 On the other hand, one might say the same of the `jing` error message “conflicting ID-types”, but I do not know what the msg is using any other validator. (I just checked `rnv`, and it does not give a msg at all, meaning it probably does not support this feature.)
9. “… and failure to do so results …” sort of implies that failure to do so might not result in this error msg.
10. I do not think “conflicting ID-types” is a validation error, rather it is a message about an incorrect schema. (I don’t think a processor _can_ valid an XML document against a RELAX NG schema unless it is a correct schema. Note that “correct” here is a technical term defined in the [spec](https://relaxng.org/spec-20011203.html), and consequently in ISO 19757-2. Of course the confusing bit (at least to me) is that the schema _is_ correct, it is just that DTD compatibility mode, particularly the ID/IDREF/IDREFS checking feature, adds an additional constraint to the schema, and without a proper exception, the definition of `<anyElement>` fails that additional constraint.)

So perhaps something like
~~~xml
<p>When a customization introduces new elements, and therefore one or
more new namespaces, users may wish to specify the
<att>defaultExceptions</att> on <gi>schemaSpec</gi> such that the new
namespace or namespaces are included. This is because if an element is
allowed to have an ID-type attribute (generally <att>xml:id</att>) and
can be matched by an <gi>anyElement</gi> (e.g., the
<gi>anyElement</gi> used as the content of <gi>xenoData</gi>), the
resulting RELAX NG schema cannot be used with the ID/IDREF/IDREFS
checking feature of DTD compatibility mode.<note place="bottom">Called
<q>Check ID/IDREF</q> in oXygen; controlled by the <code>-i</code>
switch of the <name>jing</name> processor.</note> Specifying a
namespace on <att>defaultExceptions</att> prevents elements in that
namespace from being matched by an <gi>anyElement</gi> element which
does not override the default exception.</p>
~~~
While I like this better (it addresses my concerns, above), it does not actually mention
~~~xml
<soCalled>conflicting ID-types</soCalled>
~~~
which is what users will likely actually see and be confused by. So more improvements are probably in order.

ebeshero: @raffazizzi Thanks for this helpful explanation of when to redefine `@defaultExceptions`. Here are some edits just to make it a little easier to read: 

- You start the first sentence with a single new element in a single new namespace, so we should recommend that `@defaultExceptions` be updated with "the new namespace" or "that new namespace" (rather than "any new namespaces"). 

- Tweak the end of the last sentence just to put the error language at the end:

"failure to do so results in validation errors with `<soCalled>conflicting ID-types</soCalled>`."

ebeshero (changes requested):

P5/Source/Guidelines/en/USE.xml
```diff
@@ -658,6 +658,10 @@ that practice. -->
...
+        <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="MDNS-egXML-de" xml:lang="und" source="#UND"><schemaSpec ns="http://www.example.org/ns/nonTEI" ident="example" defaultExceptions="http://www.tei-c.org/ns/1.0 http://www.tei-c.org/ns/Examples http://www.example.org/ns/nonTEI">
```
sydb: Wouldn’t the example read (a little) better if the new namespace were the first of the three URIs listed in `@defaultExceptions`?

sydb (changes requested):

P5/Source/Guidelines/en/USE.xml
ebeshero: I was thinking at first to add a comma before "because", but the explanation is necessary to the whole sentence (so no comma).

P5/Source/Guidelines/en/USE.xml
ebeshero: Yes--agreed!

P5/Source/Guidelines/en/USE.xml
ebeshero: @sydb @raffazizzi I think it's absolutely important to bring forward the "conflicting id-types" validation error since that message from jing is the real problem people need to solve. I like @sydb's longer explanation of why we see that error. Is ID/IDREF/IDREFS checking *only* part of "DTD compatibility mode", or is it not simply a fundamental aspect of Relax NG validation? (Is "DTD compatibility mode" necessary to mention in main text or should it go in the note?)  

If mentioning oXygen in the note, I think we'd specify the "oXygen XML Editor" as software name. (Is that consistent with how we refer to it elsewhere?)

ebeshero: Okay, I was just reading about how DTD and Relax NG are connected on ID/IDREF/IDREFs. If I understand correctly it's just that Relax NG inherits ID/IDREF(S) from DTD's method of validation. (The phrase "DTD compatibility mode" has made me think "this is something I set in order for my schema to be supported somehow in old DTD." That is not really what it means, but I am just realizing that now.)

George Bina has a super helpful explanation of ID/IDREF(S) inheritance from DTD in this comment: https://github.com/relaxng/jing-trang/issues/211#issuecomment-243983664

sydb: I agree it is a good idea to mention “conflicting ID-types”, but it is only the message that `jing` (and thus oXygen) uses, other software would be welcome to use a different msg. (I don’t know of any other software that has a msg, though — as I mentioned, `rnv` does not seem to do DTD compatibility mode, so this doesn’t come up.)

Yes, ID/IDREF/IDREFS checking is only part of [DTD compatibility](https://relaxng.org/compatibility-20011203.html#id), it is _not_ a fundamental aspect of RELAX NG validation. (MURATA Makoto and James Clark do not believe this kind of testing is the purview of a schema language, IIRC. In any case, the regular expression `\bid(refs?)?\b` (case insensitive) does not appear in the content of the spec.) As far as I could tell, this feature is not a part of ISO 19757-2 at all.

I don’t know about the footnote, either, but thought it was a good way to get the name as oXygen calls it in there, which (even though I grump) may be a good idea, as a sizeable percentage of users will encounter it that way. (And yes, I think @ebeshero is right, we should call it the “oXygen XML Editor” both here and in the colophon, where we currently call it “the oXygen editor”.)

raffazizzi:
I made a few changes that I believe addresses your concerns, but have some responses below. 

> 1. I would prefer “it is recommended” be toned down.

I did this, but with something more encouraging than "wish"

> 7. The phrasing “is particularly important” is misleading — it is only important when RELAX NG is used in DTD compatibility mode.

I think the "and" addressed this already. I adjusted a bit, but I don't think it's misleading when the sentence is read start to finish

Finally, I don't think Oxygen should be explicitly mentioned; we should be able to explain this issue related to formats and standards without mentioning specific implementing software.

P5/Source/Guidelines/en/USE.xml
```diff
@@ -658,10 +658,20 @@ that practice. -->
             <!-- ... -->
           </schemaSpec></egXML>
       </p>
-      <p>When a customization introduces a new element, and therefore a new namespace, it is recommended to redefine <att>defaultExceptions</att> on <gi>schemaSpec</gi> with any new namespaces because elements matched by <gi>anyElement</gi> (when present in a customization) must exclude any element that may carry an <att>xml:id</att> attribute. This is particularly important if the target schema language is RELAX NG and ID/IDREF checking is desired and failure to do so results in <soCalled>conflicting ID-types</soCalled> validation errors.
-        <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:id="MDNS-egXML-de" xml:lang="und" source="#UND"><schemaSpec ns="http://www.example.org/ns/nonTEI" ident="example" defaultExceptions="http://www.tei-c.org/ns/1.0 http://www.tei-c.org/ns/Examples http://www.example.org/ns/nonTEI">
+      <p>When a customization introduces a new elements, and therefore one or more new namespaces, users may want to consider redefining
```
ebeshero: @raffazizzi Okay, you opted to pluralize. Just remove the 'a' before "new elements".

P5/Source/Guidelines/en/USE.xml
```diff
@@ -658,10 +658,20 @@ that practice. -->
...
+        <att>defaultExceptions</att> on <gi>schemaSpec</gi> such that the new
+        namespace or namespaces are included. This is because elements matched by 
+        <gi>anyElement</gi> (e.g. the one used as the content of <gi>xenoData</gi>) must exclude any element that may carry 
+        an ID-type attribute (generally <att>xml:id</att>). 
+        This is particularly important if the target schema language is RELAX NG and one wishes to enable ID/IDREF/IDREFS compatibility mode: not adding
```
ebeshero, line 666: Well phrased!

ebeshero (commented):
Just one tiny grammatical revision, and I can add it.

raffazizzi:
Thanks for the fixes @ebeshero !

ebeshero (approved):
With this tiny grammatical edit that I just pushed, I am happy to approve the PR and will happily merge it later today.

sydb:
OK. Sigh.

The only thing I am actually going to argue for prior to merging is as follows.

The sentence

> This is because elements matched by `<anyElement>` (e.g. the one used as the content of `<xenoData>`) must exclude any element that may carry an ID-type attribute (generally `@xml:id`).

asserts that `<anyElement>` must exclude any element that meets those criteria. But those elements only need to be exluded **if** your validator is using DTD compatibility mode (particularly the ID/IDREF/IDREFS feature of DTD compatibility). If you do not want to use this feature, you can happily live your life without ever using `anyElement/@except` or `schemaSpec/@defaultExceptions` and without ever getting “conflicting ID-types for attribute "id"”.

So I think a bit more wordsmithing is in order.

ebeshero:
@sydb @raffazizzi Okay, but I don't think it needs *very much* wordsmithing to address the point. How about this revision?

```xml
This is because elements matched by <gi>anyElement</gi> (e.g. the one

used as the content of <gi>xenoData</gi>) must exclude any element that

may carry an ID-type attribute (generally <att>xml:id</att>) when the

target schema language is RELAX NG and one wishes to enable

ID/IDREF/IDREFS compatibility mode. Not adding the new namespaces to

<att>defaultExceptions</att> will result in <soCalled>conflicting ID-types</soCalled> errors.
```

raffazizzi:
Looks good to me!

Issue #2752, opened 2025-08-27
<am> should allow @place


joeytakeda:
In the Lyon in Mourning project, we (@kaitlynmacinnis and I) found a number of instances where the author uses an intra-linear cross mark to denote that an abbreviation has taken place (screenshot below). 

<img width="594" height="457" alt="Image" src="https://github.com/user-attachments/assets/8e817c22-6203-47a0-af66-bd3c3f9dd764" />

<details>
<summary>Sample Encoding</summary>

```xml
<list>
<!--...-->
<item><label>2.</label> The Gentleman killed by the Dra<pc force="weak">-</pc>
<lb/>goons in Inverness was James Aber<pc force="weak">-</pc>
<lb/>deen <choice><abbr>mer<am place="above">+</am>t</abbr><expan>merchant</expan></choice> in Old <abbr>Abd</abbr> who had a Wife
<!--[...]--></item>
<!--[...]-->
<item><label>4.</label> When <choice><abbr>Gr<am place="above">+</am>all</abbr><expan>General</expan></choice> Hally commanded that
<lb/><!--[...]--></item>
</list>
```
</details>

 While this could be handled with an `<add>` or some other element, we think this is truly a case where we have an abbreviation marker that is slotted above the line (a kind of abbreviation that is already exemplified by `abbr/@type`). 

So we would like to suggest that `<am>` be added to `att.placement` (at the very least —even better would be `<am>` becoming a member of `att.transcriptional` [c.f. #2665 for some discussion about `att.placement` and `att.transcriptional`])

joeytakeda:
<img width="889" height="160" alt="Image" src="https://github.com/user-attachments/assets/d7e44beb-01a4-47c3-ab0e-0991620c6b01" />

Here's another example with a different symbol from the same manuscript.

jamescummings:
More than happy to agree that `<am>` would benefit from claiming membership of `att.transcriptional`.

sydb:
Makes sense to me that `<am>` should be in att.placement. Membership in att.transcriptional overall seems reasonable:

- `@evidence`: although has never been requested, makes sense
- `@instant`: although has never been requested, makes sense
- `@place`: makes sense
- `@hand`: makes sense
- `@status`: umm … I guess _someone_ could make it make sense (but I can’t)
- `@cause`: does not make sense (the cause is always, by definition, marking an abbreviation)
- `@seq`: although has never been requested, makes sense

trishaoconnor:
[...]

PR #2946, opened 2026-08-16
Add `<am>` to `att.transcriptional`

trishaoconnor:
This PR addresses #2752 and adds `<am>` to `att.transcriptional`.


trishaoconnor:
Build failing due to duplicate evidence attribute. `<am>` currently inherits the `evidence` attribute from `att.editLike`. Adding `<am>` to `att.transcriptional` means that it will inherit both the `evidence` and `instant` attributes from `att.editLike`, since `att.editLike` is a member of `att.transcriptional`. So I'm removing `att.editLike` from the `<am>` spec file to address the duplication error.

torstenroeder (approved):
Perfect :)

sydb:
Does no one else want to remove `@cause`?

torstenroeder:
`@cause` on `<am>` seems, however redundant especially when within `<abbr>`, not essentially wrong to me; I would just leave it there and not exclude it.

Issue #2937, opened 2026-07-28
non-deterministic sample DTD content model


sydb:
Egads!
The sample DTD content model in the [interleave tagdoc](https://www.tei-c.org/Vault/P5/4.12.0/doc/tei-p5-doc/en/html/ref-interleave.html) is not deterministic!

How did that slip by, you ask? Because oXygen (apparently) does not flag it as an error; I do not know why not. But it is an (egregious, corrigible) error.

Basically the DTD parser cannot determine which branch of the content model is being matched. E.g., if the content in your XML file is just a `<persName>` followed by a `<note>`, it could not tell whether you were matching the 2nd clause (`( persName, note)`) or 6th clause (`( persName, note, ( placeName | location ) )`).

In case anyone wants to test it against your software, here is a tiny test file.

[sample_DTD_problem_demo.zip](https://github.com/user-attachments/files/30482094/sample_DTD_problem_demo.zip)

P.S. I have labeled this MEDIUM priority, and given it a priority field of HIGH a) just to demonstrate we have two ways of saying the same thing, which might conflict; and b) to express that this is a medium-high priority issue. 🫤

ebeshero:
Fixed with #2948

PR #2948, opened 2026-08-16
fix ambiguous sample DTD content model

sydb:
This PR addresses #2937 by replacing the sample DTD content model in the `<interleave>` tagdoc with a deterministic (aka non-ambiguous) version of same.

Recommended review process:

- Download the [error demonstration file](https://github.com/user-attachments/files/30482094/sample_DTD_problem_demo.zip) from the ticket itself; this file contains a copy of the current content model
- Convince yourself the content model in the downloaded file matches that which is in the current tagdoc for interleave (on release or dev)
- Convince yourself the content model is ambiguous, either through inspection or by validating that file (with something other than a modern version of oXygen)
-  Download the [fixed_DTD_demo.zip](https://github.com/user-attachments/files/31113169/fixed_DTD_demo.zip), the same file with the new version of the content model
- Convince yourself the content model in that new file matches that which is in the tagdoc for interleave on this branch
- Convince yourself that this new version is deterministic, either through inspection or validating that file (with something other than a modern version of oXygen)

### DIFF
--- a/P5/Source/Specs/interleave.xml
+++ b/P5/Source/Specs/interleave.xml
@@ -37,18 +37,24 @@
     <gi>note</gi> element, in any order.</p>
     <p>To express this content model in a DTD, the original schema
     language for XML, one might use
-    <eg xml:space="preserve">(   persName
-      | ( persName, note )
-      | ( persName, ( placeName | location ) ) 
-      | ( note, persName )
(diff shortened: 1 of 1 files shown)


ebeshero:
Council approves this at F2F after @sydb !

Issue #2389, opened 2023-01-26
rs should be datable 


ebeshero:
[Discussion on the TEI listserv of 25 January 2023](https://listserv.brown.edu/cgi-bin/wa?A1=ind2301&L=TEI-L#9) proposes that `<rs>` should be a member of att.datable.

skurzinz:
The example that @KollatzThomas gave in the original post was 

> From time to time we would like to encode e.g. an event with a date-attribute: 
> `<rs type="event" subtype="conference" from="2023-09-03" to="2023-09-08">TEI conference</rs>`

Given the recent changes proposed in #2427, I would rather do something like 

```
<standOff><listEvent><event xml:id="TEIMEC2023" from="2023-09-03" to="2023-09-08">
<eventName type="alternative">TEI Conference 2023</eventName>
<eventName type="official">Encoding Cultures – Joint MEC and TEI Conference 2023</eventName>
</event></listEvent></standOff>
```
(if #2427 is not considered, replace `eventName` with `label` or `head`)

and point to that in the `<rs type="event" ref="#TEIMEC2023">TEI conference</rs>` – if a/the proper name for it were used it could then even be `<eventName ref="#TEIMEC2023">TEI conference</eventName>` (or `<name type="event">` if you want to stay at 4.6.0).

It is the *event* (just as the person/object/place/bibl/relation), not the *reference* (that paradigm would also include ptr/link/ref) that should be dated.

Hence opposed to the agreement in the previous discussion on TEI-L I would argue against making <rs> a member of `att.datable`.

ebeshero:
In this particular instance of referencing a date by name, I tend to agree with @skurzinz . However, in the larger view, `<rs>` as a referencing string (rather than a named entity) might still be subject to date conditions. Can we imagine a different example to illustrate the temporality of such a string?

ebeshero:
I am thinking about references to stages in a person's life that are not necessarily events:

```
<p>In <rs notAfter="1985-02-11">a time before her marriage</rs> she thought differently on the subject of household income.</rs>
```

I just made that up--perhaps we can think of better examples.

skurzinz:
`CO-CoreElements.xml` says

> A referring string is a phrase which refers to some person, place, object, etc.

So the possible dating in question would only be used for temporalizing the validity of the string being a referring one. 

Made up example which I would argue against: 

```
[...]

PR #2952, opened 2026-08-17
#2389: rs member of att.datable

martinascholger:
implements issue #2389: rs should become a member of att.datable. Remark and example was added in addition.


martindholmes (approved):
Reviewed in Council yesterday.

PR #2954, opened 2026-08-17
added att.global.editorialStatus and realted guidelines language. Clo…

### LINKED ISSUE #2013: Add more elements to att.docStatus
Following this discussion on TEI-L:

http://tei-l.970651.n3.nabble.com/Encoding-status-of-items-rather-than-documents-td4032773.html

this ticket suggests that there is a need to be able to specify the status (draft, proofed, published etc.) not just of documents but of a variety of different elements which don't constitute documents in themselves. These are elements appearing in resource lists such as listPerson/person, listPlace/place, listBibl/bibl, list/item, listObject/object, and similar "entity records". 

The definition of att.docStatus specifically mentions that it applies to documents, but it could be re-defined so that its name is interpreted as "document the status of...". Alternatively, a new attribute class such as att.resourceStatus or att.fragmentStatus could be created.

raffazizzi:
This PR is meant to close #2013.

Initially Council suggested adding the new `@edStatus` attribute to `att.global.change`, but I didn't think this was appropriate since that class is part of the `transcr` module and is primarily meant to record authorial changes in primary sources. 

`@edStatus` is meant to track any kind of encoding workflow, so I created `att.global.editorialStatus` for `@edStatus` and placed it in the `tei` module.

I added prose in the Guidelines in `#STGAso` – this location is just a proposal and my language should be reviewed.

### DIFF
--- a/P5/Source/Guidelines/en/ST-Infrastructure.xml
+++ b/P5/Source/Guidelines/en/ST-Infrastructure.xml
@@ -367,12 +367,14 @@ with others.--></p>
       <div type="div2" xml:id="STGA">
         <head>Global Attributes</head>
         <p>The following attributes are defined in the infrastructure module for every TEI element. <specList>
-            <specDesc key="att.global" atts="xml:id n xml:lang rend style               rendition xml:base xml:space               source cert resp                "/>
+            <specDesc key="att.global" atts="xml:id n xml:lang rend style               rendition xml:base xml:space               source cert resp edStatus"/>
           </specList>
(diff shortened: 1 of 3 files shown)


martindholmes (approved):
This looks great. Should we have a Jenkins job so it's easier to look at everything in context?

ebeshero (approved):

ebeshero:
We built this locally during the last minutes of the F2F and determined that all looks good.

Issue #2846, opened 2025-12-02
remove `<schemaSpec>` from att.combinable?


sydb:
Why is `<schemaSpec>` even a member of att.combinable? I don’t think I have ever seen a `@mode` attribute on a `<schemSpec>`. What would such a thing even mean?

martindholmes:
I agree.

joeytakeda:
Also agreed. I imagine the semantics could have been “if schemaSpec says mode = change, then all descendant att.combinable elements are mode change unless otherwise specified,” but given that we don’t do that (and that’s not how it is defined), then it makes sense to remove.

ebeshero:
I faintly wonder whether `att.combinable` (`@mode`) on `<schemaSpec>` has anything to do with ODD chaining?

martinascholger:
@hennyu and I had a look at the DTA Basisformat, and it seems that using the `@source` attribute to declare the base format of an ODD is sufficient for chaining. See the example here: http://www.deutschestextarchiv.de/basisformat_ms.odd. We could not think of any use case where @mode would reasonably be used on `<schemaSpec>`

hennyu:
We could think of a theoretical useful case, though: in case `@mode` was not obligatory on `<elementSpec>`, then it would save work if one could declare `@mode` directly on `<schemaSpec>` to avoid putting in on all the `<elementSpec>`s, but only if the same mode was chosen for all of them.

ebeshero:
See my comments on [the PR](https://github.com/TEIC/TEI/pull/2958). I see a theoretical and potentially practical use-case for `@mode` on schemaSpec, but only if it is tightly constrained so as to reduce complexity, as @hennyu describes.

PR #2958, opened 2026-08-20
remove `@mode` from `<schemaSpec>`

sydb:
Per #2846, I removed `@mode` from the `<schemaSpec>` element itself, as it gets it from a subclass, not a class it is directly a member of. Passes tests in a Docker on my machine.

In case it is not obvious, this has not been deprecated because I consider it a (flagrant) corrigible error. Not only do we have no idea what `schemaSpec/@mode` means, I submit we do not have much clue as to what it _might_ mean.

I have listed all of Council except @martindholmes, who has already approved the idea, as possible reviewers. I do not really think it needs nine reviews. :smile: 

N.B. TEIC/Stylesheets#817

### DIFF
--- a/P5/Source/Specs/schemaSpec.xml
+++ b/P5/Source/Specs/schemaSpec.xml
@@ -17,7 +17,7 @@
   <classes>
     <memberOf key="att.global"/>
     <memberOf key="att.docStatus"/>
-    <memberOf key="att.identified"/>
+    <memberOf key="att.identified"/> <!-- except @mode removed -->
     <memberOf key="att.namespaceable"/>
     <memberOf key="model.encodingDescPart"/>
     <memberOf key="model.frontPart"/>
@@ -38,6 +38,7 @@
     </sequence>
   </content>
   <attList>
+    <attDef ident="mode" mode="delete"/>
     <attDef ident="start" usage="opt">
       <desc versionDate="2010-05-14" xml:lang="en">specifies entry points to the schema, i.e. which patterns may be used as the root of documents conforming to it.</desc>
       <desc versionDate="2007-12-20" xml:lang="ko">스키마에서 표제 항목 지점을 명시한다. 즉, 어떤 요소가 해당 문서의 뿌리로 사용되는 것이 허용되었는지를 명시한다.</desc>


P5/Source/Specs/schemaSpec.xml
```diff
@@ -38,6 +38,7 @@
     </sequence>
   </content>
   <attList>
+    <attDef ident="mode" mode="delete"/>
```
ebeshero, line 41: @sydb I think we may want to check and make sure that `@mode` on `<schemaSpec>` doesn't have a specific application in ODD chaining. Wouldn't it make sense in that context for :
* `@mode="add"` to mean, "this is a new schemaSpec in the chain"

* `@mode="delete"` to mean suppress the schemaSpec with this ident

* `@mode="change"`: merge this schemaSpec with another one with the same `@ident`?

* `@mode="replace"`: replace the existing schemaSpec with this one? 

Caveat: I don't do ODD chaining, but it made sense to me that this is how `@mode` might behave on this element... Should we test it first or ask around before removing it?

sydb: Perhaps those methods of processing schema specifications might make sense, but they are not how current ODD processing works nor how ATOP envisions working. (Nor what the _Guidelines_ say, but then again, they essentially say nothing about this. Nor is use of `schemaSpec/@mode` _ever_ exemplified.) In both those processing cases, the processor simply handles a single `<schemaSpec>`.¹ That `<schemaSpec>`, of course, may point to another with its `@source` attribute, in which case the two are merged.

I suppose we could set it up so that if a customization ODD has a `@mode` of "delete" it means “do not process my `@source`”, but wouldn’t it be easier to just not have a `@source`? But if a customization ODD does not have a source (either expressed or defaulted), it is not a customization ODD. So again, I just don’t know what it means. Same goes for "replace".

As for "add", what is the difference between adding two `<schemaSpec>`s together and merging them? Again, I just don’t know what it might mean. Perhaps it means “only take added constructs, ignore those with mode="change"” (or mode="delete")? Makes sense, but seems utterly useless.

**Note**
¹ For the current Stylesheets it is either the one specified by a parameter or, if the parameter is not set, the first one encountered in the input in document order; I do not remember off the top of my head what ATOP does, but we only handle one `<schemaSpec>`, and that is hard enough.

ebeshero: Rather than "utterly useless," perhaps `@mode` on `<schemaSpec>` introduces a level of complexity in processing that perhaps we prefer to dispense with. The logic of each combination of 4 possible values of `@mode` on `<schemaSpec>` AND 4 possible values of `@mode` on descendant `*Spec` would need to be explicitly sorted out.

Another possibility would be to apply the `@mode` on `<schemaSpec>` **only** to **reduce complexity** by indicating if it is present on this element, its value applies to all descendant `*Spec` elements, which are then not permitted to carry a `@mode`, since it would be inherited. That would require a Schematron constraint and some explanation in the Guidelines.

Since we don't have any examples in the Guidelines, it seems okay to remove it without deprecation, with a message in the release notes that this might affect someone's ODD (we have a categorical heading for that in release notes). 

I do wonder about the possibility of constraining the use of `@mode` rather than removing it from `<schemaSpec>` *only* in a way to simplify (not to introduce confusing combinations).

sydb: 1. I did not say `schemaSpec[@mode]` is utterly useless, I said taking `schemaSpec/[@mode eq 'add']` to mean “only take added constructs, ignore [others]” is utterly useless. While you may convince me otherwise, for now I still think it is.
2. Since there exists one extant case I know of in which a customization ODD has only one value of `@mode` on any descendant *Spec element,¹ I concede that yours is a possible interpretation of `@mode` on `<schemaSpec>`.² I will change this PR and the issue it is based on to NEEDS DISCUSSION and remove the milestone.

**Notes**
¹ I am, as is our usual, including `<attDef>` in the set of elements indicated by “*Spec”.
² The [SynSemSpec ODD](https://raw.githubusercontent.com/DARIAH-ERIC/lexicalresources/refs/heads/master/Schemas/LMF%20SynSem%20Specification/Specification/SynSemSpec.odd) file has only `@mode`s of "add", nothing else.³ Note that ATOP has created several customization ODDs that meet this criteria, but they were created for testing ATOP, not for real world use, so I am not counting them.
³ This ODD does not seem to be intended to be a TEI conformant ODD, as the added elements are in the TEI namespace, but that does not change any of the arguments on this ticket, I don’t think.

martindholmes: I think it's there entirely by accident, and inventing obscure usages that have never been handled or dreamed of and never will be doesn't help. :-)

PR #2959, opened 2026-08-21
Issue 2840 mdh

martindholmes:


P5/Source/Specs/moduleSpec.xml
```diff
@@ -36,6 +36,15 @@ module, i.e. a named and externally visible group of declarations.</desc>
         <elementRef key="listRef" minOccurs="0" maxOccurs="unbounded"/>
     </sequence>
   </content>
+  <constraintSpec ident="specs-without-module-att-deprecated" scheme="schematron" xml:lang="en">
+    <constraint>
+      <sch:rule context="tei:*[local-name() = ('elementSpec', 'classSpec', 'macroSpec', 'dataSpec')][not(@module)]">
+        <sch:report test="true()" role="warning">
+          The @module attribute will be required on *Spec elements after 2027-09-30.
```
sydb: Two thoughts:

1. Probably better to use `&lt;<sch:name/>>` instead of “*Spec”, as normal users may never have been exposed to that phrasing, and some would (quite reasonably) presume that `<schemaSpec>` belongs to that group.
2. End-of-deprecation date seems a bit close, given that this is backwards-incompatible. It means at most two releases with deprecation, but more likely only one, and at worst (say, a release in Feb and another in Oct) _none_.

sydb (changes requested):
Test itself is correct for what it is, and this passes tests locally. So all well and good. But I am a bit concerned that we (ATOP, let alone Council) have not discussed this constraint for `<macroSpec>` and `<dataSpec>` (not to mention `<constraintSpec>` and `<paramSpec>`) yet.

P5/Source/Specs/moduleSpec.xml
martindholmes: Done. I made it `<sch:name/> and other *Spec elements` because it would be infuriating to hit one error on one spec element, fix it, then get the next one, and so on. At least this way you know what to look for when the first error shows up. Date is now 2028-06-30.

martindholmes:
MH to merge dev into the branch, and fix the tests, and then Council says we can merge it.

Issue #2729, opened 2025-07-02
ODD processing for glosses and descs


martindholmes:
Arising from ATOP discussions with @sydb  and @HelenaSabel: 

If `gloss[parent::attDef[@mode="change"]]`, what should happen? Should this gloss be deemed to replace some existing gloss -- and if so, if there are multiple `<gloss>`es, then which one? The one with the same `@xml:lang`? And if there isn't one with the same `@xml:lang`, should the new `<gloss>` just be added?

And in the case of `<remarks>`, should we be matching the combination of `@xml:lang` and `@ident`? (This leaves aside for the moment the question of `<exemplum>`, which we believe should also have `@ident` and `@mode` to potentially enable the same mechanism, but which has other unsolved problems being handled in https://github.com/TEIC/TEI/issues/1933.)

peterstadler:
From my understanding (and the practice in the TEI Guidelines) there's always just one desc/gloss/remarks for each language (with the exception of `<desc type="deprecationInfo">` which is to be excluded from this rule). 
My proposal would be to make this assumption explicit in the Guidelines and proceed with the ODD processing as proposed by @martindholmes : a `@mode="change"` on the parent element would result in a replacement of the respective language if present, else it would be added.

sydb:
I think @peterstadler’s idea is the right approach, at least for P5. Have not thought through if there are wider implications, though.
In the meantime, I checked: Of the 851 elements in p5.xml that have a `<gloss>` child, _none_ has 2+ with the same value of `@xml:lang`. (I did not check if any have 2+ that _do not have_ `@xml:lang`, because I don’t think we care about those.)

I am assigning to myself to write a constraint (in p5odds.odd unless Council says otherwise) to enforce this.
I am hoping someone else self-assigns to make sure the Stylesheets do the right thing.

**Note-to-self**
In pure XPath: `for $gp in //t:*[t:gloss] return concat( local-name($gp), ' of or in ', $gp/ancestor-or-self::*[@ident][1]/@ident, ': ', count( $gp/t:gloss/@xml:lang ), ', ', count( distinct-values( $gp/t:gloss/@xml:lang ) ), ', which means ', count( $gp/t:gloss/@xml:lang ) eq count( distinct-values( $gp/t:gloss/@xml:lang ) ) )`. Yes, probably would have been a lot easier to read if it were in XQuery or XSLT with a few variables, but I happened to be in Gizmo with P5 loaded, so this is what I did.

[...]

PR #2960, opened 2026-08-22
allow only one child `<desc>`, `<gloss>`, or `<remarks>` of a given language

sydb:
Per #2729, move the "only-one-each-lang" constraint from p5odds to p5 itself.

There are some minor differences between the code that was deleted and that which was added, though:

- In p5odds each clause of the `@context` had a predicate to prevent firing on the 1st sibling (because it has no preceding siblings that might meet the criteria we are investigating). While it might have been faster, and some might think it clearer, it might have been slower, and others would say it is more confusing. So, given that those predicates were previously problematic, I removed them. The `preceding-sibling::` axis will always be empty, so the `<sch:report>` will never fire on the 1st sibling, anyway. 
- In p5odds we _knew_ there was an `@xml:lang` on each  `<desc>`, `<gloss>`, or `<remarks>` because the schema required it. Here we do not know, so we need to look at `ancestor-or-self::*[@xml:lang][1]/@xml:lang` rather than just `@xml:lang`.
- There were things in the `<desc>` that should not appear on the tagdoc, so got moved to comments.
- The previous code failed to normalize spaces of the `@type` attrs before comparison.

### DIFF
--- a/P5/Source/Specs/att.translatable.xml
+++ b/P5/Source/Specs/att.translatable.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- © TEI Consortium. Dual-licensed under CC-by and BSD2 licenses; see the file COPYING.txt for details. -->
 <?xml-model href="https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5.nvdl" type="application/xml" schematypens="http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0"?>
-<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tagdocs" type="atts" ident="att.translatable" predeclare="true">
+<classSpec xmlns="http://www.tei-c.org/ns/1.0" module="tagdocs" type="atts" ident="att.translatable" xmlns:sch="http://purl.oclc.org/dsdl/schematron" predeclare="true">
   <desc versionDate="2006-10-15" xml:lang="en">provides attributes used to  indicate the status of a translatable
 portion of an ODD document.</desc>
   <desc versionDate="2007-12-20" xml:lang="ko">ODD 문서의 번역 가능 부분의 상태를 나타내는 속성을 제시한다.</desc>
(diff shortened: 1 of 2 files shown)


P5/Source/Specs/att.translatable.xml
```diff
@@ -11,6 +11,53 @@ portion of an ODD document.</desc>
...
+                         | tei:valItem/tei:gloss
+                         | tei:valItem/tei:remarks
+                         | tei:moduleSpec/tei:desc
+                         | tei:moduleSpec/tei:gloss
+                         | tei:moduleSpec/tei:remarks
+                         ">
+        <sch:let name="myGI" value="name(.)"/>
+        <sch:let name="noType" value="not( @type )"/>
+        <sch:let name="myType" value="normalize-space( @type )"/>
+        <sch:let name="lang" value="normalize-space( ancestor-or-self::*[@xml:lang][1]/@xml:lang )"/>
```
martindholmes, line 53: This line would specify that two sibling elements with no `@xml:lang` are deemed to have the same `@xml:lang` (that of their first ancestor which has the attribute). Presumably this is the correct/intended behaviour?

P5/Source/Specs/att.translatable.xml
```diff
@@ -11,6 +11,53 @@ portion of an ODD document.</desc>
...
+        <sch:report test="preceding-sibling::*[ name(.) eq $myGI ]
+                                              [ ( not( @type) and $noType )  or  normalize-space( @type ) eq $myType ]
```
martindholmes, line 55: Why both `not(@type)` and `$noType`?

P5/Source/Specs/att.translatable.xml
sydb: It is certainly what I intended.  It is correct in that the [XML Spec](https://www.w3.org/TR/REC-xml/#sec-lang-tag) says “… `@xml:lang` applies to the element where it is specified … and to all elements in its content unless overridden with another instance of `@xml:lang`.” Whether it is correct with respect to what TEI should do here may be a matter of opinion.

P5/Source/Specs/att.translatable.xml
sydb: `$noType` is true() iff the context node (a `<gloss>`, `<desc>`, or `<remarks>` child of an ODD element) does _not_ have a `@type`. And `not( @type )` is true() iff the node currently being examined (a preceding element node of the context node) does _not_ have a `@type` attribute. Thus the predicate
~~~xml
[ ( not( @type) and $noType )  or  normalize-space( @type ) eq $myType ]
~~~
says “let’s compare the categorization of the context node with the node currently being examined: IF a) they both have the same value of `@type`, OR b) if neither one of them has a `@type`, THEN consider them to be categorized the same way and return true()”. 

I would have thought the first clause (`not( @type ) and $noType`) is redundant, because surely if neither has a `@type` attribute then `normalize-space( @type )` would return the same thing (empty string) for both the context node (`$myType`) and the currently-being-examined node. But IIRC I found this code did not work without that first clause. But it seems so weird. Maybe I should poke at it again.

But if it is required, probably a comment explaining it, or at least better variable names, are in order, eh? (`$context_node_has_no_type`?)

martindholmes: Got it. I was just being a bit dim.

raffazizzi (approved):

Issue #2949, opened 2026-08-16
two mistakes in the Guidelines


peterstadler:
as reported via `webmaster@tei-c.org`:

> There seems to be a little error in this example:
https://tei-c.org/release/doc/tei-p5-doc/en/html/CO.html#CONADA
`<date when="2001-09-11T12:48:00">Sept 11th, 12 minutes before 9 am</date>`
I think the time should be T08:48:00.

>And here is another little mistake:
https://www.tei-c.org/Vault/P5/3.2.0/doc/tei-p5-doc/en/html/ref-att.datable.custom.html
"The when-custom attribute and other custom dating are not contrained to a datatype by the TEI, ..."
I think that the word should read "constrained" (the letter S is missing).

martindholmes:
@peterstadler Is it possible that the first case is using UTC in the attribute (I think it's the default), while the text content is New York time?

martindholmes:
@peterstadler On the second example, it points to Glines 3.2.0, but I can't find the spelling error mentioned either in 3.2.0 or the current version of that page, nor can I find the typo "contrained" anywhere in the Glines.

martindholmes:
Pending waiting for a response either from @peterstadler or from the original emailer, whoever that was. I think these might just be misunderstandings.

peterstadler:
Disclaimer: I just pasted the text from the email that reached me without checking. 

So, the second issue is already fixed. In the current version of [att.datable.custom](https://www.tei-c.org/Vault/P5/4.12.0/doc/tei-p5-doc/en/html/ref-att.datable.custom.html) the "s" is present.

The first one I consider an issue, at least confusing. Either the value in the `@when` attribute should be aligned with the text or there should be some remark about time zones.

martindholmes:
@peterstadler On the first issue, the immediately-following text says: "Note in the last example the use of a normalized representation for the date string which includes a time". Isn't that quite literally what's happening? I think the date and time is that of the 9-11 attack, expressed as it was experienced in the location of the attack; the `@when` value normalizes this to UTC, and surely can't be misconstrued. But we could add a note to that effect, I suppose...

martindholmes:
@peterstadler You're right -- the Z was missing from `@when`, which would mean "local time". I'm doing a PR now to add a clarification and to add the Z.

martinascholger:
Implemented with https://github.com/TEIC/TEI/pull/2961

PR #2961, opened 2026-08-24
Fix for issue #2949: add missing Z in @when and explanation.

martindholmes:
Simple fix for a confusing datetime example, per issue #2949.

### DIFF
--- a/P5/Source/Guidelines/en/CO-CoreElements.xml
+++ b/P5/Source/Guidelines/en/CO-CoreElements.xml
@@ -2443,10 +2443,13 @@ Some typical examples follow:
 <date when="--09">September</date>
 <date when="---11">Eleventh of the month</date>
 <time when="08:48:00">8:48</time>
-<date when="2001-09-11T12:48:00">Sept 11th, 12 minutes before 9 am</date>
+<date when="2001-09-11T12:48:00Z">Sept 11th, 12 minutes before 9 am</date>
 </egXML>Note in the last example the use of a normalized representation for
 the date string which includes a time: this example could thus equally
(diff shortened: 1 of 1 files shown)


sydb (approved):

P5/Source/Guidelines/en/CO-CoreElements.xml
```diff
@@ -2443,10 +2443,13 @@ Some typical examples follow:
...
 <date when="---11">Eleventh of the month</date>
 <time when="08:48:00">8:48</time>
-<date when="2001-09-11T12:48:00">Sept 11th, 12 minutes before 9 am</date>
+<date when="2001-09-11T12:48:00Z">Sept 11th, 12 minutes before 9 am</date>
 </egXML>Note in the last example the use of a normalized representation for
 the date string which includes a time: this example could thus equally
-well be tagged using the <gi>time</gi> element. 
+well be tagged using the <gi>time</gi> element. The time in the <att>when</att>
+  attribute is normalized to UTC (signified by the <q>Z</q>), while the time in 
+  the text content of the element expresses local (New York) time, which is four
```
sydb, line 2451: I would personally prefer to leave it as an exercise for the reader to figure out that local time is NYC time. Thus “the text content of the element expresses local time, in this case four hours behind UTC”. But that may be just me.

sydb (approved):

P5/Source/Guidelines/en/CO-CoreElements.xml
peterstadler: I think it's fair to provide an explanation. To me at least it was not clear at all.

peterstadler (approved):

Issue #2938, opened 2026-07-29
The `<valDesc>` element should not be used in the Guidelines


martindholmes:
In evaluating this pull request:

https://github.com/TEIC/TEI/pull/2888

I noticed that the new att.linguistic.dependency class uses `<valDesc>` elements, and rather than a noun clause or a full sentence, uses a pronoun-dropping elliptical sentence. Since this looked odd to me, I went to look at the rest of the specs and discovered that we have only used `<valDesc>` once before, on `num/@value`, where it is a noun phrase, "a numeric value":

https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-num.html

So I went to TCW20 to see if I could find instructions on how the contents of a `<valDesc>` should be formulated. There I found only this:

"Note that we do not use `<valDesc>` at this time, instead using only a `<datatype>`."

This was added in 2013 by KSH:

`<change when="2013-12-22" who="KSH">Some clarifications, including not to use valDesc.</change>`

So I conclude that:

a) For some reason, we decided we should not use `<valDesc>` at all in the Guidelines. Does anyone remember why? 
b) If we shouldn't use it, then we should replace the instance we currently have with something else (a `<remarks>`???), and then make it invalid in p5_odds.odd.
c) If we want to reverse our decision not to use it, we should update TCW20 to remove the text which says not to use it, and also update TCW24 to include guidance on what form its content should take (noun phrase, full sentence, whatever), and bring the existing instance and the incoming instances from the PR into line with what we decide.

martindholmes:
Council meeting today: We don't use it, and so we should replace the three isolated instances of valDesc in the current Guidelines, and make valDesc invalid in p5odds. MH to do this (and change the title of this ticket appropriately).

martindholmes:
See also issue #2875, which suggests a better mechanism for a prose-based constraint that will be more widely available.

PR #2971, opened 2026-09-13
Remove the `<valDesc>` element from p5odds

sydb:
Title says it all; this is a response to #2938.

Per Council meeting of 2026-09-11, the `<valDesc>` element should no longer occur in P5 itself.


sydb:
We will use #2976 instead

Issue #2972, opened 2026-09-13
remarks about exponential notation misencoded and mildly misleading


sydb:
The `<remarks>` about exponential notation in teidata.numeric.xml (which I _think_ I wrote roughly 2 decades ago, when it was still called data.numeric) seem to me to be mis-encoded (certainly “exponential notation” is not so-called, that’s its name; and while “E notation” _may_ have been so-called back then, certainly it is accepted nomenclature now, right?¹) and the example mixes precision inappropriately.

¹ Although [this page](https://calculators.im/scientific-notation-converter) spells it “E-notation”, [Wikipedia](https://en.wikipedia.org/wiki/Scientific_notation#E_notation) does not hyphenate it.

PR #2973, opened 2026-09-13
Minor re-write of para describing expressions in exponential notation

sydb:
Addressing #2972 , title says it all.

NOTE to reviewers — the changes to attList.xml were checked in accidentally (they were sitting in my directory after a meeting last month in which this code was being talked about). The change is **only to whitespace**, there is no change to the code. Since this code is tested in detest.odd (search for “no_duplicate_attrs”), I do not think you need to even look at it as long as it pasts tests.


tuurma (approved):
thx for fixing the indenting

